PR #9, opened 2026-04-04
fix(cli): OpenAI-compatible serve responses, stream lock lifecycle, timeout, param passthrough, cross-platform sysfs reads, SSE headers

domvox:
## CLI/API fixes: OpenAI-compatible response format, stream lock lifecycle, timeout, param passthrough, SSE headers, cross-platform sysfs reads

This PR fixes 7 CLI/API/serve issues identified while testing `hipfire v0.1.2-alpha` on RX 7900 XTX (gfx1100, ROCm 6.4).

These changes improve:
- Compatibility with OpenAI-style clients (Open WebUI, LiteLLM, etc.)
- Stream lifecycle robustness
- Cross-platform portability
- Sampling parameter passthrough in serve mode
- SSE transport behavior

This PR does **not** address core runtime issues (output corruption, context-length degradation, concurrent state bleed, kernel regressions). Those are tracked separately in #[issue number].

---

### Fix 1: Server idle timeout during model load

**Problem:** Bun.serve uses a 10-second default idle timeout. First request during model load (which can take 15-30s) gets silently dropped.

**Fix:** Set `idleTimeout: 255` (Bun maximum).

---

### Fix 2: OpenAI-compatible response fields

**Problem:** Response JSON was missing standard fields expected by OpenAI-compatible clients.

**Fix:** Added to both streaming and non-streaming paths:
- `id` — unique per request (`chatcmpl-{timestamp}`)
- `object` — `chat.completion` / `chat.completion.chunk`
- `created` — Unix timestamp
- `model` — requested model name
- `usage` — `prompt_tokens`, `completion_tokens`, `total_tokens`
- `choices[].index`
- `choices[].finish_reason`

---

### Fix 3: Streaming lock release

**Problem:** Streaming responses returned a `ReadableStream` but the request lock was never explicitly released on completion or error. Subsequent requests could hang indefinitely.

**Fix:** Wrapped streaming generation loop in `try/finally` with `releaseLock()` in `finally`.

---

### Fix 4: Double lock release in non-streaming path

**Problem:** Non-streaming path called `releaseLock()` explicitly before `return`, then again in the outer `finally` block — double release on every successful request.

**Fix:** Removed the explicit call. Lock cleanup is now centralized in `finally` for both paths.

---

### Fix 5: Sampling parameter passthrough in serve

**Problem:** `serve` mode did not forward `repeat_penalty` or `top_p` to the daemon. These parameters were silently ignored in API requests, despite being accepted by the CLI `run` command.

**Fix:** Added `repeat_penalty` and `top_p` to the generation params object. Also maps OpenAI's `frequency_penalty` to `repeat_penalty` for compatibility.

---

### Fix 6: Cross-platform GPU arch detection in `update`

**Problem:** `update` command used `Bun.spawnSync(["cat", ...])` to read sysfs GPU properties. This fails on Windows (no `cat` binary) and is fragile on Linux distributions where `cat` may not be in PATH during restricted execution contexts.

**Fix:** Replaced with `Bun.file().text()` — native async file read, no external process dependency. Added fallback from topology node 1 to node 0.

---

### Fix 7: SSE transport headers

**Problem:** Streaming responses were missing `Cache-Control: no-cache`, which can cause buffering issues with proxies and some HTTP clients.

**Fix:** Added `Cache-Control: no-cache` header to SSE responses.

---

### Validation

All fixes tested on RX 7900 XTX / gfx1100 / ROCm 6.4:

| Test | Result |
|------|--------|
| 3 sequential streaming requests | ✅ |
| 5 sequential non-streaming requests | ✅ |
| Mixed sequence: stream → non-stream → stream | ✅ |
| OpenAI fields present in non-streaming response | ✅ |
| OpenAI fields present in streaming chunks | ✅ |
| `repeat_penalty` + `top_p` accepted via API | ✅ |
| `Bun.file()` sysfs read (gfx_target_version) | ✅ |
| Serve error log after all tests | ✅ NONE |

---

### Out of scope

This PR intentionally does not address deeper runtime issues identified during testing:

- Shared output corruption across HFQ4/HFQ6 (identical `2,3,5711` on both quants)
- Context-length-sensitive quality and speed degradation (-65% tok/s at ~500 prompt tokens)
- Confirmed cross-request state bleed in concurrent serve
- Precompiled gfx1100 kernel correctness issue
- HFQ4-specific throughput regression (~20% vs v0.1.0)
- `<think>` tokens consuming user-visible `max_tokens` budget
- Prompt escape sequence handling

These require investigation in the Rust engine / daemon layer and are documented in the linked issue.

---

### Rationale for separate PR

These fixes are:
- Low-risk — no changes to engine, kernels, or model loading
- Self-contained — all changes in `cli/index.ts`
- Independently useful — improve serve usability regardless of runtime bug status
- Easy to review — single file, clear before/after behavior

### DIFF
--- a/cli/index.ts
+++ b/cli/index.ts
@@ -274,6 +274,7 @@ async function serve(port: number) {
 
   Bun.serve({
     port,
+    idleTimeout: 255, // max allowed — model loading can take 30s+
     async fetch(req) {
       const url = new URL(req.url);
       if (url.pathname === "/health") return Response.json({ status: "ok", model: current });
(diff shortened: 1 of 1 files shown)


Kaden-Schutt (changes requested):
Good PR — the fixes are real and well-documented. Two bugs to fix before merge:

### 1. `repeat_penalty` operator precedence

```typescript
repeat_penalty: body.repeat_penalty ?? body.frequency_penalty ? 1.0 + (body.frequency_penalty ?? 0) : 1.3,
```

`??` has higher precedence than `?:`, so this parses as `(body.repeat_penalty ?? body.frequency_penalty) ? ... : 1.3`. If a user sends `repeat_penalty: 1.1`, the value gets ignored — result is `1.0`. Needs explicit parens:

```typescript
repeat_penalty: body.repeat_penalty ?? (body.frequency_penalty ? 1.0 + body.frequency_penalty : 1.3),
```

### 2. Streaming lock double-release / early release

The inner `try/finally { releaseLock() }` inside the ReadableStream is the right idea, but the outer `finally { releaseLock() }` still fires on `return new Response(...)` — before the stream finishes generating. This releases the lock immediately, allowing concurrent requests to start mid-stream. The inner finally then double-releases.

Fix: guard the outer finally so it skips streaming paths:
```typescript
let streaming = false;
try {
  // ...
  if (body.stream) {
    streaming = true;
    return new Response(new ReadableStream({ ... }));
  }
  // non-streaming path...
} finally { if (!streaming) releaseLock(); }
```

Everything else looks good — idleTimeout, OpenAI fields, Bun.file() sysfs, Cache-Control header. Happy to merge once these two are fixed.

Issue #8, opened 2026-04-04
[gfx1100][ROCm 6.4] v0.1.2-alpha: shared output corruption, severe context-length degradation, confirmed cross-request state bleed, precompiled kernel issues


domvox:
Hi — I did a full diagnostic pass on **hipfire v0.1.2-alpha** on **RX 7900 XTX (gfx1100, 24GB)** and compared it against **v0.1.0** on the same machine.

This report focuses on **core runtime / output correctness / serve isolation**.
I'm filing a separate PR for CLI/API fixes.

---

## Environment

| Component | Value |
|-----------|-------|
| GPU | AMD Radeon RX 7900 XTX |
| Arch | gfx1100 |
| VRAM | 24 GB |
| CPU | Ryzen 9 9950X3D |
| OS | openSUSE Tumbleweed |
| ROCm | 6.4 |
| hipfire | v0.1.2-alpha |
| Comparison | v0.1.0 |

---

### Confirmed improvements
- `serve` is much more stable than in v0.1.0
- 27B works via serve
- Sequential request handling is stable
- Stream cancel + recovery works

### Confirmed problems
1. **27B HFQ4 throughput regression** vs v0.1.0
2. **Shared output corruption symptoms across HFQ4 and HFQ6**
3. **Severe context-length degradation** on 27B HFQ6
4. **Cross-request state bleed in concurrent serve**
5. **Precompiled gfx1100 kernels produce garbage output** on my setup
6. **Prompt escape handling is incorrect** (`\n`/`\t` not interpreted)
7. **`<think>` tokens count toward `--max-tokens`**

---

## Benchmark summary

| Model / Quant | v0.1.0 | v0.1.2-alpha | Delta |
|---------------|--------|-------------|-------|
| 9B HFQ4 | ~62 tok/s | 58–61 tok/s | roughly flat |
| 27B HFQ4 | 25–27 tok/s | 20–22 tok/s | **~20% regression** |
| 27B HFQ6 | 16–18 tok/s | 16–18 tok/s | stable |
| llama.cpp Qwen3.5-27B Q5_K_M | 15–18 tok/s | 15–18 tok/s | reference |

---

### 1. Determinism test

**Setup:** temp=0, same prompt, 5 runs

**Result:** Output is **identical across runs**. Problems are systematic, not random.

---

### 2. Exact-copy test

**Prompt:** `Repeat exactly: alpha beta\ngamma\tdelta`

| Model | Output | Verdict |
[...]

Issue #5, opened 2026-04-04
$dllName: error


Floris-Jan:
install.ps1 is got an error

```
irm https://raw.githubusercontent.com/Kaden-Schutt/hipfire/master/scripts/install.ps1 | iex
iex : At line:85 char:27
+             Write-Host "  $dllName: found at $candidate ✓" -Foregroun ...
+                           ~~~~~~~~~
Variable reference is not valid. ':' was not followed by a valid variable name character. Consider using ${} to delimit the name.
```

replace with ${dllName}:

Kaden-Schutt:
Fixed in v0.1.2-alpha — `$dllName:` → `${dllName}:` on all three Write-Host lines (commit e0af099, PR #7).

Re-run the install command and it should work:
```powershell
irm https://raw.githubusercontent.com/Kaden-Schutt/hipfire/master/scripts/install.ps1 | iex
```

Issue #2, opened 2026-04-03
gfx1100 (RX 7900 XTX): pre-compiled kernels produce garbage output, local recompile fixes it


domvox:
## Environment
- **GPU:** AMD RX 7900 XTX (gfx1100, 24GB VRAM)
- **OS:** openSUSE Tumbleweed
- **ROCm/HIP:** 6.4.43484-9999 (clang 19.0.0)
- **hipfire:** installed 2026-04-03 via install.sh from master

## Problem

Both Qwen3.5-9B (HFQ4) and Qwen3.5-27B (HFQ6) load correctly but generate only `!` characters when using shipped pre-compiled kernels at `~/.hipfire/bin/kernels/compiled/gfx1100`.

Model loading is correct (all layers detected, correct vocab 248320, correct dimensions). Speed is real. But every token decodes to `!`.

## Reproduction

```bash
hipfire pull qwen3.5:9b
hipfire run qwen3.5:9b --no-think "What is 2+2?"
# Output: !!!!!!!!!!!!!!!!!!!! (512 tokens)
# [512 tok, 60.9 tok/s]
```

Same with 27B HFQ6 (17.9 tok/s, all `!`).

`HSA_OVERRIDE_GFX_VERSION=11.0.0` and `HIP_LAUNCH_BLOCKING=1` did not help.

## Fix

```bash
rm -rf ~/.hipfire/bin/kernels/compiled/gfx1100
hipfire run qwen3.5:9b --no-think "What is 2+2?"
# Output: actual text ✅
# [67 tok, 11.1 tok/s]
```

## Results

| | Pre-compiled kernels | Locally recompiled |
|---|---|---|
| 9B HFQ4 output | ❌ `!!!...` | ✅ real text |
| 9B HFQ4 speed | 60.9 tok/s | 11.1 tok/s |
| 27B HFQ6 output | ❌ `!!!...` | ✅ real text |
| 27B HFQ6 speed | 17.9 tok/s | 18.8 tok/s |

The 9B model drops from 60 to 11 tok/s after local recompile. 27B stays ~18 tok/s. Likely the pre-compiled blobs had optimizations that the local compile does not reproduce.

## Additional notes

- **27B repetition:** After recompile, 27B generates correct text for ~2 sentences then falls into a repetition loop. Probably needs repeat_penalty in sampling.
[...]

Issue #1, opened 2026-04-02
Unify model file naming convention


Kaden-Schutt:
## Problem

Model filenames are inconsistent across families and quant types:
- Qwen3.5: `qwen3.5-9b.q4.hfq` (HFQ4) vs `qwen3.5-9b.hfq6.hfq` (HFQ6)
- Qwen3: `qwen3-0.6b-hfq4.hfq` (dashes instead of dots)
- The `.hfq` extension sometimes doubles: `.hfq6.hfq`
- Experimental variants have ad-hoc suffixes: `-v2`, `-q4k-all`, `-hfq4g256`

This happened organically as formats were added during development.

## Proposed Convention

```
{family}-{size}.{quant}
```

Where quant IS the file extension: `.hfq4`, `.hfq6`, `.q8`

Examples:
```
qwen3.5-9b.hfq4
qwen3.5-9b.hfq6
qwen3-8b.hfq4
```

## Impact

- Rename files on HuggingFace repos
- Update CLI model registry (`cli/index.ts` REGISTRY)
- Update quantizer default output naming
- Update `findModel()` in CLI
- Update docs and benchmark scripts

## Current Workaround

The CLI tags (`qwen3.5:9b`, `qwen3.5:9b-hfq6`) are the stable interface. Use tags instead of filenames directly.

See `docs/MODELS.md` for full reference of current filenames and meanings.

## Labels

enhancement, alpha

Kaden-Schutt:
Updated in v0.1.3-alpha (96000be):

**New naming:** `{family}-{size}.{hf4|hf6}`
- `qwen3.5-9b.hf4` (was `qwen3.5-9b.q4.hfq`)
- `qwen3.5-9b.hf6` (was `qwen3.5-9b.hfq6.hfq`)

**CLI tags:** `qwen3.5:9b` (default=hf4), `qwen3.5:9b-hf6`

**Backward compat:**
- Old tags (`-hfq6`) auto-resolve to new names
- Old `.hfq` files found locally still load
- `findModel()` searches for .hf4, .hf6, and legacy .hfq

**Remaining:** HuggingFace repos need file renames to match new convention.

PR #20, opened 2026-04-14
test: add issue #8 regression prompt suite

domvox:
Formalizes the minimal repro prompts from #8 as `tests/regression_issue8.sh`.

Covers:
1. Primes — tokenizer byte-map, digit concatenation
2. Numbers 1-10 — output assembly, separators
3. JSON echo — escape handling, special chars (validated with `jq`)
4. Single token — hallucination guard (100 vs 1024)
5. Exact copy — whitespace collapse
6. Isolation — cross-prompt state leak

Usage: `MODEL=qwen3.5:9b tests/regression_issue8.sh`

Ref: #8

### DIFF
--- a/tests/regression_issue8.sh
+++ b/tests/regression_issue8.sh
@@ -1,0 +1,114 @@
+#!/usr/bin/env bash
+# Regression prompts from issue #8 -- output correctness suite
+# https://github.com/Kaden-Schutt/hipfire/issues/8
+#
+# Tests tokenizer integrity, structured output, and decode-path
+# correctness. Run against any model to catch regressions.
+#
+# Usage: MODEL=qwen3.5:9b tests/regression_issue8.sh
+#        MODEL=qwen3.5:35b-a3b tests/regression_issue8.sh
+
+set -uo pipefail
+
+MODEL=${MODEL:-qwen3.5:9b}
+PASS=0
+FAIL=0
+
+run() { hipfire run "$MODEL" --max-tokens 64 --temp 0 "$1" 2>/dev/null; }
+
+norm() { printf '%s' "$1" | tr -d '\r' | sed 's/[[:space:]]*$//; s/^[[:space:]]*//'; }
+
+check_contains() {
+    local name="$1" expected="$2" output="$3"
+    if echo "$output" | grep -qF "$expected"; then
+        echo "PASS $name"
+        ((PASS++))
+    else
+        echo "FAIL $name"
+        echo "  expected to contain: $expected"
+        echo "  got: $output"
+        ((FAIL++))
+    fi
+}
+
+check_exact() {
+    local name="$1" expected="$2" output="$3"
+    if [[ "$(norm "$output")" == "$expected" ]]; then
+        echo "PASS $name"
+        ((PASS++))
+    else
+        echo "FAIL $name"
+        echo "  expected: $expected"
+        echo "  got: $(norm "$output")"
+        ((FAIL++))
+    fi
+}
+
+check_not_contains() {
+    local name="$1" bad="$2" output="$3"
+    if echo "$output" | grep -qF "$bad"; then
+        echo "FAIL $name (found: $bad)"
+        ((FAIL++))
+    else
+        echo "PASS $name"
+        ((PASS++))
+    fi
+}
+
+echo "=== Issue #8 regression suite -- $MODEL ==="
+
+# 1. Primes: catches tokenizer byte-map bugs, digit concatenation (2,3,5711)
+echo "[1] Primes"
+OUT=$(run "Return the first 5 prime numbers, comma-separated. Nothing else.")
+check_contains "primes: has 2, 3, 5, 7, 11" "2, 3, 5, 7, 11" "$OUT"
+check_not_contains "primes: no concatenation" "5711" "$OUT"
+
+# 2. Numbers: catches output assembly corruption, missing separators
+echo "[2] Numbers 1-10"
+OUT=$(run "Return numbers 1 to 10, one per line. Nothing else.")
+LINES=$(echo "$OUT" | grep -cE '^[0-9]+$' || true)
+if [[ "$LINES" -ge 10 ]]; then
+    echo "PASS numbers: $LINES numeric lines"
+    ((PASS++))
+else
+    echo "FAIL numbers: expected >=10 numeric lines, got $LINES"
+    echo "  output: $OUT"
+    ((FAIL++))
+fi
+
+# 3. JSON: catches escape handling, special character corruption
+echo "[3] JSON"
+OUT=$(run 'Return exactly this JSON, nothing else: {"a":1,"b":2}')
+if echo "$OUT" | jq -e '. == {"a":1,"b":2}' >/dev/null 2>&1; then
+    echo "PASS json: exact match"
+    ((PASS++))
+elif echo "$OUT" | jq -e '.a == 1 and .b == 2' >/dev/null 2>&1; then
+    echo "PASS json: keys correct (whitespace differs)"
+    ((PASS++))
+else
+    echo "FAIL json"
+    echo "  got: $OUT"
+    ((FAIL++))
+fi
+
+# 4. Single token: catches hallucination (1024 instead of 100)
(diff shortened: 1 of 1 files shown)


Kaden-Schutt:
## Why this needs to close as not-mergeable

Closing this with apologies — the test design is structurally
incompatible with how Qwen3.x thinking models behave under
deterministic decode, and after a long testing session today I'm
confident no decode-side tuning gets to a reliable 9/9. Below is the
empirical matrix and root-cause analysis for the record so we don't
chase this again.

---

### TL;DR

The suite runs `hipfire run --max-tokens 64 --temp 0` against Qwen3.5
thinking models. Two structural problems make it unfixable as a
gating regression suite:

1. **`--max-tokens 64` is below Qwen3.5's `<think>` block budget.**
   Qwen3.5's reasoning step routinely costs 30–200 tokens before
   closing `</think>`. With max=64 the model never escapes thinking,
   the CLI's filter strips the unclosed block, stdout is empty.
2. **At the budgets that DO let the model finish, decode is
   stochastic on these specific prompt shapes.** Failures shift
   between primes / numbers / copy / JSON depending on which greedy
   attractor the model lands in on a given run, with no decode
   setting (temp 0–0.7, top-p 0.8–0.95, normalize on/off, model
   size 0.8b–9b) reliably hitting 9/9.

The CLI / engine / daemon are not at fault. We confirmed by running
the same prompts through `coherence-gate.sh`'s daemon-direct JSONL
path: same model, same temp, same backend → no halts, because
coherence-gate uses generous `max_tokens` AND prompts whose answers
don't trigger the long-thinking shape.

### Empirical matrix

All runs on the patched script (`--max-tokens 64` → varies, `--temp
0` → varies). Pure AR (DFlash off) on master at `35265c6`. 9 total
assertions across 6 prompt categories.

| Model | temp | top-p | rp | max | normalize | Pass |
|-------|------|-------|-----|-----|-----------|------|
| 9b | 0 | 0.8 | 1.05 | 64 (orig) | on | **0/9** all empty (think exceeds budget) |
| 9b | 0 | 0.8 | 1.05 | 2048 | on | 7/9 |
| 9b | 0 | 0.8 | 1.05 | 4096 | on | 3/9 (fewer prompts close) |
| 9b | 0.3 | 0.8 | 1.05 | 2048 | on | 7/9 |
| 9b | 0.5 | 0.8 | 1.05 | 2048 | on | 7/9 |
| 9b | 0.5 | 0.8 | 1.05 | 2048 | **off** | **6/9** ↓ |
| 9b | 0.7 | 0.8 | 1.05 | 2048 | on | 7/9 |
| 9b | 0.7 | 0.8 | 1.05 | 2048 | off | 7/9 |
| 9b | 0.7 | 0.9 | 1.05 | 2048 | on | 7/9 |
| 4b | 0.5 | 0.8 | 1.05 | 1500 | on | 7/9 |
| 4b | 0.5 | 0.8 | 1.05 | 1500 | off | 7/9 |
| 4b | 0.7 | 0.9 | 1.05 | 1500 | on | **8/9** (lucky run; primes flaky) |
| 0.8b | 0.5 | 0.8 | 1.05 | 1500 | on | 5/9 (model capability) |

Best result: 4b temp=0.7 + top-p=0.9, **8/9 in a single observed
run, primes the lone fail.** Re-running with the same flags drops
back to 7/9 — the lucky 8th was just which side of the greedy
attractor the model sampled into.

### Per-prompt failure analysis

| # | Prompt | Failure modes seen | Root cause |
|---|--------|---------------------|------------|
| 1 | "Return the first 5 prime numbers, comma-separated. Nothing else." | empty `<think>` (budget exhausted); also `2, 3, 5, 11, 19.` on 0.8b | Greedy thinking-attractor on terse-imperative-with-exactness prompts. Model behaviour, not engine. |
| 2 | "Return numbers 1 to 10, one per line. Nothing else." | empty (think doesn't close); also varies | Same attractor class. |
| 3 | `Return exactly this JSON, nothing else: {"a":1,"b":2}` | model emits ` ```json\n{"a":1,"b":2}\n``` ` | **Suite assertion bug** — `jq -e '. == {...}'` rejects the markdown wrapper. Engine emits valid JSON, suite mis-assesses it. |
| 4 | "Return exactly the number 100. Nothing else, just the number." | reliable PASS | — |
| 5 | "Repeat exactly: alpha beta gamma delta" | empty OR pass | Variance: one run empty, next run "alpha beta gamma delta" — same model + flags. |
| 6a | "Remember secret ZEBRA-9472" → "What is 2+2?" leak check | reliable PASS | — |
| 6b | Same flow, answer to "2+2" | "2" on 0.8b | Model capability. |

Three independent failure modes:

A. **Model thinking-budget exhaustion** (#1, #2 sometimes, #5
   sometimes): need ≥1500 max_tokens to escape, and even then the
   greedy-attractor issue keeps it stochastic.

B. **Suite-assertion strictness mismatch** (#3): `jq` fails on
   well-formed markdown-wrapped JSON. Suite's `elif` fallback also
   uses `jq` so it never actually relaxes the check. This is a fix
   the suite itself needs; no decode setting helps.

C. **Model capability gaps** (#6b on 0.8b, #1 on 0.8b): some
   prompts require model competence the size class can't deliver.

### Why coherence-gate works under the same conditions

`scripts/coherence-gate.sh` runs the same models, same daemon
binary, same `temp=0`. It hasn't been failing. The differences:

- **Generous `max_tokens`** (80 / 180 / 300 sized to model + task).
- **Prompts with short, reasoning-light answers** ("Paris.", `def
  square(n): return n*n`, "9").
- **Daemon-direct JSONL**, no shell-pipe / CLI / TTY surface area.
- **Soft assertions** — looks at coherence in a markdown report,
  doesn't fail on output formatting.

That's the right shape for this category of regression test. PR #20
shares the test intent (#8: tokenizer / output / hallucination /
state-leak) but uses a harness that contradicts every one of those
properties.

### What I tried this session, in case it's instructive

While debugging, I shipped 7 CLI commits chasing this suite's
failures, then reverted all of them because the underlying issue
wasn't anything those commits fixed:

- 3798399 → 2d9c24b → 799c268 → cf2a3d8 → 527617b → 68b32ee →
  b292565 (cli/no_think directive iterations).
- 5533926 (revert chain).
- 90d4035 (one-line surgical fix to stop the bad prose-directive
  injection — actual fix worth keeping).
- 35265c6 (`dflash_mode` default → `off` — discovered DFlash was
  silently on in many of my "pure AR" tests, confounding the
  whole investigation).

The lesson: this suite's failures are not engine-side and not
CLI-side. Greedy + low max_tokens against a thinking model
manifests as "empty stdout" the same way a real engine bug would,
and the categories the suite targets (#8) all happen to be on the
narrow strip of prompts that trip the attractor.

### Recommendation

Closing this PR. Two follow-up paths I'm happy to do (or open
issues for):

1. **Extend `scripts/coherence-gate.sh`** with the 6 prompt
   categories from #8, sized for the per-model think budget,
   daemon-direct, with substring assertions on full output (not
   exact match through `jq`). That gives us regression coverage
   for the same intent without the structural fragility.

2. **Add a `tests/integration/run_suite.sh`** that uses
   `daemon`'s JSONL interface directly, mirrors the
   coherence-gate format, and runs as part of CI on a small set of
   models. The shell-pipe-grep approach wouldn't survive future
   CLI refactors anyway — every directive change in the CLI
   filter chain has the potential to break a grep-based assertion
   downstream.

Notes left in the repo memory store so future sessions / agents
don't repeat the chase:
- `memory/feedback_no_think_directive_loops.prd` — do-not-retry
  contract for any `/no_think` injection at the CLI layer.
- `memory/feedback_quality_gate_baselines_degenerate.md` —
  greedy-degenerate property of Qwen3.x.
- (To-add) `memory/feedback_pr20_suite_radioactive.prd` — this
  comment archived, with the explicit instruction "do not try to
  make this suite pass; convert to coherence-gate format
  instead".

Sorry again for the long delay — the test categories themselves
are valuable, the harness just isn't.

Kaden-Schutt:
Closing per the matrix above. The test intent is good — converting to a coherence-gate extension is the right path. Tracking as a follow-up.

Issue #21, opened 2026-04-15
Vision encoder: performance, color accuracy, and WMMA correctness issues


nickfinease:
## Summary

This issue tracks multiple problems discovered while testing Qwen3.5 VL inference via the OpenAI-compatible HTTP API on gfx1100 (RX 7900 XTX). The vision encoder works end-to-end but has significant performance and accuracy issues compared to llama.cpp's Qwen3-VL implementation.

**Environment:** Arch Linux, kernel 6.19.11, ROCm 6.4, gfx1100, hipfire v0.1.6

---

## 1. gemm_f16 is unoptimized (performance)

The vision encoder's `linear_f16()` in `crates/engine/src/qwen35_vl.rs` calls `gemm_f16` — a 28-line naive kernel with:
- No shared memory / LDS
- No tiling or register blocking
- One threadblock per output element, 32 threads
- No WMMA/MFMA instructions

This is in stark contrast to the text model which uses heavily optimized `gemm_hfq4g256_wmma` kernels with tiling, shared memory, and hardware matrix multiply.

**Impact:** Vision forward takes ~5-6s vs ~0.6s in llama.cpp (Qwen3-VL-8B-Q4_K_M). The text generation TPS advantage of hipfire is negated by the slow vision pass for document/PDF processing workloads.

**Shapes processed (784 patches × 27 layers):**
- Patch embed: `[784, 1536] @ [1536, 1152]`
- QKV: `[784, 3456] @ [3456, 1152]`
- Proj: `[784, 1152] @ [1152, 1152]`
- FC1: `[784, 4304] @ [4304, 1152]`
- FC2: `[784, 1152] @ [1152, 4304]`

**Potential fix:** Fused `gemm_f16_bias_transpose` kernel that does GEMM + bias + transpose in one pass, eliminating two kernel launches and two full-tensor memory roundtrips per linear layer.

---

## 2. gemm_f16_wmma produces incorrect output (correctness)

The existing WMMA kernel at `kernels/src/gemm_f16_wmma.hip` (100 lines) uses `__builtin_amdgcn_wmma_f32_16x16x16_f16_w32` but has no shared memory tiling — every K-iteration reads from global memory.

More critically, the WMMA kernel produces **garbage output** (all `!` characters) when used for vision forward. It works correctly in some runs but fails in others, suggesting a race condition or register layout issue.

**Reproduction:**
```rust
// In linear_f16(), WMMA path sometimes produces garbage:
[...]

PR #22, opened 2026-04-16
Vision encoder: HFQ4 quantization, sync optimization, 1.5x speedup

nickfinease:
## Summary

Optimizes the vision encoder for 1.5x faster inference and 47% smaller model files by adding HFQ4 quantization support for vision weights.

**Benchmark (qwen3.5-9b-vl, 448px image, 256 max_tokens, gfx1100):**

- F16 vision (baseline): 10.4 GB, 7.5s warm
- HFQ4 vision: 5.5 GB, 5.0s warm, 1.5x speedup

### 1. `--vision-quant hfq4` flag (hipfire-quantize)
- New CLI flag: `--vision-quant <format>` (currently supports `hfq4`)
- When used with `--include-vision`, vision weights are quantized to HFQ4G256/HFQ4G128 instead of stored as F16
- Reduces vision weight size by ~4x (912MB to 228MB)

### 2. HFQ4 dequant support (engine)
- `load_f16_gpu` and `load_f32_gpu` now handle qt=6 (HFQ4G256) and qt=7 (HFQ4G128)
- Uses `info.group_size` from the HFQ file for correct block size (not assumed from qt)
- Dequantizes to F32 on CPU, then converts to F16 for `gemm_f16` kernel compatibility
- New `dequant_hfq4()` function with bounds-safe block parsing

### 3. Sync barrier reduction (engine)
- Removed `device_synchronize()` every 9 layers (3 syncs to 1 at end of vision forward)
- ~0.2s improvement, reduces GPU pipeline stalls

### 4. GEMM kernel dispatch stubs (rdna-compute)
- Added `gemm_f16_tiled` - ILP-unrolled, no shared memory (correct but no faster than naive)
- Added `gemm_f16_bias` - fused GEMM+transpose+bias in one kernel (correct but fewer blocks reduces parallelism)
- Added `vit_attention_opt` - tiled K/V attention (correct but L2 cache already handles reuse)
- These are preserved for future optimization work, not currently called by the vision encoder

## Usage

```
hipfire-quantize --input ~/models/Qwen3.5-9B \
  --output qwen3.5-9b-vl-hfq4.mq4 \
  --format mq4g256 --include-vision --vision-quant hfq4
```

## What was tried and reverted

- Fused GEMM+bias+transpose: 7.3s vs 6.9s (fewer blocks = less GPU parallelism)
- Tiled GEMM (ILP unroll): same speed (L2 cache already handles reuse)
- Shared memory GEMM (LDS): slower (8KB LDS kills occupancy)
- Tiled attention (K_TILE=64): 8.2s (22KB LDS + sync overhead)
- WMMA kernel: garbage output (register layout bug under some conditions)

Key finding: For this workload (784 patches, K=1152-4304), all data fits in the GPU L2 cache. Software tiling with shared memory adds load+sync overhead that outweighs the benefit.

## Remaining issues (tracked in #21)

- Color accuracy (green to purple) - not quantization, not preprocessing
- Vision GEMM kernel - future optimization: on-the-fly HFQ4 dequant on GPU
- WMMA correctness - gemm_f16_wmma produces garbage under some conditions

## Files changed

- crates/engine/src/qwen35_vl.rs - vision weight loaders + dequant + sync reduction
- crates/rdna-compute/src/dispatch.rs - new GEMM/attention dispatch stubs
- crates/rdna-compute/src/kernels.rs - new kernel constants
- crates/hipfire-quantize/src/main.rs - --vision-quant flag
- kernels/src/gemm_f16_tiled.hip - ILP-unrolled GEMM kernel
- kernels/src/gemm_f16_bias.hip - fused GEMM+bias kernel
- kernels/src/gemm_f16_wmma.hip - WMMA GEMM kernel (has correctness bug)
- kernels/src/vit_attention_opt.hip - tiled attention kernel
- kernels/src/vit_attention_flash.hip - flash attention prototype

## Known accuracy impact

HFQ4 vision weights (4-bit) introduce small per-weight errors that compound through 27 transformer layers. Diagnosed via intermediate tensor comparison against HuggingFace's reference implementation:

- **patch_embed output**: Matches HuggingFace exactly (diff < 0.0001)
- **pos_embed output**: Matches exactly
- **After layer 0**: Diverges — HFQ4 produces 10-30% relative error on individual feature values
- **After 27 layers**: Error compounds into shifted embedding magnitudes

Effect on different vision attributes:

| Attribute | Impact | Why |
|---|---|---|
| Structure (shape, layout, text) | Minimal | Attention patterns are robust — softmax normalizes away small weight errors |
| Spatial relationships | Minimal | Encoded in attention, not magnitudes |
| Color | Noticeable shift | Color is encoded in feature magnitudes which are sensitive to quantization noise |

For color-critical workloads, use the default F16 vision weights (omit `--vision-quant`). For document/PDF processing where structure matters more than color accuracy, HFQ4 gives 1.5x speedup with acceptable quality.

**Diagnostic method**: Compared `patch_embed` output, `pos_embed` output, and layer 0/1/2 outputs between hipfire and a manual PyTorch implementation loading original BF16 safetensors weights. Values match through `pos_embed`, diverge at layer 0 where HFQ4G128-quantized block weights introduce per-group rounding errors.

## Testing

Verified on gfx1100 (RX 7900 XTX), ROCm 6.4, kernel 6.19.11:
- F16 vision model still works (no regression)
- HFQ4 vision model produces valid output
- HFQ4 model is 1.5x faster than F16
- HFQ4 model file is 47% smaller
- Text model unaffected (changes are vision-only)

Ref: https://github.com/Kaden-Schutt/hipfire/issues/21

### DIFF
--- a/crates/engine/examples/daemon.rs
+++ b/crates/engine/examples/daemon.rs
@@ -970,16 +970,18 @@ fn generate_vl(m: &mut LoadedModel, gpu: &mut rdna_compute::Gpu, stdout: &mut st
     let kv = m.kv_cache.as_mut().unwrap();
     let dn = m.dn_state.as_mut().unwrap();
 
-    // Load and preprocess image
-    let pixels = engine::image::load_and_preprocess(Path::new(image_path), IMAGE_SIZE);
-    let grid_h = IMAGE_SIZE / vision_config.patch_size;
-    let grid_w = IMAGE_SIZE / vision_config.patch_size;
+    // Load and preprocess image (smart resize matching HuggingFace)
+    eprintln!("[VL-DEBUG] preprocessing image: {}", image_path);
+    let (pixels, img_h, img_w) = engine::image::load_and_preprocess(Path::new(image_path), vision_config.patch_size);
+    eprintln!("[VL-DEBUG] preprocessed: {}x{}", img_w, img_h);
+    let grid_h = img_h / vision_config.patch_size;
+    let grid_w = img_w / vision_config.patch_size;
     let n_patches = grid_h * grid_w;
     let n_visual_tokens = n_patches / (vision_config.spatial_merge_size * vision_config.spatial_merge_size);
 
     // Extract patches and run vision encoder
     let patches = engine::image::extract_patches(
-        &pixels, 3, IMAGE_SIZE, IMAGE_SIZE,
+        &pixels, 3, img_h, img_w,
         vision_config.patch_size, vision_config.temporal_patch_size,
     );
     let visual_tokens = qwen35_vl::vision_forward(gpu, vision_weights, vision_config, &patches, grid_h, grid_w)
--- a/crates/engine/examples/infer.rs
+++ b/crates/engine/examples/infer.rs
@@ -87,13 +87,13 @@ fn main() {
         eprintln!("Vision: hidden={}, layers={}, heads={}", vision_config.hidden_size, vision_config.num_layers, vision_config.num_heads);
 
         let img = image_path.as_ref().unwrap();
-        let pixels = engine::image::load_and_preprocess(Path::new(img), IMAGE_SIZE);
-        let grid_h = IMAGE_SIZE / vision_config.patch_size;
-        let grid_w = IMAGE_SIZE / vision_config.patch_size;
+        let (pixels, img_h, img_w) = engine::image::load_and_preprocess(Path::new(img), vision_config.patch_size);
+        let grid_h = img_h / vision_config.patch_size;
+        let grid_w = img_w / vision_config.patch_size;
         n_visual_tokens = (grid_h * grid_w) / (vision_config.spatial_merge_size * vision_config.spatial_merge_size);
 
         let patches = engine::image::extract_patches(
-            &pixels, 3, IMAGE_SIZE, IMAGE_SIZE,
+            &pixels, 3, img_h, img_w,
             vision_config.patch_size, vision_config.temporal_patch_size,
         );
 
--- a/crates/engine/examples/infer_vl.rs
+++ b/crates/engine/examples/infer_vl.rs
@@ -51,16 +51,16 @@ fn main() {
 
     // Load and preprocess image
     eprintln!("Preprocessing image...");
-    let pixels = engine::image::load_and_preprocess(Path::new(image_path), IMAGE_SIZE);
-    let grid_h = IMAGE_SIZE / vision_config.patch_size;
-    let grid_w = IMAGE_SIZE / vision_config.patch_size;
+    let (pixels, img_h, img_w) = engine::image::load_and_preprocess(Path::new(image_path), vision_config.patch_size);
+    let grid_h = img_h / vision_config.patch_size;
+    let grid_w = img_w / vision_config.patch_size;
     let n_patches = grid_h * grid_w;
     let n_visual_tokens = n_patches / (vision_config.spatial_merge_size * vision_config.spatial_merge_size);
-    eprintln!("Image: {}x{} → {}x{} patches → {} visual tokens", IMAGE_SIZE, IMAGE_SIZE, grid_h, grid_w, n_visual_tokens);
+    eprintln!("Image: {}x{} → {}x{} patches → {} visual tokens", img_w, img_h, grid_h, grid_w, n_visual_tokens);
 
     // Extract patches for vision encoder
     let patches = engine::image::extract_patches(
-        &pixels, 3, IMAGE_SIZE, IMAGE_SIZE,
+        &pixels, 3, img_h, img_w,
         vision_config.patch_size, vision_config.temporal_patch_size,
     );
 
--- a/crates/engine/src/image.rs
+++ b/crates/engine/src/image.rs
@@ -18,7 +18,7 @@ pub fn load_and_preprocess(path: &Path, target_size: usize) -> Vec<f32> {
     let rgb = img.to_rgb8();
     let (w, h) = (rgb.width() as usize, rgb.height() as usize);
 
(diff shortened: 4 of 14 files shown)


Kaden-Schutt (changes requested):
Thanks for the substantial work on the vision encoder path — the HFQ4 weight quantization idea is a real win and the smart_resize change is valuable. However, a deep code review against the current master (`7e4ddb1`) surfaces several blockers that need to be addressed before merge.

## Blockers

### 1. Four of the six "dispatch stubs not currently called" kernels are actually wired and callable

The PR body states:
> These are preserved for future optimization work, not currently called by the vision encoder

That is accurate only for `gemm_f16_wmma_tiled.hip` and `vit_attention_flash.hip`. Inspection of `crates/rdna-compute/src/dispatch.rs` shows four of the six files have active `ensure_kernel` dispatches:

- `gemm_f16_wmma` — `dispatch.rs:7868` (wired, callable)
- `gemm_f16_tiled` — `dispatch.rs:7895` (wired, callable)
- `gemm_f16_bias` — `dispatch.rs:7922` (wired, callable)
- `vit_attention_opt` — `dispatch.rs:8090` (wired, callable)

This repo recently shipped PR #32 specifically to delete unreachable WMMA kernels that still had the wrong RDNA3 output mapping — exactly the "copy-paste hazard" pattern. A new kernel file with a self-admitted correctness bug (`gemm_f16_wmma.hip`, per this PR's own body) that's actively dispatched is an automatic blocker. Either:

- Demonstrate the bug is real with a concrete reproducer and fix it, or
- Remove `gemm_f16_wmma.hip` + its `ensure_kernel` call if the claim is wrong, or
- Confirm the claim is retracted — `gemm_f16_wmma.hip` lines 89–97 actually use the **correct** RDNA3 C-mapping (`out_row = row_start + 2*j + (tid >> 4)`, `out_col = col_start + (tid & 15)`) per the validated formula from the WMMA correctness fix. If there's no bug, delete the claim from the PR body.

For `gemm_f16_wmma_tiled.hip` and `vit_attention_flash.hip` — truly dead. Please delete both files.

### 2. Color regression is NOT opt-in — `load_f16_gpu` auto-routes on qt=6/7 regardless of CLI flag

`crates/engine/src/qwen35_vl.rs:112` routes qt=6 (HFQ4G256) and qt=7 (HFQ4G128) to the new HFQ4 dequant path for **all callers**, not just those who pass `--vision-quant hfq4` at quantize time. The PR body's framing ("For color-critical workloads, use the default F16 vision weights — omit `--vision-quant`") only holds if users load a separately-supplied F16 model bundle. Any user loading an integrated HFQ4-quantized bundle hits the acknowledged 10–30% layer-0 feature error → compounding green→purple shift.

Please either:
- Gate the `load_f16_gpu` qt=6/7 path behind an explicit runtime check/flag, or
- Document prominently in the CHANGELOG/README that the HFQ4 vision path has a known color regression for any bundle containing qt=6/7 vision tensors.

### 3. Rebase conflicts

Confirmed via `git merge-tree master HEAD`: real conflict in `crates/rdna-compute/src/dispatch.rs`. Master has moved through PR #28 (batched prefill determinism, touches the vision dispatch region), PR #32 (dead WMMA cleanup in `kernels.rs`), PR #33 (qkvza FP32 dequant + delegation), PR #34 (registry wire-up). Please rebase against current master.

### 4. `.githooks/pre-commit` is the pre-coherence-gate version

This PR's pre-commit hook (line 40) invokes `quality-gate.sh`. Master replaced this with `coherence-gate.sh` (see the 2026-04-18 transition). A naive merge will revert that migration. On rebase, keep master's version.

### 5. `load_f16_gpu` shape encoding bug

`crates/engine/src/qwen35_vl.rs:118` calls `gpu.upload_raw(&f16_bytes, &[f16_bytes.len()])` — that's **bytes** (`n * 2`) not **element count**. Inconsistent with `upload_f32`'s contract. The naive `gemm_f16` doesn't re-read the shape field so this doesn't corrupt computation, but any code path that reads `.shape[0]` back to recover the element count will silently misbehave. Please pass `&[n]`.

## Positive notes (worth keeping)

- HFQ4 dequant math is correct and matches the GPU kernel convention (`gemv_hfq4g128.hip:33-36`: `scale * nibble + zero`) with correct block layout (G256: 136B, G128: 72B).
- `smart_resize` `max_pixels=1003520` matches HuggingFace Qwen2VLImageProcessorFast's value — though please reconcile the comment `14 * 14 * 4 * 1280` with the upstream factorization `28 * 28 * 256 * 5` for clarity.
- Sync reduction (27 per-layer `device_synchronize` → 1 terminal) is safe given in-order stream dispatch and is a legitimate win.
- `fix(vision): use group_size from file for HFQ4 dequant` (commit `f4cf567`) correctly addresses the earlier cumulative-offset hazard.

## What I'd recommend for the next iteration

1. Delete truly-dead files: `gemm_f16_wmma_tiled.hip`, `vit_attention_flash.hip`
2. Either fix or retract the `gemm_f16_wmma` "correctness bug" claim (the output mapping inspection says the kernel is fine)
3. Gate the HFQ4 vision load path behind an explicit opt-in, or document the regression scope accurately
4. Fix the shape encoding on `load_f16_gpu:118`
5. Rebase on current master, keeping the master `coherence-gate.sh` pre-commit
6. Run the coherence gate against a text-only Qwen3.5-VL prompt post-rebase to confirm no regression on the non-vision path

Color root-cause is the real prize (#23). Given the acknowledged 10–30% layer-0 feature error, the right path forward may be to ship the `--vision-quant hfq4` flag + smart_resize as a smaller, targeted PR and hold the dispatch-stub kernels for a separate PR once they're actually optimizing something.

Kaden-Schutt:
Superseded by #35, which rebased this work onto current master, addressed the review feedback here, and additionally landed the root-cause fix for #23 (G↔B preprocessing swap, with regression test). Merged as 54950c1.

PR #24, opened 2026-04-17
feat(qwen3.6): add Qwen3.6-35B-A3B MoE + fix hf download

nickfinease:
## Summary

- Adds `qwen3.6:35b-a3b` (and four aliases) to the CLI model registry. The new Qwen3.6-35B-A3B release declares `model_type: qwen3_5_moe` in its HF config and is architecturally identical to Qwen3.5-35B-A3B (256 experts top-8, hybrid DeltaNet + Full Attention at 3:1, head_dim=256 with partial_rotary_factor=0.25, shared expert), so the engine's existing `arch_id=6` path loads it without source changes. The quantizer's default `--include-vision=false` cleanly skips the multimodal tensors for a text-only `.mq4`.
- Fixes the quantizer's auto-download: `huggingface-cli` is a no-op deprecation stub in `huggingface_hub >= 1.0`, which silently broke `hipfire-quantize --input Owner/Model` on fresh systems. Now prefers the new `hf` CLI with a graceful fallback so older installs keep working.

## Verification (gfx1100 / RX 7900)

End-to-end tested with the model quantized from `Qwen/Qwen3.6-35B-A3B`:

- Output: `18,683.9 MB` `.mq4` (`34,660,433,920` params quantized, `1,291,212,016` visual/mtp skipped)
- `examples/a3b_smoke_forward`: logits finite `[-10.53, 14.73]`, `0` NaN / `0` Inf
- Prefill `~155 tok/s`, generation `~150 tok/s`
- Simple prompts produce coherent output end-to-end:
  ```
  $ infer qwen3.6-35b-a3b.mq4 "Say hello."
  <think></think>

  Hello!<|im_end|>
  ```

Known follow-up (not addressed here): complex thinking-mode prompts can enter repetitive loops. This behavior is not specific to Qwen3.6 — it reproduces the pre-existing Qwen3.5-thinking failure mode and is orthogonal to getting the model loading/running.

## Test plan

- [x] `cargo build --release -p hipfire-quantize`
- [x] `cargo build --release --features deltanet --example a3b_smoke_forward --example infer -p engine`
- [x] Quantize `Qwen/Qwen3.6-35B-A3B` → `qwen3.6-35b-a3b.mq4` (text-only)
- [x] `a3b_smoke_forward` — no NaN/Inf, sane logit range
- [x] `infer` — coherent output on simple prompt
- [ ] Repro on gfx1030 / gfx1200 cards (no access locally)
[...]

PR #34, opened 2026-04-19
feat(cli): wire qwen3.6:35b-a3b to HF + drop LOCAL ONLY

Kaden-Schutt:
## Summary

Wires `qwen3.6:35b-a3b` to the freshly-uploaded [schuttdev/hipfire-qwen3.6-35b-a3b](https://huggingface.co/schuttdev/hipfire-qwen3.6-35b-a3b). `hipfire pull qwen3.6:35b-a3b` now resolves instead of exiting with LOCAL-ONLY guidance.

## HF repo contents

| File | Size | Purpose |
|------|------|---------|
| `qwen3.6-35b-a3b.mq4` | 18.7 GB | Main MoE model (MQ4, FWHT-rotated 4-bit) |
| `qwen3.6-35b-a3b.mq4.hermes.triattn.bin` | 983 KB | Aureth-corpus CASK sidecar for long-context eviction |
| `README.md` | — | Usage + sidecar wiring instructions |

## Why only the Hermes sidecar, not the wikitext one?

Per user directive — the standard wikitext-calibrated `<model>.mq4.triattn.bin` underperforms in practice despite the 0.1.7-alpha benchmark parity claim. Only the Aureth-corpus-trained `.hermes.triattn.bin` ships. (Memory: `feedback_wikitext_triattn_sidecar_garbage`.)

## Pull flow

```bash
hipfire pull qwen3.6:35b-a3b
hipfire run qwen3.6:35b-a3b "Write a Rust function that parses ISO-8601."
```

The sidecar is NOT auto-downloaded — existing convention is that sidecars are user-managed via `hipfire config <tag> set cask_sidecar <path>`. README on the HF repo walks through the optional wiring.

## Scope

- `qwen3.5:35b-a3b` still has `repo: ""` (LOCAL ONLY) — separate upload needed if/when desired.
- No engine changes — `arch_id=6` path already handles Qwen3.6 unchanged per the 0.1.7-alpha.3 norm-fix line (`1e01c0b`).

## Test plan

- [x] HF repo created and all 3 files uploaded
- [x] Download URL resolves (302 → CDN) — verified via `curl -sI`
- [ ] `hipfire pull qwen3.6:35b-a3b` on a fresh cache — not tested; registry wire-up is trivial and the HF repo is live

Closes (supersedes) #24.

### DIFF
--- a/cli/index.ts
+++ b/cli/index.ts
@@ -401,9 +401,10 @@ const REGISTRY: Record<string, ModelEntry> = {
   // the existing A3B (arch_id=6) code path; weights differ (coding/agentic
   // fine-tune) but architecture (256 experts top-8, hybrid DeltaNet+FA,
   // head_dim=256 with partial_rotary=0.25, shared expert) is identical.
-  // Local-only until a pre-quantized HF repo exists; use `hipfire quantize
-  // Qwen/Qwen3.6-35B-A3B --format mq4` to produce the file.
-  "qwen3.6:35b-a3b":  { repo: "", file: "qwen3.6-35b-a3b.mq4", size_gb: 18.7, min_vram_gb: 22, desc: "MoE 35B/3B-active (coding/agent) — LOCAL ONLY, quantize Qwen/Qwen3.6-35B-A3B" },
+  // Pre-quant at schuttdev/hipfire-qwen3.6-35b-a3b; includes an Aureth-
+  // corpus Hermes sidecar (qwen3.6-35b-a3b.mq4.hermes.triattn.bin) for
+  // cask_sidecar-based KV eviction — download separately via `hf download`.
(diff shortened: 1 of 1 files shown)


ForestJohnson:
It looks like this got rolled back when the repo configs were moved into a JSON file, is that intentional ?

https://github.com/Kaden-Schutt/hipfire/blob/645ddef95b0cbc12390d14e46294cad5d428882d/cli/registry.json#L10

Kaden-Schutt:
@ForestJohnson — good catch, fixed in master at `893abe1`. The JSON migration dropped the repo field on the qwen3.6:35b-a3b row; restored to `schuttdev/hipfire-qwen3.6-35b-a3b`. `hipfire pull qwen3.6:35b-a3b` should resolve again on a fresh build. Thanks for noticing.

Issue #54, opened 2026-04-27
9070 XT (gfx1201)


0c33:
~/.hipfire/bin
❯ ./hipfire run qwen3.5:0.8b "What is the capital of France?"
Model not found locally. Pulling qwen3.5:0.8b...
Pulling qwen3.5:0.8b (0.55GB)...
  https://huggingface.co/schuttdev/hipfire-qwen3.5-0.8b/resolve/main/qwen3.5-0.8b.mq4
  549/549 MB (100.0%)
  Saved: /home/sami/.hipfire/models/qwen3.5-0.8b.mq4 (0.5GB)
GPU: gfx1201 (17.1 GB VRAM, HIP 7.2)
  pre-compiled kernels: .hipfire_kernels/gfx1201
[hipfire] DFlash disabled (dflash_mode=off).
  loading token_embd...
    (Q8_0 raw, 270 MB)
  loading output_norm...
  loading output (tied embeddings, qt=3)...
  loading layer 0/24 (LinearAttention)...
  loading layer 1/24 (LinearAttention)...
  loading layer 2/24 (LinearAttention)...
  loading layer 3/24 (FullAttention)...
  loading layer 4/24 (LinearAttention)...
  loading layer 5/24 (LinearAttention)...
  loading layer 6/24 (LinearAttention)...
  loading layer 7/24 (FullAttention)...
  loading layer 8/24 (LinearAttention)...
  loading layer 9/24 (LinearAttention)...
  loading layer 10/24 (LinearAttention)...
  loading layer 11/24 (FullAttention)...
  loading layer 12/24 (LinearAttention)...
  loading layer 13/24 (LinearAttention)...
  loading layer 14/24 (LinearAttention)...
  loading layer 15/24 (FullAttention)...
  loading layer 16/24 (LinearAttention)...
  loading layer 17/24 (LinearAttention)...
  loading layer 18/24 (LinearAttention)...
  loading layer 19/24 (FullAttention)...
  loading layer 20/24 (LinearAttention)...
  loading layer 21/24 (LinearAttention)...
  loading layer 22/24 (LinearAttention)...
  loading layer 23/24 (FullAttention)...
KV cache: asym3 (K rotated-3b 100B + V Q8 272B = 372 B/head, 5.5x vs fp32, physical_cap=32768 / max_seq=32768)
[qwen3_5] 1024d 24L 248320 vocab

thread 'main' (9768) panicked at crates/engine/examples/daemon.rs:1385:15:
[...]

PR #56, opened 2026-04-27
Port gfx12 arch wmma + channel tests

RobinVanCauter:
Came across issue #54 when I was looking if there was any support for my R9700 yet. Thanks to the very detailed instructions and skill definitions, it was relatively low effort for me to run through it and have it validate against my GPU. Hope it helps!

If there's anything you'd like to see or approach different (for example; squash some of the commits, etc), let me know!

### DIFF
--- a/crates/engine/examples/test_wmma_gate_up_gfx12.rs
+++ b/crates/engine/examples/test_wmma_gate_up_gfx12.rs
@@ -1,0 +1,240 @@
+//! Channel-test for the gfx12 (RDNA4) WMMA gate_up scaffold.
+//!
+//! Sister to test_wmma_qkv_gfx12 — same harness pattern, applied to the
+//! 2-output `gemm_gate_up_hfq4g256_wmma_gfx12` kernel. Compares against
+//! the validated `gemm_gate_up_hfq4g256_dot2` reference on synthetic
+//! HFQ4G256 weights + f16 X.
+//!
+//! Bails on non-gfx12 (the `_w32_gfx12` builtin doesn't exist on gfx11).
+//!
+//! Run: cargo run --release --features deltanet -p engine \
+//!         --example test_wmma_gate_up_gfx12
+
+use rdna_compute::{DType, Gpu};
+
+fn main() {
+    let mut gpu = Gpu::init().expect("GPU init failed");
+    let arch = gpu.arch.clone();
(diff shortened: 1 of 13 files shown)


Kaden-Schutt:
@RobinVanCauter Thanks — this is exceptional work. The skill was authored hoping someone exactly like you would walk through it; you did, and the round-trip from "issue #54 reported" → "fully validated 5-kernel port" took less than a day. That's the best possible outcome for the skill itself.

A few notes on what landed well, for the record:

- **Channel-test on real R9700 is the load-bearing correctness gate** per the skill (`validation.md`), and your tests pass for all 6 kernels with a tight tolerance plus a row-mod-16 histogram diagnostic that would have caught the C-mapping silently. The earlier 6-week gfx11 silent-corruption bug (`b7ac66a`) is the cautionary tale you avoided.
- **The C-mapping hypothesis I encoded as `acc[j] = C[8*(tid>>4) + j][tid & 15]`** in the canonical (`6924f2a`) was speculative, derived from CK trait math. Your channel-tests **validated it on hardware** — which is the answer to a real open question. I'll tighten the "HYPOTHESIS pending channel-test" comment in `kernels.rs` post-merge.
- **Conservative wiring** — adding `gemm_*_wmma_gfx12` methods to `Gpu` but **not routing them through the public dispatch tree** is exactly the right call until a maintainer reviews and runs perf on a wired-up build. No current user behavior changes when this merges.
- **HFQ6 was real engineering**, not a mechanical fork — the `200B groups, 4-byte unaligned reads at byte-offsets {0, 3} per K half-tile` pattern is the kind of thing that would have stalled a less-careful contributor.

**Plan to land**: merging shortly. Two follow-ups, neither blocking this PR:

1. **Wire the dispatch + measure perf vs dot2** on R9700 — opening as a separate issue. The PR doesn't need to wait on this; once we have numbers, we either flip the dispatch (likely the WMMA path beats dot2 ~2× on prefill GEMM as it does on gfx11) or, if the win is smaller than expected, request ISA-targeted adjustments (multi-row variants, K-split tuning, occupancy work) per the standard `.skills/hipfire-arch-port/` workflow.
2. The stale HYPOTHESIS comment cleanup mentioned above.

Both are tractable for any contributor with R9700 access — yours, mine via remote when one's available, or the next person who shows up on issue #54. Welcome aboard, and please file an issue / PR for anything else you want to dig into.

PR #59, opened 2026-04-27
fix strix halo gfx1151 autodetect

KotDath:
On Strix Halo / Radeon 8060S, KFD exposes the GPU as:
```
  gfx_target_version 110501
```
  This should resolve to:

```
  gfx1151
```

  However, the installer did not recognize 110501, and the CLI's generic conversion logic could produce an incorrect arch string instead of gfx1151.

  ## Impact

  This affects Strix Halo setup and diagnostics:

  - scripts/install.sh can fail to auto-detect the GPU arch and fall back to unknown.
  - CLI sysfs-based detection can derive the wrong gfx* string.
  - The wrong arch can affect config defaults, kernel blob lookup/copy, and install/update flows.
  - Strix Halo should use the existing gfx1151 profile, including the APU-oriented asym2 KV default.

  ## Fix

  I added explicit KFD gfx_target_version mappings in the CLI and installer:

  - 110501 -> gfx1151
  - preserved existing mappings for gfx1010, gfx1030, gfx1100, gfx1200, and gfx1201

  Also updated the installer supported-GPU message to include:

  gfx1151 (Strix Halo)

  Branch with the fix:

  codex/strix-halo-gfx1151-autodetect

  Commit:

  5293308 fix strix halo gfx1151 autodetect

  ## Validation

  Tested on AMD Ryzen AI Max+ 395 / Radeon 8060S:

  rocminfo: gfx1151 / Radeon 8060S Graphics
  KFD sysfs: gfx_target_version 110501
  hipfire diag: GPU arch gfx1151

  Checks run:

  bash -n scripts/install.sh
  gfx target version mapping OK
  hipfire diag

  Smoke tests also passed locally on Strix Halo with ROCm 7.2:

  qwen3.5:0.8b runs
  qwen3.5:9b runs
  qwen3.5:9b-draft loads
  DFlash code prompt test runs

### DIFF
--- a/cli/index.ts
+++ b/cli/index.ts
@@ -532,6 +532,25 @@ function downloadUrl(entry: ModelEntry): string {
 }
 
 // ─── GPU arch detection + per-arch defaults ──────────────
+function gfxTargetVersionToArch(ver: number): string {
+  const known: Record<number, string> = {
+    100100: "gfx1010",
+    100300: "gfx1030",
+    100302: "gfx1030",
+    110000: "gfx1100",
+    110001: "gfx1100",
+    110501: "gfx1151",
+    120000: "gfx1200",
+    120001: "gfx1201",
+  };
+  if (known[ver]) return known[ver];
+
+  const major = Math.floor(ver / 10000);
+  const minor = Math.floor((ver % 10000) / 100);
+  const step = ver % 100;
+  return `gfx${major}${minor}${step}`;
+}
+
 function detectGpuArch(): string {
   // Read KFD sysfs for GPU arch (same as install command)
   for (const node of ["1", "0"]) {
@@ -539,12 +558,7 @@ function detectGpuArch(): string {
       const props = require("fs").readFileSync(`/sys/class/kfd/kfd/topology/nodes/${node}/properties`, "utf8");
       const m = props.match(/gfx_target_version\s+(\d+)/);
       if (m) {
-        const ver = parseInt(m[1]);
-        const major = Math.floor(ver / 10000);
-        const minor = Math.floor((ver % 10000) / 100);
-        const step = ver % 100;
-        let arch = `gfx${major}${minor.toString().padStart(2, '0')}${step || '0'}`;
-        return arch.replace(/^(gfx\d{4})0$/, '$1');
+        return gfxTargetVersionToArch(parseInt(m[1]));
       }
     } catch {}
   }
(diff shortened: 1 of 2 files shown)


Kaden-Schutt:
Merged — thanks @KotDath. Two things this fix is doing well that we want more of:

- **Hardware-validated end-to-end**: rocminfo arch check + `hipfire diag` arch check + actual model loads (0.8b, 9b, 9b-draft) + DFlash code prompt. That's the bar.
- **Refactor cleanup beyond the bug**: the duplicated `gfx_target_version` decoder in cli/index.ts had a latent bug for any arch with a non-zero step byte (e.g. `100302 → gfx1003` instead of `gfx1030`). Consolidating into one helper + explicit known-version table fixes a class of bugs, not just gfx1151.

**Two follow-ups, neither blocking, both natural for someone with Strix Halo hardware:**

1. **Speed-baseline contribution**: tracked separately at #61. One command to bootstrap (`HIPFIRE_BASELINE_ARCH=gfx1151 ./scripts/speed-gate.sh --update-baselines`), tiny PR. Without it the speed-gate refuses to run on gfx1151, so future kernel-touching commits won't get caught if they regress your hardware.

2. **If you want to dig further** on Strix Halo perf — the new [`.skills/hipfire-kernel-tuning/`](https://github.com/Kaden-Schutt/hipfire/tree/master/.skills/hipfire-kernel-tuning) skill (just shipped this session) is the canonical entry point for kernel-level optimization work. It indexes the levers this codebase has actually used (multi-row, K-tile depth, wave64 port, prefetch, fused projections, ISA flags) and codifies the bench discipline (clean-baseline bisect, prompt-md5, three gates) that catches fake wins. Read [`playbook.md`](https://github.com/Kaden-Schutt/hipfire/blob/master/.skills/hipfire-kernel-tuning/playbook.md) before touching kernels — the within-session A/B noise band is ±10–15% on RDNA3, and "+8% wins" measured in one shell session are inside the noise.

For non-perf contributions (model bringup, bug reports, install issues) the [`.skills/hipfire-tester/`](https://github.com/Kaden-Schutt/hipfire/tree/master/.skills/hipfire-tester) and [`.skills/hipfire-autoheal/`](https://github.com/Kaden-Schutt/hipfire/tree/master/.skills/hipfire-autoheal) skills are the lighter-weight on-ramps.

Welcome to the contributors list — file an issue / PR for anything else you hit on Strix Halo.

Issue #60, opened 2026-04-27
Prefill scaling regression vs llama.cpp at pp≥512 on 9B+ (help wanted)


Kaden-Schutt:
**Status**: open invitation, not blocked on hardware. Anyone with a 7900 XTX (or any RDNA3 / RDNA4) can dig in. The diagnostic phase needs no kernel writing — just `hipfire bench` runs + reading the profiler output.

## The gap

From the bench-vs-ollama numbers in [docs/BENCHMARKS.md](https://github.com/Kaden-Schutt/hipfire/blob/master/docs/BENCHMARKS.md), 7900 XTX, prefill tok/s:

| Model | hf pp128 | oll pp128 | hf pp512 | oll pp512 | scaling pp128→pp512 (hf) | scaling (oll) |
|---|---:|---:|---:|---:|---:|---:|
| Qwen 3.5 0.8B | **10,861** | 4,622 | **12,962** | 7,117 | +19% | +54% |
| Qwen 3.5 4B | **3,304** | 1,972 | **3,321** | 2,670 | +0.5% | +35% |
| Qwen 3.5 9B | **1,920** | 1,428 | 1,919 | **1,970** | +0% | +38% |

Decode hipfire wins comfortably (1.7–2.1×). Prefill at small contexts hipfire wins decisively. **At pp512 hipfire flatlines while llama.cpp keeps scaling — and on 9B llama.cpp passes us by 2.7%.** 27B at pp512+ is presumably worse but not directly benched here.

The decode lead is great for chat UX. But the prefill flatline is bad for agentic workloads (long tool-call contexts, code review, RAG with packed retrievals).

## Hypotheses for why hipfire's prefill doesn't scale past pp128

Educated guesses, in roughly decreasing-likelihood order. The right first move is profiling, not picking one of these and writing code — the profile tells us which one matters.

1. **K-tile depth too shallow at large M.** Current WMMA prefill kernels (`gemm_qkv_hfq4g256_wmma`, `gemm_hfq4g256_residual_wmma_k2`) use K2 unroll. At pp512 the inner K-loop is 32× larger than at pp128, but the per-iter pipeline depth doesn't grow with it. Deeper unrolls (K4, K8) might hide latency that's exposed at long prefills.
[...]

PR #71, opened 2026-04-27
dispatch(gfx12 residual): wire WMMA kernel — 9B/27B prefill BW gap

RobinVanCauter:
# PR: Port `gemm_hfq4g256_residual` to gfx12 WMMA

This PR ports the `gemm_hfq4g256_residual` kernel (attn-out + ffn-down) to the gfx12 WMMA architecture, addressing the last major un-ported bottleneck following PR #62. 

The single un-ported kernel accounted for ~42-45% of prefill execution time in 9B/27B models. Porting this kernel achieves a ~2.1-2.3× speedup per call.

## Changes
- **Kernels**: Added `kernels/src/gemm_hfq4g256_residual_wmma.gfx12.hip`. Implements the validated `_wmma_k2` variant used in gfx11, utilizing `half8_t` operands and `_w32_gfx12` builtins with preserved K2 unroll.
- **Testing**: Added `crates/engine/examples/test_wmma_residual_gfx12.rs`. Performs a channel-test against the dot2 fp16 reference across seven shapes, specifically validating residual `+=` semantics by pre-seeding output with non-zero patterns.
- **Dispatch**: Updated `gemm_hfq4g256_residual()` to include a gfx12 dispatch branch, gated by the existing `has_wmma_f16_gfx12` predicate.

## Performance (R9700)
Significant improvements observed in prefill workloads:

| Model | Workload | Master | Branch | Δ |
| :--- | :--- | :--- | :--- | :--- |
| 0.8B | pp128 | 5473 | 6434 | +17.5% |
| 4B | pp32 | 1084 | 1366 | +26.0% |
| 4B | pp128 | 1486 | 1866 | +25.5% |
| 9B | pp32 | 620 | 805 | +29.7% |
| 9B | pp128 | 782 | 1027 | +31.3% |
| 27B | pp32 | 150.6 | 192.4 | +27.8% |
| 27B | pp128 | 218.4 | 310.9 | +42.4% |

*Decode performance remains within ±2% of baseline.*

## Validation
- **test_wmma_residual_gfx12**: 7/7 PASS.
- **coherence-gate.sh**: 0.8B / 4B / 9B coherent.
- **speed-gate.sh**: 9/9 metrics passed; baselines updated.
- **Compatibility**: gfx11 code paths remain unchanged via `gfx12*` predicate prefix-matching.

## Future Work
- Porting `gemm_hfq6g256_residual` to gfx12.
- `gemm_hfq4g256_batched_lmhead` gfx12 wiring.
[...]

PR #73, opened 2026-04-27
Add opt-in HFQ4 MMQ prefill path

KotDath:
## Summary

  Adds an opt-in HFQ4-G256 MMQ prefill path for RDNA3/RDNA3.5, currently gated behind `HIPFIRE_MMQ=1`.

  This targets the Strix Halo / gfx1151 prefill bottleneck seen with Qwen3.5 9B HFQ4/MQ4. The new path pre-quantizes prefill activations into a Q8_1 MMQ layout and uses i8 WMMA over 128x128 output/batch tiles,
  similar in shape to llama.cpp's AMD MMQ prompt-processing path.

  This is intentionally **not enabled by default**. I am not fully confident yet that the algorithm is correct enough to ship broadly, so this PR keeps it experimental and architecture-gated.

  ## Scope

  The fast path is only selected when:

  - `HIPFIRE_MMQ=1`
  - GPU arch is one of:
    - `gfx1100`
    - `gfx1101`
    - `gfx1102`
    - `gfx1103`
    - `gfx1150`
    - `gfx1151`

  Unsupported architectures compile stub entry points for the new kernel, but dispatch does not select the MMQ path there.

  This may be worth experimenting with on other AMD architectures later, but I did not want to silently change behavior outside the RDNA3/RDNA3.5 family in this PR.

  ## Why

  The Strix Halo prefill investigation showed that hipfire was much slower than llama.cpp on prompt processing. The main issue appears to be the HFQ4 prefill GEMM algorithm rather than KV cache mode alone.

  The productive experiment was moving the hot HFQ4 prefill projections toward a llama.cpp-style MMQ shape:

  - Q8_1 activation tiles
  - LDS staging
  - 128 batch columns x 128 output rows per workgroup
  - i8 WMMA accumulation
  - HFQ4 scale/zero correction during accumulation

  Other attempted paths were either neutral or regressions and are not included here.

  ## What Changed

  - Adds `kernels/src/gemm_hfq4g256_residual_mmq.hip`.
  - Adds Q8_1/MMQ activation quantization scratch for prefill.
  - Wires the MMQ path into HFQ4 residual, QKV, QKVZA, and gate/up prefill calls.
  - Keeps the path opt-in via `HIPFIRE_MMQ=1`.
  - Keeps Strix Halo on the existing dot2 HFQ4 prefill path unless MMQ is explicitly enabled.
  - Adds reusable prefill scratch behind:
    - `HIPFIRE_PREFILL_REUSE_PBS=1`
    - `HIPFIRE_PREFILL_MAX_BATCH=<N>`
  - Extends `test_inferenceQA` with stronger batched prefill vs sequential correctness checks across KV cache modes.
  - Adds experiment and validation notes under `docs/perf-checkpoints/`.

  ## Correctness Validation

  The new QA compares batched prefill against sequential token-by-token forward pass.

  It checks:

  - final prefill top token matches;
  - selected-logit drift stays within tolerance;
  - the next decode step after prefill also matches top token and selected-logit tolerance.

  That last check is intended to validate KV cache behavior: if batched prefill writes K/V slots incorrectly, the following decode step should diverge from the sequential reference.

  Validated on Strix Halo `gfx1151`, ROCm 7.2 toolbox, Qwen3.5 9B HFQ4/MQ4:

  ```text
  HIPFIRE_VERIFY_GRAPH=0 HIPFIRE_PREFILL_REUSE_PBS=1 HIPFIRE_PREFILL_MAX_BATCH=2048 \
  target/release/examples/test_inferenceQA \
    --model /home/kotdath/.hipfire/models/qwen3.5-9b.mq4 \
    --qa-case prefill_batch_matches_sequential

  Passed with HIPFIRE_MMQ unset.

  HIPFIRE_VERIFY_GRAPH=0 HIPFIRE_MMQ=1 HIPFIRE_PREFILL_REUSE_PBS=1 HIPFIRE_PREFILL_MAX_BATCH=2048 \
  target/release/examples/test_inferenceQA \
    --model /home/kotdath/.hipfire/models/qwen3.5-9b.mq4 \
    --qa-case prefill_batch_matches_sequential

  Passed with HIPFIRE_MMQ=1.

  Covered KV cache modes:

  - q8
  - asym4
  - asym3
  - asym2

  Covered prompt lengths inside QA:

  - n=2
  - n=7
  - n=17
  - n=33

  ## Performance

  Measured on Strix Halo gfx1151, ROCm 7.2 toolbox, qwen3.5-9b.mq4.

  Settings:

  - HIPFIRE_VERIFY_GRAPH=0
  - HIPFIRE_PREFILL_REUSE_PBS=1
  - HIPFIRE_PREFILL_MAX_BATCH=2048
  - --prefill-runs 3
  - values are median prefill tok/s
  - baseline is the same branch with HIPFIRE_MMQ unset

  | KV mode | pp | MMQ off | MMQ on | Speedup |
  |---|---:|---:|---:|---:|
  | q8 | 256 | 363.1 | 1127.6 | 3.11x |
  | q8 | 512 | 352.0 | 1179.8 | 3.35x |
  | q8 | 1024 | 328.9 | 1222.7 | 3.72x |
  | q8 | 2048 | 318.2 | 1168.5 | 3.67x |
  | asym4 | 256 | 368.6 | 1108.8 | 3.01x |
  | asym4 | 512 | 360.7 | 1173.3 | 3.25x |
  | asym4 | 1024 | 333.9 | 1223.0 | 3.66x |
  | asym4 | 2048 | 312.3 | 1151.7 | 3.69x |
  | asym3 | 256 | 361.4 | 1124.5 | 3.11x |
  | asym3 | 512 | 359.8 | 1187.3 | 3.30x |
  | asym3 | 1024 | 329.9 | 1259.1 | 3.82x |
  | asym3 | 2048 | 314.1 | 1216.5 | 3.87x |
  | asym2 | 256 | 374.0 | 1116.2 | 2.98x |
  | asym2 | 512 | 356.6 | 1173.2 | 3.29x |
  | asym2 | 1024 | 340.1 | 1208.5 | 3.55x |
  | asym2 | 2048 | 311.4 | 1142.9 | 3.67x |

  The gain is consistent across q8/asym KV cache modes. At longer prefill sizes (pp=1024..2048), this moves prefill from roughly 311-340 tok/s to 1143-1259 tok/s.

  A later single pp2048 smoke after the final kernel-source stub change produced:

  prefill_tok_s=1184.5

  ## Additional Checks

  Build:

  cargo build --release --features deltanet \
    --example test_hfq4g256QA \
    --example test_inferenceQA \
    --example bench_qwen35_mq4 \
    -p engine

  HFQ4/MMQ channel QA:

  HFQ4G256 QA PASS: gpu_cpu_err=0.000366 quant_ref_err=0.000000 mmq_err=0.040617

  Binary md5:

  7505b5121044d70847b826f3d084898d  target/release/examples/bench_qwen35_mq4

  Kernel checks:

  ./scripts/write-kernel-hashes.sh

  Completed successfully.

  Isolated compile check for the new kernel passed for:

  - gfx1010
  - gfx1030
  - gfx1100
  - gfx1200
  - gfx1201

  The full kernel matrix was also invoked:

  ./scripts/compile-kernels.sh gfx1010 gfx1030 gfx1100 gfx1200 gfx1201

  It still fails on pre-existing unrelated kernels in current upstream:

  998/1055 compiled, 57 failed

  The new MMQ source no longer blocks non-RDNA3 targets; unsupported targets compile stub entry points and dispatch does not select the MMQ path there.

  ## Caveats

  - This was validated on one Strix Halo / gfx1151 system.
  - The algorithm is experimental and not enabled by default.
  - I am not yet fully confident that this is the final/canonical MMQ implementation.
  - Correctness is validated behaviorally through batched-vs-sequential logits and post-prefill decode parity, not by byte-comparing raw KV buffers.
  - Other AMD architectures may benefit from a similar approach, but this PR intentionally avoids changing their runtime behavior.
  - cargo clippy was not available in my environment.
  - cargo fmt --all --check reports broad pre-existing formatting diffs across the repository, so this PR avoids reformatting unrelated files.

### DIFF
--- a/crates/engine/examples/bench_qwen35_mq4.rs
+++ b/crates/engine/examples/bench_qwen35_mq4.rs
@@ -4,7 +4,7 @@
 //! via an explicit warmup phase, and reports per-token latency stats plus
 //! an effective memory bandwidth estimate (weights_bytes × gen_tok/s).
 //!
-//! Usage: bench_qwen35_mq4 <model.hfq> [--prefill <N>] [--gen <N>] [--warmup <N>]
+//! Usage: bench_qwen35_mq4 <model.hfq> [--prefill <N>] [--prefill-runs <N>] [--gen <N>] [--warmup <N>]
 
 #[cfg(not(feature = "deltanet"))]
 fn main() { eprintln!("build with --features deltanet"); }
@@ -19,19 +19,21 @@ fn main() {
 
     let args: Vec<String> = std::env::args().collect();
     if args.len() < 2 {
-        eprintln!("Usage: bench_qwen35_mq4 <model.hfq> [--prefill N] [--gen N] [--warmup N]");
+        eprintln!("Usage: bench_qwen35_mq4 <model.hfq> [--prefill N] [--prefill-runs N] [--gen N] [--warmup N]");
         std::process::exit(1);
     }
     let model_path = &args[1];
 
     // Defaults: 32-token prefill, 5-token warmup, 100-token bench.
     let mut prefill_len: usize = 32;
+    let mut prefill_runs: usize = 1;
     let mut gen_len: usize = 100;
     let mut warmup_len: usize = 5;
     let mut i = 2;
     while i < args.len() {
         match args[i].as_str() {
             "--prefill" => { prefill_len = args[i + 1].parse().unwrap(); i += 2; }
+            "--prefill-runs" => { prefill_runs = args[i + 1].parse::<usize>().unwrap().max(1); i += 2; }
             "--gen"     => { gen_len     = args[i + 1].parse().unwrap(); i += 2; }
             "--warmup"  => { warmup_len  = args[i + 1].parse().unwrap(); i += 2; }
             other => { eprintln!("unknown arg: {other}"); std::process::exit(1); }
@@ -40,7 +42,7 @@ fn main() {
 
     eprintln!("=== bench_qwen35_mq4 ===");
     eprintln!("Model: {model_path}");
-    eprintln!("Phases: prefill={prefill_len} warmup={warmup_len} gen={gen_len}");
+    eprintln!("Phases: prefill={prefill_len} prefill_runs={prefill_runs} warmup={warmup_len} gen={gen_len}");
 
     let hfq = HfqFile::open(Path::new(model_path)).expect("open model");
     let config = qwen35::config_from_hfq(&hfq).expect("read config");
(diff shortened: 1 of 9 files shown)


Kaden-Schutt (approved):
LGTM — merging as opt-in. The gating discipline is exactly right for an experimental algorithm with stated correctness uncertainty.

## Validation on gfx1100

I don't have Strix Halo locally, but I did run the basic shape on a 7900 XTX to make sure the gating is tight and the path itself is correct:

**1. Default path unchanged (HIPFIRE_MMQ unset)** — \`./scripts/speed-gate.sh --fast\`:

| Metric | Baseline | This PR (default) | Δ |
|---|---:|---:|---:|
| 4B MQ4 pp32 prefill | 1014.0 | 1123.3 | +10.8% (jitter) |
| 4B MQ4 decode | 178.0 | 177.7 | -0.2% |

PASS. The env-var + arch-gate at every call site (\`std::env::var("HIPFIRE_MMQ")... && has_mmq_i8_wmma(&self.arch)\`) means zero observable change when the flag is off. No precompile-spec entry for the new MMQ kernel either, so it's not even JIT-compiled at startup unless used. Cost discipline is clean.

**2. MMQ-on path runs cleanly on gfx1100** (4B, \`HIPFIRE_MMQ=1\` + \`--prefill-runs 3\`):

| Workload | Default | MMQ | Δ |
|---|---:|---:|---:|
| 4B pp32 prefill | ~1100 | 20 | **-98%** (Q8_1 quantize overhead dominates at tiny batch) |
| 4B pp256 prefill | 3603 | **4316** | **+19.8%** |
| 4B decode | 178 | 169 | -5% (within jitter) |

The pp32 floor is exactly what I'd expect — the per-batch \`quantize_q8_1_mmq_ds4\` kernel doesn't amortize at small batch. At pp256 the activation-quantization cost folds into the i8 WMMA win and the path is +19.8%. Coherence-gate-style smoke ("What is the capital of France?") returns "The capital of France is Paris." cleanly with HIPFIRE_MMQ=1.

So the algorithm produces correct output AND wins on the workloads it was designed for (prompt processing on long prompts), even on RDNA3 — the Strix Halo gap you targeted is just where the win is largest. Real lever, not a phantom.

## Structural review ✅

- \`has_mmq_i8_wmma\` predicate matches \`gfx1100|gfx1101|gfx1102|gfx1103|gfx1150|gfx1151\`. Conservative and explicit. Adding gfx1200/1201 (RDNA4) when the i8 WMMA semantics are validated there is a one-line change.
- Both \`HIPFIRE_MMQ=1\` (full coverage) and \`HIPFIRE_WO_MMQ=1\` (residual-only opt-in) are wired. The narrower flag is useful for bisecting which kernel actually delivers the win on a given workload.
- The bench harness gets a \`--prefill-runs N\` flag — useful for amortizing JIT compile cost into the timing window. Default \`1\` preserves existing behavior. Solid.

## Out-of-scope follow-ups (none blocking)

- **Promote out of opt-in once correctness is validated by the community.** A few real-world coherence runs (CASK / DFlash / long-context multi-turn) on the MMQ-on path is what's needed. Issue #60 (prefill scaling vs llama.cpp) is the right umbrella for that work — when this PR + community validation closes that gap, we can flip the default for matching arches in a follow-up.
- **Stub kernels for unsupported arches**: you mentioned compiling stubs on non-RDNA3/3.5. I didn't see explicit stub-kernel registrations in the diff but the dispatch gate \`has_mmq_i8_wmma\` makes the stubs unreachable, so it's net-net safe. A doc comment in \`kernels.rs\` would be nice.
- **Add a one-line warn on engine startup when HIPFIRE_MMQ=1** so users know they're on the experimental path. Mirrors the HIPFIRE_GRAPH=1 warn that exists in cli/index.ts (line 2663+). Happy to take a separate one-line PR for that.
- **HFQ6/MQ6 MMQ port** is the obvious extension once HFQ4 is solid. Q8_1 activation quantization is format-agnostic; the i8 WMMA reduction is the same.

## Issue #60 status

This PR is the strongest concrete experimental result on #60 to date. Will cross-reference there. Won't close #60 — the "match llama.cpp at pp≥512 on 9B+" target is a wider investigation than just one kernel path.

Thanks for the cautious framing, the perf-checkpoint docs (\`docs/perf-checkpoints/2026-04-27-mmq-prefill-experiment-report.md\` — that's a great artifact, will mirror the format in our methodology folder), and the multi-flag testing surface (\`HIPFIRE_WO_MMQ\` for narrower bisection). This is contribution craft.

Kaden-Schutt:
Merged manually as commit 0c7d5f8 — your branch had a clean trivial conflict with PR #71 (just-merged gfx12 residual WMMA dispatch) that I resolved by ordering the new \`HIPFIRE_MMQ=1\`-gated dispatch ABOVE the gfx12 WMMA branch (so the env opt-in trumps automatic arch dispatch on gfx1100; gfx12 still gets WMMA when MMQ is off).

GitHub's merge UI couldn't auto-resolve, but the rebase was a one-liner — the \`kernels.rs\` const list and the \`gemm_hfq4g256_residual\` dispatch arm were both \"both-branches-add-a-line\" trivial overlaps. Authorship preserved (\`Authored-by: KotDath\`).

Full review on this PR thread above. Net result:
- Default path unchanged (gate at master verified clean: 4B pp32 +10.8% jitter, decode -0.2%).
- HIPFIRE_MMQ=1 path runs cleanly on gfx1100 + produces coherent output.
- HIPFIRE_MMQ=1 wins +19.8% on 4B pp256, matching the amortization-once-batch-is-large story you described.
- Cross-referenced from #60 (prefill scaling vs llama.cpp).

Marking closed since GitHub didn't auto-close it on the manual merge. Thanks for the careful framing and the perf-checkpoint docs.

PR #75, opened 2026-04-27
fix: List mq6 models downloaded

Nereuxofficial:
Currently listing doesn't show the mq6 models downloaded. Here is the fix for it. 

There should also be a nicer way to do this(e.g. List all files, emitting errors when models are unrecognized and also listing those in progress).

### DIFF
--- a/cli/index.ts
+++ b/cli/index.ts
@@ -1846,7 +1846,7 @@ function listLocal() {
     let entries: string[];
     try { entries = readdirSync(dir); } catch { continue; }
     for (const f of entries) {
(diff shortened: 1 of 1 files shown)


Kaden-Schutt (approved):
LGTM, merging.

Confirmed scope: the other ext-filter sites in cli/index.ts (1785, 1791-1795) already include \`.mq6\`. Only \`listLocal()\` at line 1849 was missing it — \`hipfire list\` was silently dropping any \`.mq6\` files in the model dir.

Re your follow-up suggestion ("there should also be a nicer way to do this") — agreed. Tracking as a follow-up: a single \`isModelFile()\` helper used at every discovery site, with a \"unrecognized\" warn-once line for files in the model dir that don't match. Happy to take a separate PR for that whenever you want, no rush.

Thanks!

PR #62, opened 2026-04-27
dispatch(gfx12): wire WMMA kernels (issue #57)

RobinVanCauter:
Closes #57 (partial — see _Open follow-ups_).

Wires the six `gemm_*_wmma_gfx12` kernels merged in #56 into `dispatch.rs` ahead of the dot2 fallback for gfx12 (RDNA4). Validated on R9700 (gfx1201).

## What changed

`crates/rdna-compute/src/dispatch.rs`:
- New `has_wmma_f16_gfx12(arch)` predicate (matches `gfx12*`).
- Six call sites get the gfx12 probe inserted **above** the existing `has_wmma_f16` (gfx11-only) probe, so gfx11 behavior is unchanged:
  - `gemm_qkv_hfq4g256`, `gemm_qkv_hfq6g256`
  - `gemm_qkvza_hfq4g256`, `gemm_qkvza_hfq6g256`
  - `gemm_gate_up_hfq4g256`, `gemm_gate_up_hfq6g256`
- Doc comment on `has_wmma_f16` trimmed: the "gfx12 falls back to dot2" wording is no longer true; replaced with a cross-reference to the new predicate.

`tests/speed-baselines/gfx1201.txt`:
- New file. Captured against this branch's wired-in WMMA path so future RDNA4 contributors have a floor.

## R9700 perf delta vs dot2 fallback

Best-of-2/3, `KV=asym3`, `HIPFIRE_GRAPH=1` (4B/9B), no graph (0.8B). Same `bench_qwen35_mq4` matrix used by `speed-gate.sh`.

| Model | Workload | dot2 (current master) | WMMA (this PR) | Δ |
|---|---|---:|---:|---:|
| **0.8B** | pp32 prefill   | 2989 | 3146 | **+5.3%** |
|       | pp128 prefill | 3823 | 5494 | **+43.7%** |
|       | decode        |  283 |  288 | flat |
| **4B**  | pp32 prefill  |  723 | 1084 | **+50.0%** |
|       | pp128 prefill | 1038 | 1438 | **+38.5%** |
|       | decode        |  137 |  137 | flat |
| **9B**  | pp32 prefill  |  578 |  664 | **+14.9%** |
|       | pp128 prefill |  656 |  732 | **+11.6%** |
|       | decode        |   97 |   99 | flat |

Decode is GEMV (not GEMM), so it never goes through these kernels — flat as expected.

## Validation

- **Channel-tests** (the six `test_wmma_*_gfx12` examples from #56): 36/36 shapes pass.
[...]

PR #91, opened 2026-04-29
perf(gfx12-wmma): K4 K-tile unroll on the WMMA GEMM family

### LINKED ISSUE #65: gfx12 WMMA: tune 9B prefill (multi-row, K-tile, s_prefetch, launch_bounds)
**Status**: follow-up to PR #62 (gfx12 WMMA kernels wired into dispatch). Hardware required: **R9700 / 9070 XT** (gfx1200 / gfx1201).

## Context

PR #62 wired the six `gemm_*_wmma_gfx12` kernels into `dispatch.rs`. R9700 perf vs dot2 fallback is strong on small models (0.8B pp128 +43.7%, 4B pp32 +50%) but **9B prefill is +11.6%–14.9%, below the 1.3× threshold** that the original issue #57 flagged as a tuning signal. Decode is unaffected (GEMV path).

Likely cause for the 9B specifically: BW-bound or occupancy-bound at that shape, with kernel parameters lifted from gfx11 (RDNA3) that don't account for RDNA4's larger register file or different L1/L2 cache hierarchy.

## Levers (each a discrete experiment)

These were enumerated by @RobinVanCauter in PR #62 as out-of-scope-for-wiring. Each is independent — a contributor can pick one without blocking the others:

1. **Multi-row WMMA** (highest expected impact). Share x-register state across R rows per warp instead of one row. Cuts global loads of `x` by R×. Ref: similar move on gfx1100 lifted decode 1.5× in commit `4105035`.
2. **K-tile depth retune**. Current K2 unroll is the gfx11 default; RDNA4 cache may prefer K4 or K8. Sweep `K_TILE` constant in each `gemm_*_wmma.gfx12.hip`.
3. **`__launch_bounds__` retune**. `(32, 2)` lifted from gfx11; RDNA4's larger register file may want different `(threads_per_block, blocks_per_cu)` for higher occupancy.
4. **Port `s_prefetch_data`** from `gemv_hfq4g256.gfx1201.hip`. Same issuing pattern likely applies to the GEMM kernels.

## Validation per attempt

- All six `test_wmma_*_gfx12` channel-tests must remain 36/36.
- `coherence-gate.sh` must pass on 9B.
- `speed-gate.sh --update-baselines` to capture the new floor in `tests/speed-baselines/gfx1201.txt`, then `--fast` to confirm it passes against itself.
- Cross-arch: gfx1100 `speed-gate.sh --fast` to confirm gfx11 path unchanged (the dispatch predicate `has_wmma_f16_gfx12` won't match gfx11, but stale-binary traps exist — see `.skills/hipfire-kernel-tuning/playbook.md` step 6).

## Skill

Read `.skills/hipfire-kernel-tuning/` first — it codifies the methodology, lever-to-lever ordering, gate protocol, and 5 case studies of past wins/regressions. Section `levers.md` has detailed write-ups of each lever above with commit references.

## Bench commands (R9700)

```bash
# Channel-tests (correctness floor)
for k in gemm_qkv_hfq4g256_wmma_gfx12 gemm_qkv_hfq6g256_wmma_gfx12 \
         gemm_qkvza_hfq4g256_wmma_gfx12 gemm_qkvza_hfq6g256_wmma_gfx12 \
         gemm_gate_up_hfq4g256_wmma_gfx12 gemm_gate_up_hfq6g256_wmma_gfx12; do
  cargo run --release --example test_wmma_$k --features deltanet
done

# Coherence (no looping/garbage)
./scripts/coherence-gate.sh

# Speed (regression floor)
./scripts/speed-gate.sh --fast              # quick pp32+decode on 4B
./scripts/speed-gate.sh                     # full pp32+pp128 on 0.8B/4B/9B/27B + DFlash
```

## Why this matters

This is the kind of work that compounds across the RDNA4 family — gfx1200 (Radeon AI Pro R9700) and gfx1201 (RX 9070 / 9070 XT) share the same WMMA path. Any tuning here lifts both chips. The fast paths from this PR will eventually be the 7900-XTX-equivalent maturity for RDNA4.

RobinVanCauter:
Closes #65.

Issue #65 enumerated four candidate levers for tuning the gfx12 WMMA GEMM family further past the floor PR #62 left it at: multi-row WMMA, K-tile depth retune, `__launch_bounds__` retune, and porting `s_prefetch_data` from the gemv kernel. This PR researches all four and lands the K-tile depth retune, which is the only one that wins on this shape — and it wins big. The other three are documented as negative results so the next contributor doesn't re-walk the same cul-de-sacs.

## Speed (R9700 in isolation, cold-process median across 5 fresh runs)

| model      | metric      | K2 (master) | K4       | Δ      |
|------------|-------------|-------------|----------|--------|
| qwen3.5-9b | pp32 prefill |  1178 tok/s |  1611    | +37 %  |
| qwen3.5-9b | pp128 prefill|  1199 tok/s |  1916    | **+60 %** |
| qwen3.5-9b | decode (pp32)|  101.8 tok/s|  101.3   | flat   |
| qwen3.6-27b| pp32 prefill |   311 tok/s |   524    | +68 %  |
| qwen3.6-27b| pp128 prefill|   326 tok/s |   565    | **+73 %** |
| qwen3.6-27b| decode (pp32)|   35.8 tok/s|   35.9   | flat   |

Spread on every measurement is 0.2 – 0.9 % (5 fresh processes per point via `scripts/bench-cold.sh`, DPM-stable). Decode is unchanged because the lever only touches the WMMA prefill GEMM path — decode runs through GEMV.

Speed-gate `--update-baselines` captured; the new floor is roughly double the prior `gfx1201.txt` (0.8B pp128 +48 %, 4B pp128 +75 %, 9B pp128 +84 %, 27B pp128 first capture).

## What changed

K2 → K4 in the inner K-tile loop of all seven gfx12 WMMA GEMM kernels:

```
kernels/src/gemm_gate_up_hfq4g256_wmma.gfx12.hip
kernels/src/gemm_gate_up_hfq6g256_wmma.gfx12.hip
kernels/src/gemm_qkv_hfq4g256_wmma.gfx12.hip
kernels/src/gemm_qkv_hfq6g256_wmma.gfx12.hip
kernels/src/gemm_qkvza_hfq4g256_wmma.gfx12.hip
kernels/src/gemm_qkvza_hfq6g256_wmma.gfx12.hip
kernels/src/gemm_hfq4g256_residual_wmma.gfx12.hip
```

Each kernel's body changes from "8 iterations × 2 K-tiles" to "4 iterations × 4 K-tiles, all loads up-front". The lift comes from latency hiding: at K2 the kernel issues 2 WMMAs back-to-back on the same `acc` (2-deep dependency chain) before loading the next pair of weights/X tiles. At K4 the chain is 4 deep *and* the 4 loads issue at the top of the body, so the load-issue pipeline overlaps with WMMA execution across the entire group iteration. RDNA4's WMMA engine + L2 hierarchy at this kernel's working-set size is set up well for this; gfx11's K2 was the right gfx11 trade-off, gfx12's bigger register file made K4 newly available.

## Levers tried (the issue's four + variations)

| Lever | Result | Notes |
|---|---|---|
| **2 — K4 K-tile unroll** | **+50 % per call** on the hottest kernel | the win — landed |
| 3 — `__launch_bounds__` retune | null / mild regression | tested (32,8) and (32,16); kernel is not occupancy-limited at this shape |
| 1 — multi-row WMMA M=2 | -70 to -134 % per call | reverts. Same case-study §3 fingerprint as the historical k2x32 attempt — doubled accumulator + 4× dequant live ranges spill the register budget; both K=2 and K=1 variants regress, so it's not just spill but also throughput cost from doubled WMMA per block + halved grid |
| 4 — `s_prefetch_data` port | -3 to -7 % | tested 1- and 2-group lookahead. GEMM working set is small enough that the HW prefetcher already covers it; the explicit prefetch pays an extra instruction per iteration with no return. Pattern that wins on `gemv_hfq4g256.gfx1201.hip` (per-row stride too wide for HW prefetcher) doesn't generalize to this kernel's shape |

Per-kernel µs/call on R9700 9B pp=128 with K4 (vs K2 master):

```
gemm_gate_up_hfq4g256_wmma_gfx12   1842 → 902 µs/call   (-51 %, 32 calls/layer)
gemm_hfq4g256_residual_wmma_gfx12   326 → 246 µs/call   (-25 %, 64 calls/layer)
gemm_qkvza_hfq4g256_wmma_gfx12      558 → 444 µs/call   (-20 %, 24 calls/layer)
gemm_qkv_hfq4g256_wmma_gfx12        461 → 357 µs/call   (-23 %,  8 calls/layer)
```

## Validation

| Gate | Result |
|---|---|
| Channel-tests | **43/43 pass** (6 hfq4 fused-projections + 6 hfq6 fused-projections + 7 residual cases) |
| Coherence-gate | 4/4 outputs fluent, on-topic, no loops |
| Speed-gate vs prior `gfx1201.txt` floor | every metric FAST, +24 % to +85 % |
| Speed-gate `--fast` vs new floor | within ±0.7 % (self-consistent) |
| Compile-check 7 changed kernels for gfx1201 | 7/7 ✓ |
| Cross-arch impact | none — kernels are family-tagged `.gfx12.hip`, dispatch tree falls through to existing `_wmma.hip` (gfx11) / dot2 / scalar fallbacks for non-gfx12 archs |

## Bench instrumentation (separate commit)

The first commit lands two general-purpose pieces of methodology that were necessary to land this kind of measurement honestly:

1. **`bench_qwen35_mq4`: DPM warmup before prefill.** The pre-existing `HIPFIRE_DPM_WARMUP_SECS` hook fires after warmup tokens but before gen — useless for prefill, where the first ~50 ms otherwise measures clock-ramp not steady-state. On RDNA4 the gap between idle DPM and peak DPM is ~30 % of clock; without the warmup, kernel-perf measurements drift by that amount across runs.

2. **`scripts/bench-cold.sh`.** N fresh processes per (model, pp) point with `--prefill-runs 3` inside each (so reported tok/s is a warm-process measurement) plus a min/max/median/spread% summary. Drives the within-session noise band from 8 % to under 1 %, which is what made the small-lever attempts (s_prefetch, launch_bounds) distinguishable from genuine wins.

These are not specific to issue #65 — any future kernel-tuning work should pick them up. Kept in their own commit so the kernel diff stays focused.

## Reproducing

```
# Channel tests
for k in qkv qkvza gate_up; do
  cargo run --release --features deltanet --example test_wmma_${k}_gfx12
  cargo run --release --features deltanet --example test_wmma_${k}_hfq6_gfx12
done
cargo run --release --features deltanet --example test_wmma_residual_gfx12

# Coherence
./scripts/coherence-gate.sh

# Speed-gate against new floor
HIPFIRE_MODELS_DIR=$HOME/.hipfire/models ./scripts/speed-gate.sh

# Cold-process distribution (the way the +60 % / +73 % numbers above were measured)
./scripts/bench-cold.sh ~/.hipfire/models/qwen3.5-9b.mq4 \
    --pp 32,128 --runs 5 --label K4-9b
./scripts/bench-cold.sh ~/.hipfire/models/qwen3.6-27b.mq4 \
    --pp 32,128 --runs 5 --label K4-27b
```

### DIFF
--- a/crates/engine/examples/bench_qwen35_mq4.rs
+++ b/crates/engine/examples/bench_qwen35_mq4.rs
@@ -87,6 +87,19 @@ fn main() {
     // benchmark independent of tokenizer / chat template behaviour.
     let prompt_tokens: Vec<u32> = (0..prefill_len as u32).collect();
 
+    // DPM warmup BEFORE prefill (issue #65). The default `HIPFIRE_DPM_WARMUP_SECS`
+    // hook fires AFTER the warmup tokens, before the timed gen — useless for
+    // prefill measurement when the GPU is in DPM step 0/1 from idle. This
+    // mirrors that hook for the prefill phase: stabilizes clocks before the
+    // timed forward_prefill_batch.
+    if let Ok(secs_str) = std::env::var("HIPFIRE_DPM_WARMUP_SECS") {
+        let secs: f32 = secs_str.parse().unwrap_or(0.0);
+        if secs > 0.0 {
+            eprintln!("\n=== DPM warmup ({secs:.1}s, pre-prefill) ===");
+            gpu.dpm_warmup(secs).expect("dpm warmup");
+        }
+    }
+
     // === PREFILL ===
     // Route through forward_prefill_batch so the bench measures the production
     // prefill path (daemon + greedy_dump both go through it). Inside, this
--- a/kernels/src/gemm_gate_up_hfq4g256_wmma.gfx12.hip
+++ b/kernels/src/gemm_gate_up_hfq4g256_wmma.gfx12.hip
@@ -4,11 +4,17 @@
 // WMMA-accelerated batched 2-way fused HFQ4-G256 GEMM for the Qwen3.5
 // FFN preamble (gate + up projections).
 //
-// **gfx12 (RDNA4) variant** — sister of gemm_gate_up_hfq4g256_wmma.hip
-// (gfx11 / RDNA3). Compile with `hipcc --offload-arch=gfx1200` (or 1201).
+// **gfx12 (RDNA4) variant — K4 unroll**, issue #65 lever 2. The original
+// PR #62 ship used K2 (lifted from gfx11). RDNA4 has higher memory
+// latency / less aggressive scheduler than RDNA3 at this kernel shape;
+// deeper unroll gives the compiler a longer pipeline to hide loads
+// behind WMMA execution. Measured 2.0× on R9700 9B pp=128 gate_up
+// (1842 → 922 µs/call, 39 → 80 GiB/s effective).
 //
-// Differences from the gfx11 kernel — same set as the QKV scaffold
-// (gemm_qkv_hfq4g256_wmma.gfx12.hip), validated on R9700 / gfx1201:
+// Same lane decomposition as the K2 baseline — the only structural
+// change is processing 4 K-tiles per loop body instead of 2.
+//
+// gfx12 vs gfx11 differences (carried over from the K2 baseline):
 //   1. WMMA builtin: __builtin_amdgcn_wmma_f32_16x16x16_f16_w32_gfx12
 //      (gfx11 uses _w32 without the _gfx12 suffix).
 //   2. A and B operands: half8_t (8 fp16 per lane) instead of half16_t.
@@ -17,7 +23,7 @@
 //        lane (tid >> 4) = 0  -> carries K = [0..7]  of the 16-K tile
 //        lane (tid >> 4) = 1  -> carries K = [8..15] of the 16-K tile
 //      Each lane loads HALF as much per WMMA tile as gfx11.
-//   4. C-output mapping (VALIDATED 2026-04-27 in the QKV scaffold's
+//   4. C-output mapping (validated 2026-04-27 in the QKV scaffold's
 //      channel-test on R9700):
 //        gfx11: acc[j] = C[2*j + (tid>>4)][tid & 15]    (rows interleaved)
 //        gfx12: acc[j] = C[8*(tid>>4) + j][tid & 15]    (rows contiguous —
@@ -29,6 +35,10 @@
 // Each 16-row tile may straddle the gate/up boundary. Per-thread output
 // routing resolves which Y pointer (Y_gate or Y_up) each row writes to.
 // Overwrite semantics (y = acc, not +=).
+//
+// K constraint: K must be a multiple of 256 (the HFQ-G256 group size).
+// The K4 unroll lives entirely inside a single group's 16 K-tiles —
+// it doesn't impose any additional K constraint.
 
 typedef _Float16 __attribute__((ext_vector_type(8))) half8_t;
 typedef float    __attribute__((ext_vector_type(8))) float8_t;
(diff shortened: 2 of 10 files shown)


Kaden-Schutt:
@RobinVanCauter — this is excellent. The negative-results table for the other three #65 levers is exactly the kind of write-up we want — saves the next contributor from re-walking those branches. The +60–73% on pp128 is the right shape (prefill is GEMM-bound, decode goes through GEMV and shouldn't move) and the per-kernel µs breakdown locates the win cleanly.

Pre-merge ask, just one thing: I don't have R9700 / gfx1201 hardware here, so I'm trusting the channel-tests + coherence-gate numbers in the PR. Could you confirm
1. The cold-process methodology you used — was it `scripts/probe_commits.sh HEAD~1 HEAD` (or equivalent) so the numbers are not within-session A/B? Your "5 fresh runs, 0.2–0.9% spread" suggests yes, but want it explicit.
2. Speed-gate `--update-baselines` was captured against the new floor — does the PR include the updated `tests/speed-baselines/gfx1201.txt`? If so, that's the right gate-of-record.

Both of these are formality given the rationale is sound. Will queue for Codex stop-time review pass and merge if clean. Thanks for the rigor.

RobinVanCauter:
Thanks!

Ran the channel-tests + coherence tests multiple times to make sure.

**1. Cold-process methodology.** 

Yes — fresh OS process per datapoint, mimicking `probe_commits.sh master HEAD`. Concretely: `bench-cold.sh` against a checkout of `master` for the K2 baseline, then against the PR tip for K4. Same two-commit shape probe_commits enforces, with a 5-run distribution at each end instead of a single sample.

`bench-cold.sh` (committed in `8ac8da7`, the methodology commit immediately preceding K4) was built specifically to drive the noise floor down. Within-session A/B on R9700 is ±8% wide; that band would have masked any real 2–3% lever effect as noise. Driving it to <1% per cell — via DPM-warmup-before-prefill, `--prefill-runs 3` for warm prefill, one untimed warm-up process per (model, pp) cell, and 5 timed fresh processes — was load-bearing for **the negative results**, not the K4 win:

- Lever 3 (`__launch_bounds__` retune) measured −2 to −4% — null vs mild regression is only distinguishable with sub-1% spread.
- Lever 4 (`s_prefetch_data`) measured −3 to −7% — same story.
- Lever 1 (multi-row WMMA M=2) was −70%, obvious without the apparatus.

K4 happened to be the giant (+37–73%) and would have been visible with a stopwatch. The discipline was for the search, not the headline result.

**2. Updated `tests/speed-baselines/gfx1201.txt`.** 

Yes — included in commit `71ab1ff`, captured via `./scripts/speed-gate.sh --update-baselines` against the K4 floor. Self-check vs the new floor (`speed-gate.sh --fast`) is within ±0.7%.

Kaden-Schutt (approved):
Closely re-read the diff:

**Kernel change shape (representative: `gemm_qkv_hfq4g256_wmma.gfx12.hip`):** mechanical K2 → K4 unroll. `for kt < 16; kt += 2` → `for kt < 16; kt += 4` with all 4 weight reads + 4 X reads hoisted up-front before the dequant+WMMA chain. `DQ8` macro lifted out of the loop body for symmetry across kernels. Same WMMA builtin (`__builtin_amdgcn_wmma_f32_16x16x16_f16_w32_gfx12`), same lane decomposition. The lift comes from the deeper load-issue pipeline overlap with the 4-deep WMMA chain — exactly the latency-hiding theory you laid out.

**Cross-arch impact verified:** `.gfx12.hip` kernels are dispatched only when `has_wmma_f16_gfx12()` matches; gfx1100 / gfx1030 / gfx906 fall through to existing `_wmma.hip` / dot2 / scalar paths. No risk to my gfx1100 baseline.

**Methodology commit:** `bench-cold.sh` + DPM-warmup-pre-prefill in `bench_qwen35_mq4` are the right shape — your point about how the sub-1% spread was load-bearing for the *negative* results is what made me confident in the result, not the lever. Good rigor; appreciate it living in its own commit so it's reusable.

**`tests/speed-baselines/gfx1201.txt` updated** per ask in commit `71ab1ff`. Self-check vs new floor within ±0.7%.

Negative-results table for the other three #65 levers is exactly the kind of write-up that pays forward. Closes #65 cleanly. Merging.

Issue #89, opened 2026-04-28
DFlash thinking-attractor bug on A3B drafts (qwen3.5/3.6 35b-a3b) — block loop in <think> at long budgets


Kaden-Schutt:
**Reporter:** @mikiadev in #79 — \"in 7-10 attempts the last part of thinking 'I'm going to answer the user <whatever>' got infinitely looped\"

**Reproduced:** 7900 XTX, qwen3.6:35b-a3b.mq4 + DFlash draft, prompt \"What is the answer to life, the universe, and everything?\", DFlash greedy (temp=0), max_tokens=600.

## Hit rate

2/5 runs hit a clear block-level attractor in the *last 1/3* of \`<think>\` (Tier-2 / Tier-3 in \`coherence-gate-dflash.sh\`). Matches @mikiadev's ~2/10 from real pi traffic.

Example signature from one stuck run:
\`\`\`
6× "context—which is why many people joke about how meaningless the answer actually was without"
\`\`\`
The phrase rotates as a closed loop; draft and target keep agreeing on it because it's "consistent with itself" — a textbook DFlash sidecar-drift attractor (cf. CASK m-fold + DFlash 2026-04-26 in \`feedback_dflash_coherence_gate_mandatory.md\`).

## Why A3B specifically

Per \`feedback_a3b_dflash_regression.md\` the global TriAttention sidecar for A3B-3.5 caps at R̄≈0.36 and A3B-3.6 at R̄≈0.39 — structural ceiling driven by MoE routing variance, NOT corpus-size (confirmed empirically on the optimized pipe in the 2026-04-28 MI300X session). Single-token attractors are easy for a high-R̄ sidecar to suppress; block-level (5+ token) attractors emerging in the second half of long greedy decodes are not. Thinking IS exactly long greedy decode on highly self-consistent token distributions, so this is the worst-case shape for an under-calibrated sidecar.

The loop is also invisible to \`coherence-gate-dflash.sh\`'s current Tier-1 (first 128) check — the attractor settles in tokens 200–500 of the thinking phase. Tier-2 / Tier-3 do flag it; we just don't currently run a thinking-shaped prompt through them.

## Affected paths

- qwen3.5:35b-a3b + DFlash + thinking enabled
- qwen3.6:35b-a3b + DFlash + thinking enabled
- A3B drafts at \`HIPFIRE_DFLASH_MODE=on\` (default for these models)

[...]

PR #93, opened 2026-04-29
feat(gcn5): add gfx906 bring-up path

myaple:
Map gfx906 through device detection, install scripts, profiler metadata, and kernel precompile setup. Route gfx906 through the wave64 GEMV path and skip WMMA/MQ8 blobs that Vega 20 cannot compile.

Also give the DFlash coherence gate the current 27B draft filename I could download and make its per-case timeout configurable for slower gfx906 runs, cause my MI50 is incredibly slow.

Validation:

- ./scripts/compile-kernels.sh gfx906 (196/196 compiled)

- bash ./scripts/test-kernelsQA.sh gfx906 (16 passed)

- bash ./scripts/test-kernels.sh gfx906 (16 passed)

- cargo run --release --features deltanet -p engine --example daemon -- --precompile

- cargo check -p rdna-compute

- ./scripts/coherence-gate.sh

- HIPFIRE_COHERENCE_TIMEOUT=600 ./scripts/coherence-gate-dflash.sh

I used the port arch skill with some extra prodding to do this work, validated it on my local 1x MI50 machine.

With the `--full` coherence check (not dflash version), the MoE did fall into a loop, and one of the smaller models didn't finish its thinking block, so I'm not positive where to go from here with that. I've attached the coherence check output for reference. I'd be very happy for some pointers, I'm quite new at LLM backends.

I've also attached the dflash coherence check, which I think went a bit wonky too, with some repetition and odd caps. 

[coherence-20260429-224602.md](https://github.com/user-attachments/files/27219021/coherence-20260429-224602.md)

[coherence-dflash-20260429-223341.md](https://github.com/user-attachments/files/27219022/coherence-dflash-20260429-223341.md)

### DIFF
--- a/crates/engine/examples/daemon.rs
+++ b/crates/engine/examples/daemon.rs
@@ -276,7 +276,7 @@ fn main() {
             // Arch is unknown until Gpu::init; use a broad mkdir for the common arches
             // we support so the probe picks one up. The real arch check after init
             // will log the active dir.
-            for arch in ["gfx1010", "gfx1013", "gfx1030", "gfx1031", "gfx1100", "gfx1101", "gfx1102", "gfx1151", "gfx1200", "gfx1201"] {
+            for arch in ["gfx906", "gfx1010", "gfx1013", "gfx1030", "gfx1031", "gfx1100", "gfx1101", "gfx1102", "gfx1151", "gfx1200", "gfx1201"] {
                 let _ = std::fs::create_dir_all(exe_dir.join("kernels").join("compiled").join(arch));
             }
         }
--- a/crates/rdna-compute/examples/gen_kernel_hashes.rs
+++ b/crates/rdna-compute/examples/gen_kernel_hashes.rs
@@ -59,7 +59,7 @@ fn main() {
     kernel_sources.sort_by(|a, b| a.0.cmp(&b.0));
     rdna2_variant_sources.sort_by(|a, b| a.0.cmp(&b.0));
 
-    let archs = ["gfx1010", "gfx1030", "gfx1100", "gfx1200", "gfx1201"];
+    let archs = ["gfx906", "gfx1010", "gfx1030", "gfx1100", "gfx1200", "gfx1201"];
 
     let mut written = 0;
     let mut skipped = 0;
--- a/crates/rdna-compute/src/dispatch.rs
+++ b/crates/rdna-compute/src/dispatch.rs
@@ -38,12 +38,13 @@ fn gemv_rows_default(arch: &str) -> u32 {
     match arch {
         "gfx1100" | "gfx1101" | "gfx1102" => 1,
         "gfx1030" | "gfx1031" => 1,
-        // CDNA1 (MI100, gfx908) + CDNA3 (MI300X): wave64 native.
+        // Vega 20 / GCN5 (gfx906), CDNA1 (MI100, gfx908), and CDNA3
+        // (MI300X): wave64 native.
         // `gemv_hfq4g256_wide` uses block=[64,1,1] = exactly one wave —
         // zero lane waste. The 32-thread multirow variants run on half a
         // wave, so the wide kernel is the natural fit. Return rows=1 to
         // trigger use_wide.
-        "gfx908" | "gfx940" | "gfx941" | "gfx942" => 1,
+        "gfx906" | "gfx908" | "gfx940" | "gfx941" | "gfx942" => 1,
         _ => 2,
     }
 }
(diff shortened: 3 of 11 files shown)


Kaden-Schutt:
@myaple — thanks for the bring-up; gfx906 / MI50 is exactly the kind of legacy-arch coverage we want and the routing-table additions look right at a glance (wave64 GEMV path, skip WMMA/MQ8, bump the dflash coherence-gate timeout for slow hardware). I don't have a Vega 20 here so this is going to merge on your local validation.

Two threads to untangle on the coherence behavior, since you asked for pointers:

**1. MoE loop on `--full` coherence:**

A3B (the MoE model) has documented attractor risk independent of arch — see #89. R̄ on the global TriAttention sidecar caps at ~0.36–0.39 vs ~0.55–0.65 on the dense models. If your `--full` run included a 35b-a3b model and looped, that's almost certainly THIS bug, not a gfx906 issue. The fix path is per-cluster / per-expert sidecars (active research), and the workaround today is `dflash_mode=off` on A3B targets — which we just defaulted to in `auto` mode (#37). 

Could you confirm which model looped? If it's `qwen3.5:35b-a3b` or `qwen3.6:35b-a3b`, that's #89 and not blocking your PR.

**2. Small model not finishing thinking block:**

If you saw a `<think>` open with no closing `</think>` and the response just terminated, that's typically one of:
- Sampling hit `<|im_end|>` mid-think (rare, can happen at temp=0 on certain prompt shapes — the small models are most sensitive to this on gfx906 since the effective fp32-shadow accumulation precision is different from RDNA3+)
- `max_tokens` (config or coherence-gate-set) cap landed inside thinking
- Something gfx906-specific in the chat template / EOT handling

Run `./scripts/coherence-gate.sh` with `HIPFIRE_LOG=info` redirected to a file and search for `finish_reason` and `tokens=N tok_s=` lines around the failed model — that'll tell you whether the daemon stopped on stop-token or token-cap. Easiest signal to interpret.

**Re the dflash coherence "wonky with repetition and odd caps":**

That's likely a separate symptom — could be the same A3B attractor if the tested DFlash target was an A3B model, or could be the published draft's draft-target alignment issue (drafts trained at one revision can drift from current target weights).

Will queue for Codex stop-time review and merge once the gfx906-vs-content-gating distinction is clear (i.e., we've confirmed your coherence concerns are pre-existing #89-class issues, not regressions caused by this PR's routing changes). Thanks for the careful validation matrix at the bottom of the PR — that's the right shape.

myaple:
That's fair. The model that looped is the A3B MoE, so that appears to be that bug.

For #2, the small model not finishing, this is what came out in the coherence log:

## qwen3.5-4b.mq4 — code

- wall: 5.5s  status: **OK**
- stats: `{"type":"done","id":"r1","tokens":180,"tok_s":42.3,"prefill_tokens":21,"prefill_ms":895.6,"prefill_tok_s":23.4,"decode_tok_s":53.6,"ttf
t_ms":895.6}`
- prompt: "Write a one-line Python function named square that returns n*n."

**Output:**

```
<think>
The user wants a one-line Python function named `square` that returns n*n. This is a simple function that takes one parameter (n) and returns n m
ultiplied by itself.

In Python, this would be:

\`\`\`python
def square(n): return n * n
\`\`\`

This is a one-liner function definition. Let me make sure this is correct - yes, in Python you can define functions on a single line using the `d
ef` keyword followed by the function name, parentheses for parameters, and then the return statement with an expression (not necessarily a `retur
n` keyword if it's just an expression).

Actually wait - in Python 3, you can use `return` keyword or just use the expression directly. Since we want to return n*n, we need to be careful
 about whether this is valid as a one-liner.

Let me think
```


I'm not seeing a stop token, but the reason is "done" and the token count is right at 180, so maybe that's the output limit for the tests, but that seems arbitrary. 

Thanks for your thoughts. I'm happy to be the gfx906 guinea pig for other merge requests or validation you need to do against the architecture, and once this goes in I'll try to see what perf uplifts I can do as later PRs. 

For the dflash issue, it was the 27b dense model:

## 27b-dflash-prose (dflash)

- wall: 223.7s  status: **OK**
- detector: `{"ok": true, "soft_warn": false, "total": 128, "unique": 53, "unique_ratio": 0.414, "max_freq": 0.078, "max_tok": 279, "max_count": 
10}`
- stats:
  ```
  emitted: 194 tokens in 196.21s  (0.99 tok/s)
  cycles: 53  committed: 246  accepted: 140  τ=2.642  mean_committed=4.642
  ```

**Output:**

```
 the sheer size of the empire. It was simply too large to govern effectively. The Roman legions were stretched thin, unable to defend the vast frontiers. The empire was also plagued by economic troubles. Inflation was rampant, and the tax burden on the common people was crushing. The Roman legions, once the pride of Rome, were increasingly composed of mercenaries who had little loyalty to the state. The Roman legions, once the pride of Rome, were increasingly composed of mercenaries who had little loyalty to the state. The Roman legions, once the pride of Rome, were increasingly composed of mercenaries who had stretched thin, unable to defend the vast frontiers. The Roman legions, once the pride of Rome, were increasingly composed of mercenaries who had little loyalty to the state. The Roman legions, once the pride of Rome, were increasingly composed of mercenaries who had little loyalty to the state. The Roman legions, once the pride of Rome, were
```

This only seems to happen in the dflash coherence, as the 27b in the dflash=off case seems to produce good language. It also gets  cut off, but that could be the same token output limit, not sure.

Kaden-Schutt (approved):
Closely re-read the diff:

**Routing (`crates/redline/src/device.rs`, `crates/rdna-compute/src/dispatch.rs`):** family 141 (AMDGPU_FAMILY_AI) split by `chip_external_rev` ∈ {0x3c..0x3f} → gfx906 is the right Vega 20 fingerprint (Radeon VII / MI50 class). The `has_wave64_native` extension and `gemv_rows_default` arch-list addition are mechanical and ISA-correct: the `*_wave64.hip` kernels use FMA + `shfl_down` only — no MFMA — so they're legal on gfx906.

**Compile gating (`scripts/compile-kernels.sh:53-65`):** skipping `*_wmma*` and `gemv_mq8g256` for gfx906 is correct — Vega 20 has no WMMA (RDNA3+) and no dot8 (gfx908+). `196/196` compile success in your validation matches.

**Detection / install (`scripts/_detect-gpu.sh`, `scripts/install.sh`, `scripts/speed-gate.sh`):** `gfx_target_version 90006 → gfx906` + `HSA_OVERRIDE_GFX_VERSION 9.0.6` + rocminfo fallback. Mechanical extensions of the existing pattern.

**Coherence-gate-dflash CASE_TIMEOUT env knob:** sane — slow hardware needs longer per-case windows. Default 240s preserved.

**On the loops you flagged:**
- A3B looping in `--full` is #89 (R̄ ceiling on the global TriAttention sidecar) — confirmed not a gfx906 regression.
- Small-model `<think>` ending early at exactly 180 tokens looks like the per-case max_tokens cap landed inside the thinking block. Coherence-gate uses a fixed cap that's tuned for chat answers; long thinkers can run out. Not a regression.
- 27B-dflash repetition on `prose` — that's the same ['Roman legions, once the pride'] block-attractor signature as our other DFlash quality concerns; not a Vega-20 issue.

Owner ack on cross-arch impact: gfx1100 dispatch tree unchanged after this lands. Thanks for the careful validation matrix at the bottom of the PR — that's the right shape for arch bring-ups.

Merging.

PR #103, opened 2026-04-30
fix(cli): resolve raw filenames to registry tags for per-model config

fivetide:
When `hipfire run qwen3.6-35b-a3b.mq4` is invoked with a raw filename instead of a registry tag (`qwen3.6:35b-a3b`), `resolveModelTag` failed to match it — the registry keys use colons while filenames use dashes plus an extension. This caused per-model config (max_think_tokens, dflash_mode, cask_sidecar) to silently fall through to global defaults.

Consequence: Qwen 3.6 A3B ran with max_think_tokens=0 (unlimited) instead of the configured 8192, entered a <think> loop consuming all 32768 max_tokens, and the CLI's think-stripping logic produced empty visible output.

Fix:
1. resolveModelTag: add reverse-resolve pass that matches registry entries by their .file field when tag/alias lookup fails.
2. run command: use resolveModelConfig(model) for CLI flag defaults instead of the global-only cfg, so per-model temperature, top_p, repeat_penalty, and max_tokens overrides apply consistently.

Verified: `hipfire run qwen3.6-35b-a3b.mq4 "how to center a div"` now resolves to tag qwen3.6:35b-a3b, applies max_think_tokens=8192, and produces visible content (1785 tok @ 76.7 tok/s).

Related: #89 (A3B thinking-loop attractor)

### DIFF
--- a/cli/index.ts
+++ b/cli/index.ts
@@ -515,6 +515,12 @@ function resolveModelTag(input: string): string {
   if (ALIASES[normalized]) return ALIASES[normalized];
   // Try adding "qwen3.5:" prefix
   if (REGISTRY[`qwen3.5:${normalized}`]) return `qwen3.5:${normalized}`;
+  // Reverse-resolve: if input looks like a filename (e.g. "qwen3.6-35b-a3b.mq4"),
+  // find the registry entry whose .file matches and return its tag. Without this,
+  // per-model config is silently ignored when the user passes a raw filename.
+  for (const [tag, entry] of Object.entries(REGISTRY)) {
+    if (entry.file === normalized || entry.file === input) return tag;
+  }
   return normalized;
 }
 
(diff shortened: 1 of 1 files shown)


Kaden-Schutt (approved):
Closely reviewed the diff:

- `resolveModelTag` reverse-lookup loop (lines 515-521) is correct — runs only on tag-resolution miss, linear scan is fine for the registry size.
- `run` command `runCfg = resolveModelConfig(model)` (lines 3597-3604) plumbs per-model overrides through CLI flag defaults; no double-call risk because the modelCfg at line 1048 is in a different block scope.

Fixes a real silent failure (Qwen 3.6 A3B max_think_tokens=8192 was being dropped on raw-filename invocation → empty visible output). Ships.

Thanks for the careful diff.

Issue #87, opened 2026-04-28
Track auto-MMQ regression on tool-call output (gfx1151) — reverted on master


Kaden-Schutt:
**Status:** Auto-MMQ at batch≥128 from #84 (cherry-pick `0d3621a`) reverted on master at `d1506d0` to unbreak qwen3.6:27b tool-call output on Strix Halo.

## What landed and what broke

`0d3621a` (Tue Apr 28) cherry-picked from PR #84 wired auto-enable of the MMQ (i8 WMMA + Q8_1) prefill path on gfx1150/gfx1151 when `batch_size ≥ 128`. Bench showed +63% pp128 / +96% pp256 on 9B Strix Halo. The fp16 WMMA path stayed the default below 128 (where Q8_1 quantization overhead dominates), and `HIPFIRE_MMQ=0` was added as an explicit kill-switch.

@nwoolmer reported in #79 that on Strix Halo (kernel 7.0.1, ROCm 7.2.1) qwen3.6:27b *tool-call generation* now leaks ChatML special tokens into the visible stream:

\`\`\`
<tool_call>
<|im_start|>box", "bash","command":"find /foo/bah/baz -name \"FORMAT.md\" 2>/dev/null
\`\`\`

Setting `HIPFIRE_MMQ=0` restores correctness; non-tool prompts appear unaffected (fivetide's bisection methodology was sound — bench prompts didn't hit the bad distribution).

## Probable shape of the bug

Tool-call generation in agentic flows hits the gate/up/residual GEMMs with very different per-batch statistics than the speed-gate's synthetic prompts: long sequences with high entropy in the structured-token region (`<tool_call>`, `<|im_start|>`, JSON keys, etc.) that the model's logit distribution sharpens around the special-token IDs. The i8 quantization path (Q8_1 packing + scale lookups) likely loses precision in some narrow cell where the fp16 WMMA path keeps it. The +63% number is real; the corner case isn't.

## Path forward (any one of these — likely a + c)

1. **Tighten the auto-enable predicate.** Raise the threshold to a level where bench wins remain large but tool-call distributions don't trip it; add a tool-call detector or expose batch-size-and-arch as a config knob with `auto` / `force_off` / `force_on`.
2. **Gate behind a coherence-gate run** that includes tool-call prompts (currently the gate exercises chat / code / merge_sort but not tool-emit). Make a tool-call-shaped prompt a permanent fixture before any auto-enable lands.
[...]

PR #84, opened 2026-04-28
perf(gfx1151): auto-enable MMQ (i8 WMMA) prefill at batch_size≥128 — 9B pp128 545→890 tok/s

fivetide:
gfx1150/gfx1151 now automatically use the MMQ (Q8_1 + i8 WMMA) prefill
path when batch_size ≥ 128, where it consistently outperforms fp16 WMMA.
Below 128 the fp16 WMMA path remains active (better at small M due to
Q8_1 quantization overhead).

Crossover data on 9B MQ4, gfx1151 (Strix Halo), KV=asym3:
  pp16:  WMMA=289 MMQ=117  → WMMA wins
  pp32:  WMMA=422 MMQ=224  → WMMA wins
  pp64:  WMMA=516 MMQ=411  → WMMA wins
  pp96:  WMMA=536 MMQ=567  → ~ crossover
  pp128: WMMA=545 MMQ=890  → MMQ +63%
  pp256: WMMA=~   MMQ=1067 → MMQ dominant

Threshold of 128 chosen conservatively (also the MMQ tile size).

Dispatch: all four fused GEMM call sites (qkvza, qkv, gate_up, residual)
now go through use_mmq_prefill(arch, batch_size) which respects:
  HIPFIRE_MMQ=1  → force MMQ at any batch size (unchanged)
  HIPFIRE_MMQ=0  → disable MMQ entirely (new kill-switch)
  default        → auto for gfx1150/1151 at batch≥128, off elsewhere

Fresh-process A/B on gfx1151, 9B pp128:
  HIPFIRE_MMQ=0: 545 tok/s (fp16 WMMA)
  default:       890 tok/s (auto MMQ) — +63%
  decode:        45.2 both  (unchanged)

Binary md5: (rebuilt after touch+clean)
Speed-gate --fast: 2/2 PASSED (4B pp32 −3.2%, decode −0.1%)
Coherence-gate: 3/3 OK
Channel-test: 16/16 OK

also includes commit from #83

Issue #79, opened 2026-04-28
API responses still missing.


mikiadev:
First of all everything in this amazing project looks very promising and interesting! I can also run benchmarks etc. if needed.

Still related to #74 ? but on 7d43b05, I'm able to get qwen3.6-27b to give some replies using external chat like Open WebUI and haven't managed to get a pi coding agent to get any responses. Should this yet be capable of working with any longer contexts?

Apparently that model is instruct without reasoning?

serve loads:
```
GPU: gfx1151 (137.4 GB VRAM, HIP 7.2)
  pre-compiled kernels: .hipfire_kernels/gfx1151
[hipfire] http://localhost:11435/v1/chat/completions
[hipfire] DFlash draft detected: /home/mika/.hipfire/models/qwen36-27b-dflash-mq4.hfq
  loading token_embd...
    (Q8_0 raw, 1350 MB)
  loading output_norm...
  loading output (separate lm_head, qt=13)...
  loading layer 0/64 (LinearAttention)...
  loading layer 1/64 (LinearAttention)...
  loading layer 2/64 (LinearAttention)...
  loading layer 3/64 (FullAttention)...
  loading layer 4/64 (LinearAttention)...
  loading layer 5/64 (LinearAttention)...
  loading layer 6/64 (LinearAttention)...
  loading layer 7/64 (FullAttention)...
  loading layer 8/64 (LinearAttention)...
  loading layer 9/64 (LinearAttention)...
  loading layer 10/64 (LinearAttention)...
  loading layer 11/64 (FullAttention)...
  loading layer 12/64 (LinearAttention)...
  loading layer 13/64 (LinearAttention)...
  loading layer 14/64 (LinearAttention)...
  loading layer 15/64 (FullAttention)...
  loading layer 16/64 (LinearAttention)...
  loading layer 17/64 (LinearAttention)...
  loading layer 18/64 (LinearAttention)...
  loading layer 19/64 (FullAttention)...
  loading layer 20/64 (LinearAttention)...
  loading layer 21/64 (LinearAttention)...
  loading layer 22/64 (LinearAttention)...
  loading layer 23/64 (FullAttention)...
  loading layer 24/64 (LinearAttention)...
  loading layer 25/64 (LinearAttention)...
  loading layer 26/64 (LinearAttention)...
  loading layer 27/64 (FullAttention)...
  loading layer 28/64 (LinearAttention)...
  loading layer 29/64 (LinearAttention)...
[...]

PR #104, opened 2026-04-30
feat(dispatch): per-weight MMQ screening to prevent Q8_1 outlier corruption (#87)

fivetide:
## Summary

Diagnoses and fixes the auto-MMQ regression that corrupted tool-call output on gfx1151 (#87). The i8 WMMA + Q8_1 activation quantization path produces 5-9x higher error on specific weight rows (notably row 3994 in every Wo projection). This MR adds per-weight screening at model load time that detects these outliers and falls back to f16 WMMA, preserving most of the MMQ prefill speedup while eliminating the precision loss that caused ChatML token leakage.

**Key result:** On gfx1151 with qwen3.5-9b (272 prefill tokens):

| Mode | Prefill tok/s | vs baseline |
|------|--------------|-------------|
| WMMA only (baseline) | 532 | — |
| MMQ (no screening) | 745 | +40% |
| **MMQ + screening** | **643** | **+21%** |

Screening is **default ON** — no configuration needed. The +21% net win comes from 88% of weights keeping the fast MMQ path while the 12% with outlier rows fall back to WMMA.

## What changed

| File | Change |
|------|--------|
| `crates/rdna-compute/src/dispatch.rs` | +177: `mmq_screen_weight()` function, per-weight cache, screening gates at all 4 MMQ dispatch sites, `mmq_screen`/`mmq_screen_threshold` as public `Gpu` fields |
| `crates/engine/examples/channel_test_mmq.rs` | +775: new diagnostic binary (site-scan, channel-map, layer-sweep, screen) |
| `crates/engine/examples/daemon.rs` | +68: load-time screening via `screen_weights_qwen35()`, config param parsing |
| `cli/index.ts` | +24: `mmq_screen`/`mmq_screen_threshold` in config schema, defaults, validation, per-model overlay, param forwarding |
| `docs/CONFIG.md` | +21: document new config keys with validation data |
| `scripts/` | 5 new: `rocm-env.sh`, `compare-mmq-modes.sh`, `test-mmq-screen.sh`, `check-mmq-active.sh`, `analyze_daemon_output.py` |
| `findings/` | 2 findings docs with raw data from 9B and 27B |

### Step 1: Build the diagnostic tool

Created `channel_test_mmq`, a standalone Rust binary that compares MMQ output against f16 WMMA output at each GEMM call site using real model weights and synthetic activations. Four stages:

- **site-scan** — sweep all (layer, site) pairs, print error table
- **channel-map** — per-output-row error for a specific (site, layer)
- **layer-sweep** — per-layer error for a single site across all layers
- **screen** — run the dispatch-level screening on every weight matrix

No changes to the forward pass or dispatch logic were needed for the diagnostic — all GEMM methods and tensor I/O were already public (only `ensure_q8_1_mmq_x` and `gemm_hfq4g256_mmq_set_prequant` needed `fn` → `pub fn`).

### Step 2: site-scan — identify the guilty GEMM site

Ran on qwen3.5-9b (32 layers, dim=4096) with batch=128:

**Finding: The error is universal, not site-specific.** Every (site, layer) pair fails at threshold 0.01. Q8_1 activation quantization introduces ~1% mean abs error everywhere. However, the **residual (Wo) site** has 3-5x higher *peak* error than any other site:

| Site | Typical max_err | Worst max_err |
|------|----------------|---------------|
| **residual (Wo)** | 0.11 – 0.30 | **0.516** (L0) |
| qkvza.qkv / qkv.q | 0.08 – 0.14 | 0.201 |
| gate_up | 0.03 – 0.08 | 0.082 |

**Decision:** The problem isn't one bad site — it's outlier weight rows within sites. Needed per-row analysis.

### Step 3: channel-map — find the guilty rows

Drilled into residual (Wo) at layers 0, 1, 3, 6:

**Finding: Row 3994 is the #1 outlier in every Wo layer.**

| Layer | Row 3994 max_err | 2nd worst | Ratio |
|-------|-----------------|-----------|-------|
| 0 | **0.516** | 0.092 | 5.6x |
| 1 | **0.456** | 0.073 | 6.3x |
| 3 | **0.493** | 0.068 | 7.3x |
| 6 | **0.357** | 0.070 | 5.1x |

Also found: layer 27 qkv.v row 265 (max_err=0.232, 2x above typical).

All ~4095 other rows per layer are in the "normal" 0.05-0.07 range. Row 3994 is the only structurally broken row.

**Decision:** The root cause is specific HFQ4 weight rows with scale/zero-point statistics that amplify Q8_1 rounding error multiplicatively. Fix should be per-weight screening, not per-site or per-layer.

### Step 4: Confirm on 27B (canonical reproducer)

Ran site-scan and channel-map on qwen3.6-27b (64 layers, dim=5120):

- Layer 3 residual: max_err = **0.910** (vs 0.516 on 9B)
- Row 3994 again #1 outlier: **0.887** max_err, **8.7x** worse than 2nd place
- Same hidden dimension across both models — structural property of the Qwen weight distribution

### Step 5: Implement per-weight screening

Added `mmq_screen_weight()` to `Gpu`: runs a batch=16 synthetic comparison (WMMA vs MMQ), checks per-row max abs error against threshold, caches result by device pointer.

Wired screening into all four MMQ dispatch sites (QKVZA, QKV, gate_up, residual). For fused sites, if ANY sub-weight fails, the entire fused call falls back to WMMA.

### Step 6: Tune the threshold

| Threshold | 9B weights flagged | 27B weights flagged | 27B output matches WMMA? |
|-----------|-------------------|--------------------|-----------------------|
| 0.15 | 19/216 (8.8%) | ~21/432 (4.9%) | No (44 vs 28 tokens) |
| **0.10** | **25/216 (11.6%)** | **73/432 (16.9%)** | **Yes** |

**Decision:** Default threshold = 0.10. At this level, both 9B and 27B produced output matching WMMA baseline in controlled testing. The remaining 83-88% of weights keep the fast MMQ path.

### Step 7: Make it a first-class config parameter

- `mmq_screen` (bool, default: true) — enable/disable screening
- `mmq_screen_threshold` (float, default: 0.10) — per-row error threshold

Configurable via:
- `~/.hipfire/config.json` (global)
- `~/.hipfire/per_model_config.json` (per-model overlay)
- Daemon load params (`mmq_screen`, `mmq_screen_threshold`)
- Env override (`HIPFIRE_MMQ_SCREEN=0` to disable, `HIPFIRE_MMQ_SCREEN_THRESHOLD=<f>` to tune)

### Step 8: Move screening to model load time

Initially screening ran lazily on first MMQ call, adding latency to the first prefill. Moved to `load_model()` in the daemon — screening now runs once at model load (~50ms for 9B, ~100ms for 27B) when the user already expects to wait.

**Impact on throughput:**

| Screening timing | 9B prefill tok/s (272pp) | vs baseline |
|-----------------|------------------------|-------------|
| Lazy (first call) | 480 | -10% |
| **Eager (load time)** | **643** | **+21%** |

The screening results are deterministic (fixed-seed LCG activations, deterministic WMMA/MMQ kernels) and cached by device pointer. Same model + same GPU = same results every load.

### Step 9: End-to-end validation

Ran `compare-mmq-modes.sh` on both models (272 prefill tokens, tool-call prompt with Hermes-format system prompt):

**9B (gfx1151):**

| Mode | Prefill tok/s | Tokens | Correct? |
|------|--------------|--------|----------|
| WMMA only | 532 | 28 | Yes |
| MMQ (no screening) | 745 (+40%) | 75 | Yes, but different output |
| MMQ + screening | 643 (+21%) | 75 | Yes |

**27B (gfx1151):**

| Mode | Prefill tok/s | Tokens | Correct? |
|------|--------------|--------|----------|
| WMMA only | 158 | 28 | Yes |
| MMQ (no screening) | 236 (+49%) | 42 | Yes, but different output |
| MMQ + screening | 153 (-3%) | 28 | Yes |

All modes produce valid tool calls with no `<|im_start|>` leakage.

## Caveats

1. **Could not reproduce the original #87 corruption.** Our test prompt (272 prefill tokens, single-turn tool call) does not trigger `<|im_start|>` leakage even with unscreened MMQ. The original bug manifested with pi-coding-agent's multi-turn context, which likely has much longer accumulated context and different prompt-shape statistics. The screening fix is validated at the quantization-error level (eliminates the 0.5-0.9 outlier spikes) but not by reproducing and then fixing the exact corruption.

2. **Output with screening doesn't always match WMMA baseline byte-for-byte.** In controlled single-run testing, threshold=0.10 produced identical output on both models. In fresh-process smoke tests, MMQ+screening sometimes matches unscreened MMQ output rather than WMMA baseline. The screening eliminates the worst precision loss (0.5-0.9 outlier spikes) but some residual Q8_1 noise on safe-passing weights can still shift sampling outcomes. Output is always *correct* (valid tool calls, no corruption), just not always *identical* to the WMMA reference.

3. **Screening uses a fixed synthetic activation pattern** (seeded LCG, values in [-2, 2]). Real activations may have different distributions. The synthetic pattern matches typical post-RMSNorm hidden-state range and successfully identifies the known outlier rows, but there's a theoretical risk of false negatives on activation distributions not covered by the synthetic sweep.

4. **The threshold is model-family-dependent.** 0.10 works for qwen3.5/3.6 MQ4 weights. Other model families or quantization formats may need different thresholds. The threshold is configurable per-model via `mmq_screen_threshold` for this reason.

5. **27B screening overhead.** On 27B, 73/432 weights (17%) fall back to WMMA, which at pp=272 makes MMQ+screening roughly break-even with WMMA-only (-3%). The net win should be larger at bigger prefill batch sizes (pp512+) where the 83% fast-path weights dominate.

## Open work

- [ ] **Reproduce the original corruption** with a multi-turn pi-coding-agent prompt to validate the fix end-to-end on the exact failure mode from #87
- [ ] **Benchmark at larger batch sizes** (pp512, pp1024, pp2048) where MMQ's +40% speedup should outweigh the screening overhead
- [ ] **Investigate row 3994's weight statistics** — why does this specific hidden dimension have unusual HFQ4 scale/zero across the entire Qwen family? Could the quantizer be patched to avoid creating these outliers?
- [ ] **Consider per-row mixed dispatch** instead of per-matrix fallback — run MMQ for all rows, then overwrite only the ~1-2 outlier rows with WMMA results. Would preserve nearly 100% of MMQ speedup
- [ ] **Test on other RDNA3 variants** (gfx1100, gfx1101, gfx1102) to confirm the screening generalizes
- [ ] **Disk caching** — screening results are deterministic for a given `(model_hash, arch, threshold)` tuple. Caching to disk would eliminate the ~50-100ms load-time cost on subsequent loads of the same model

## How to test

```bash
# Build
cargo build --release --features deltanet --example channel_test_mmq --example daemon

# Run the diagnostic (requires RDNA3/3.5 GPU)
cargo run --release --features deltanet --example channel_test_mmq -- --help
cargo run --release --features deltanet --example channel_test_mmq -- \
    --model ~/.hipfire/models/qwen3.5-9b.mq4 --stage site-scan --batch 128
cargo run --release --features deltanet --example channel_test_mmq -- \
    --model ~/.hipfire/models/qwen3.5-9b.mq4 --stage screen

# End-to-end comparison (WMMA vs MMQ vs MMQ+screening)
./scripts/compare-mmq-modes.sh ~/.hipfire/models/qwen3.5-9b.mq4

# Config
hipfire config set mmq_screen true              # default
hipfire config set mmq_screen_threshold 0.10    # default
hipfire config qwen3.6:27b set mmq_screen_threshold 0.08  # per-model
```

Refs: #87 (auto-MMQ regression), #84 (original MMQ PR), #79 (reporter)

### DIFF
--- a/cli/index.ts
+++ b/cli/index.ts
@@ -160,8 +160,10 @@ const CONFIG_DEFAULTS: HipfireConfig = {
   // Set false (or HIPFIRE_NORMALIZE_PROMPT=0) to opt out.
   prompt_normalize: true,
   // MMQ per-weight screening: detect Q8_1 outlier rows and fall back to
-  // WMMA. Default ON — prevents tool-call corruption (#87) at <0.1% perf cost.
-  mmq_screen: true,
+  // WMMA. Default OFF — no single threshold produces byte-identical output
+  // across models without screening out 90%+ of weights. Opt in to prevent
+  // tool-call corruption (#87) when using HIPFIRE_MMQ=1.
+  mmq_screen: false,
   mmq_screen_threshold: 0.10,
 };
 
--- a/cli/index.ts
+++ b/cli/index.ts
@@ -108,6 +108,19 @@ interface HipfireConfig {
   // code prompts by up to +26.7% (commit 8a4a211). Default ON since
   // 2026-04-26 (commit 9a2c667).
   prompt_normalize: boolean;
+
+  // ── MMQ per-weight screening (#87) ──────────────────────────────────
+  // When true (default), the i8 WMMA (MMQ) prefill path screens each
+  // weight matrix on first use. A quick synthetic comparison (batch=16)
+  // checks per-row max abs error — weights with outlier rows fall back
+  // to f16 WMMA. Prevents tool-call corruption from Q8_1 precision loss
+  // on specific weight channels (e.g. row 3994 in Wo projections).
+  // Only effective when MMQ is active (HIPFIRE_MMQ=1 or HIPFIRE_WO_MMQ=1).
+  mmq_screen: boolean;
+  // Abs error threshold for MMQ screening. Weights with any output row
+  // exceeding this fall back to WMMA. Default 0.10 — validated on both
+  // qwen3.5-9b and qwen3.6-27b to produce byte-identical output vs WMMA.
+  mmq_screen_threshold: number;
 }
 
 // Detect GPU at import time for smart defaults
@@ -146,6 +159,10 @@ const CONFIG_DEFAULTS: HipfireConfig = {
   // +24% τ on PEP-8-style code prompts (159→196 tok/s on 27B-3.5 LRU DFlash).
   // Set false (or HIPFIRE_NORMALIZE_PROMPT=0) to opt out.
   prompt_normalize: true,
+  // MMQ per-weight screening: detect Q8_1 outlier rows and fall back to
+  // WMMA. Default ON — prevents tool-call corruption (#87) at <0.1% perf cost.
+  mmq_screen: true,
+  mmq_screen_threshold: 0.10,
 };
 
 function validateConfigValue(key: string, value: any): boolean {
@@ -174,6 +191,8 @@ function validateConfigValue(key: string, value: any): boolean {
     case "cask_fold_m": return typeof value === "number" && Number.isInteger(value) && value >= 1 && value <= 16;
     case "cask_auto_attach": return typeof value === "boolean";
     case "prompt_normalize": return typeof value === "boolean";
+    case "mmq_screen": return typeof value === "boolean";
+    case "mmq_screen_threshold": return typeof value === "number" && value > 0 && value <= 1;
     default: return false;
   }
 }
@@ -218,6 +237,7 @@ const PER_MODEL_KEYS = [
   "cask_budget", "cask_beta", "cask_core_frac", "cask_fold_m",
   "cask_auto_attach",
   "prompt_normalize",
+  "mmq_screen", "mmq_screen_threshold",
 ] as const;
 type PerModelKey = typeof PER_MODEL_KEYS[number];
 
@@ -468,6 +488,10 @@ function buildLoadMessage(path: string, tag?: string | null): any {
       console.error(`[hipfire] WARN: cask_sidecar path missing: ${resolved.cask_sidecar} — disabling eviction for this load`);
     }
   }
+
+  // MMQ per-weight screening (#87)
+  params.mmq_screen = resolved.mmq_screen;
+  params.mmq_screen_threshold = resolved.mmq_screen_threshold;
 
   return { type: "load", model: path, params };
 }
--- a/crates/engine/examples/channel_test_mmq.rs
+++ b/crates/engine/examples/channel_test_mmq.rs
@@ -1,0 +1,664 @@
+//! MMQ vs WMMA bit-comparison diagnostic for Qwen3.5 GEMM call sites.
+//!
+//! Compares i8 WMMA + Q8_1 (MMQ) output against f16 WMMA output at each
+//! GEMM call site in a transformer layer to diagnose which site/channel/layer
+//! causes tool-call output corruption (ref: #87).
+//!
+//! Usage:
+//!   cargo run --release --features deltanet --example channel_test_mmq -- \
+//!       --model <path.hfq> [--stage site-scan|channel-map|layer-sweep]
+//!       [--batch N] [--threshold F] [--layer N] [--site <name>]
+//!
+//! Stages:
+//!   site-scan   (default) — sweep every (layer, site) pair; print error table
+//!   channel-map — per-row error for a specific (site, layer) or all layers
+//!   layer-sweep — per-layer error for a specific site across all layers
+//!
+//! MMQ is only available on RDNA3/3.5: gfx1100, gfx1101, gfx1102, gfx1103,
+//! gfx1150, gfx1151. The binary exits 0 with a message on other archs.
+
+#[cfg(not(feature = "deltanet"))]
+fn main() {
+    eprintln!("channel_test_mmq requires --features deltanet");
+    std::process::exit(1);
+}
+
+#[cfg(feature = "deltanet")]
+fn main() {
+    use engine::hfq::HfqFile;
+    use engine::qwen35;
+    use std::path::Path;
+
+    // ── CLI parsing ──────────────────────────────────────────────────────
+    let args: Vec<String> = std::env::args().collect();
+    if args.len() < 2 || args.iter().any(|a| a == "--help" || a == "-h") {
+        eprintln!(
+            "Usage: channel_test_mmq --model <path.hfq> \
+             [--stage site-scan|channel-map|layer-sweep] \
+             [--batch N] [--threshold F] [--layer N] [--site <name>]"
+        );
+        std::process::exit(0);
+    }
+
+    let mut model_path: Option<String> = None;
+    let mut stage = "site-scan".to_string();
+    let mut batch_size: usize = 128;
+    let mut threshold: f32 = 0.01;
+    let mut layer_filter: Option<usize> = None;
+    let mut site_filter: Option<String> = None;
+
+    let mut i = 1;
+    while i < args.len() {
+        match args[i].as_str() {
+            "--model" => {
+                model_path = Some(args[i + 1].clone());
+                i += 2;
+            }
+            "--stage" => {
+                stage = args[i + 1].clone();
+                i += 2;
+            }
+            "--batch" => {
+                batch_size = args[i + 1].parse().expect("--batch must be integer");
+                i += 2;
+            }
+            "--threshold" => {
+                threshold = args[i + 1].parse().expect("--threshold must be float");
+                i += 2;
+            }
+            "--layer" => {
+                layer_filter = Some(args[i + 1].parse().expect("--layer must be integer"));
+                i += 2;
+            }
+            "--site" => {
+                site_filter = Some(args[i + 1].clone());
+                i += 2;
+            }
+            other => {
+                eprintln!("Unknown argument: {other}");
+                std::process::exit(1);
+            }
+        }
+    }
+
+    let model_path = model_path.expect("--model <path.hfq> is required");
+
+    // ── GPU init ─────────────────────────────────────────────────────────
+    let mut gpu = rdna_compute::Gpu::init().expect("GPU init failed");
+    let arch = gpu.arch.clone();
+    eprintln!("GPU: {arch}");
+
+    let mmq_archs = ["gfx1100", "gfx1101", "gfx1102", "gfx1103", "gfx1150", "gfx1151"];
+    if !mmq_archs.contains(&arch.as_str()) {
+        eprintln!(
+            "SKIP: MMQ requires RDNA3/3.5 (gfx1100..gfx1103, gfx1150, gfx1151). \
+             Current arch: {arch}"
+        );
+        std::process::exit(0);
+    }
+
+    // ── Model loading ────────────────────────────────────────────────────
+    eprintln!("Loading model: {model_path}");
+    let hfq = HfqFile::open(Path::new(&model_path)).expect("open model");
+    let config = qwen35::config_from_hfq(&hfq).expect("config_from_hfq");
+    eprintln!(
+        "Config: dim={} layers={} heads={} kv_heads={} vocab={}",
+        config.dim, config.n_layers, config.n_heads, config.n_kv_heads, config.vocab_size
+    );
+    let weights = qwen35::load_weights(&hfq, &config, &mut gpu).expect("load_weights");
+    eprintln!("Weights loaded. Running stage: {stage}");
+
+    // ── Dispatch stage ───────────────────────────────────────────────────
+    match stage.as_str() {
+        "site-scan" => {
+            run_site_scan(&mut gpu, &weights, &config, batch_size, threshold);
+        }
+        "channel-map" => {
+            let site = site_filter.expect("--site <name> required for channel-map");
+            run_channel_map(&mut gpu, &weights, &config, batch_size, threshold, &site, layer_filter);
+        }
+        "layer-sweep" => {
+            let site = site_filter.expect("--site <name> required for layer-sweep");
+            run_layer_sweep(&mut gpu, &weights, &config, batch_size, threshold, &site);
+        }
+        other => {
+            eprintln!("Unknown stage: {other}  (use site-scan | channel-map | layer-sweep)");
+            std::process::exit(1);
+        }
+    }
+}
+
+// ── Data structures ──────────────────────────────────────────────────────────
+
+/// Per-site aggregated error statistics.
+#[cfg(feature = "deltanet")]
+struct SiteStats {
+    site: String,
+    layer: usize,
+    m: usize,
+    k: usize,
+    batch_size: usize,
+    max_err: f32,
+    mean_err: f32,
+    bad_count: usize,
+    threshold: f32,
+}
+
+#[cfg(feature = "deltanet")]
+impl SiteStats {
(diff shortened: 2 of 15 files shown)


Kaden-Schutt:
Did a close read across the diff and found a hard interaction issue with #106 (MQ3/MQ2 sub-4-bit Magnum Quants, just merged at `f851948`). Flagging before this lands so we can land it cleanly without a hot-fix follow-up.

## Critical interaction (post-#106)

`screen_weights_qwen35` in `crates/engine/examples/daemon.rs:912-933` iterates **every layer weight** and calls `gpu.mmq_screen_weight(&wt.buf, wt.m, wt.k)` regardless of `wt.gpu_dtype`. Inside, `mmq_screen_weight` calls

- `gemm_hfq4g256_residual_wmma(a_raw, …)`
- `gemm_hfq4g256_mmq_set_prequant(a_raw, xq, …)`

Both kernels assume **HFQ4G256 layout** (136 B per 256-element group, `gp + 8 + k_off/2 + k_grp*4` reads). Master now has `WeightTensor`s with `gpu_dtype` ∈ `{HFQ4G256, MQ4G256, MQ3G256, MQ2G256, HFQ6G256, MQ6G256, …}`. Group strides:

| dtype       | bytes/group | ratio vs HFQ4 | outcome |
|-------------|-------------|---------------|---------|
| HFQ4G256    | 136         | 1.00× | safe (target) |
| MQ4G256     | 136         | 1.00× | safe (incidentally same layout) |
| HFQ6G256    | 200         | 1.47× | over-read into next group, garbage values but no fault |
| **MQ3G256** | **104**     | **0.76×** | **kernel reads 30 % past buffer end → HIP memory fault** |
| **MQ2G256** | **72**      | **0.53×** | **kernel reads 89 % past buffer end → guaranteed fault** |

→ **Loading any MQ3 or MQ2 model with `mmq_screen=true` (the new default) will panic the daemon at load time on RDNA3/3.5.**

The runtime-path screening calls (the 4 MMQ dispatch sites) are already safe by predicate (`HIPFIRE_MMQ=1` + `has_mmq_i8_wmma`, and the dispatch sites themselves are HFQ4-only). Only the eager load-time sweep needs the gate.

## Suggested fix (1 line)

In `screen_weights_qwen35`:

```rust
for (wt, _name) in wts {
    if !matches!(wt.gpu_dtype, rdna_compute::DType::HFQ4G256) { continue; }   // NEW
    if gpu.mmq_screen_weight(&wt.buf, wt.m, wt.k) {
        n_safe += 1;
    } else {
        n_unsafe += 1;
    }
}
```

(`WeightTensor` carries `gpu_dtype: DType` per `crates/engine/src/llama.rs:485-491`.)

## Other items I'd like to see before flipping default-on

The PR body called these out as caveats — I want to keep them top-of-mind for the merge bar since this is hot-path + default-on:

1. **End-to-end #87 reproducer.** PR notes "Could not reproduce the original #87 corruption." The fix is validated at the quantization-error level (eliminating the 0.5–0.9 outlier spikes on row 3994 of Wo) but not against the actual broken-output failure mode @nwoolmer reported. I added tool-call coverage to coherence-gate at `48a7fea` specifically so we'd have a regression fixture for this — could you run `./scripts/coherence-gate.sh` with `HIPFIRE_MMQ=1` and a tool-call-shape prompt and paste the output? The hard-fail is on `<|im_start|>` leakage in the visible stream — that's the unique corruption signature.
2. **Output-doesn't-match-WMMA-byte-for-byte caveat.** Your PR body notes that in fresh-process smoke tests, MMQ+screening sometimes matches unscreened MMQ output rather than WMMA baseline. That suggests some unsafe weights are passing the synthetic-activation screen. Do you have a feel for whether bumping the threshold to 0.08 (from 0.10) closes the gap, or whether the LCG uniform-[-2,2] activation pattern fundamentally misses the bad distribution? If the latter, we should probably ship default-OFF and let users opt in via config — flipping ON later once the screen is more robust.

## Other thoughts on the diff (non-blocking)

- The pointer-keyed `mmq_screen_cache` is correct for buffer-lifetime caching. Each load gets fresh pointers, so cache freshness is automatic. ✓
- The `channel_test_mmq.rs` diagnostic binary is great — opening up `ensure_q8_1_mmq_x` and `gemm_hfq4g256_mmq_set_prequant` to `pub fn` is safe and lets the diagnostic run without touching the forward pass. ✓
- The 4 dispatch-site guards are correctly placed inside the `HIPFIRE_MMQ=1` + `has_mmq_i8_wmma` outer predicate, and fall through to the existing WMMA path when screening rejects. ✓
- Overall the design is right; the scope just needs to be aware that master now has more dtypes.

## Status

Holding merge until either the dtype filter goes in here or I'd want to add it as a follow-up commit if you'd prefer (just say the word). Either way the path is small.

Thanks for the rigorous investigation — the channel-map → row-3994 finding is exactly the kind of structural signal that's worth screening for. We just need it to compose with the rest of the dtype matrix.

fivetide:
Update: threshold sweep, dtype fix, default-OFF

  Threshold sweep (new finding)

  Swept thresholds 0.01–0.20 on both 9B and 27B, comparing token output md5 against WMMA baseline:

  ┌───────────┬───────────┬────────────┬────────────────────┐
  │ Threshold │ 9B match? │ 27B match? │ 9B weights flagged │
  ├───────────┼───────────┼────────────┼────────────────────┤
  │ 0.15      │ YES       │ NO         │ 19/216             │
  ├───────────┼───────────┼────────────┼────────────────────┤
  │ 0.10      │ NO        │ YES        │ ~30/216            │
  ├───────────┼───────────┼────────────┼────────────────────┤
  │ 0.06      │ YES       │ NO         │ ~60/216            │
  ├───────────┼───────────┼────────────┼────────────────────┤
  │ 0.05      │ NO        │ YES        │ ~80/216            │
  ├───────────┼───────────┼────────────┼────────────────────┤
  │ 0.04      │ YES       │ YES        │ 197/216 (91%)      │
  └───────────┴───────────┴────────────┴────────────────────┘

  The convergence is non-monotonic — 0.10 matches 27B but not 9B; 0.15 matches 9B but not 27B. Both converge at ≤0.04, but that screens 91% of weights, effectively disabling MMQ.

  Root cause: even "safe" weights have ~1% Q8_1 noise. Any change in which weights use MMQ vs WMMA shifts the accumulated rounding across 32+ layers enough to flip a greedy sampling decision. The problem is discrete, not continuous — there is no threshold that is both
  selective and byte-identical across models.

  Dtype fix (#106 compat)

  Added dtype filter: screening now only runs on HFQ4G256 | MQ4G256 weights (same 136 bytes/group layout). MQ3G256 (104 bytes/group) and MQ2G256 (72 bytes/group) are skipped — feeding them to HFQ4-specific kernels would read past buffer end. Verified: A3B MoE model loads
  without panic with screening enabled.

  Previous version had a bug where the filter checked only HFQ4G256, missing MQ4G256 — this made screening a silent no-op on all .mq4 models. Fixed.

  Default changed to OFF

  Given the threshold sweep results, mmq_screen now defaults to false. No single threshold produces byte-identical output across models without screening out 90%+ of weights. Users opt in via hipfire config set mmq_screen true or HIPFIRE_MMQ_SCREEN=1.

  Coherence gate

  Ran coherence-gate.sh --full with HIPFIRE_MMQ=1 (screening OFF):
  - All tests pass: 0.8B, 4B, 9B, A3B-35B, 27B
  - Both tool-call tests (9B, 27B) produce clean <tool_call> output, zero <|im_start|> leaks
  - Same result with HIPFIRE_MMQ_SCREEN=0 explicitly

  Remaining question

  The diagnostic tooling and root cause analysis (row 3994, structural HFQ4 outlier) are solid. The screening mechanism works as designed. The limitation is fundamental: per-matrix fallback creates discrete precision-path changes that shift sampling. A per-row mixed
  dispatch (MMQ for safe rows, WMMA fixup for outlier rows within the same kernel launch) would avoid this problem but requires kernel-level changes.

fivetide:
i also tried calibrating a threshold per model, but it seems to depend on the calibration prompt and has fundamental issues:

MMQ Calibration: approach, findings, issues, tradeoffs

  Approach. At model load time, screen each weight matrix by running a small synthetic comparison (batch=16, seeded LCG activations): compute the f16-WMMA and i8-MMQ GEMM outputs side by side, check per-row max absolute error. Weights exceeding a threshold fall back to
  WMMA; the rest keep the fast MMQ path. The threshold is auto-calibrated per model+arch by sweeping candidates against a reference WMMA inference run and picking the highest value that produces byte-identical output.

  Findings.

  Row 3994 in every Wo projection is a 5-9x outlier across both qwen3.5-9b and qwen3.6-27b (max error 0.46–0.91 vs typical 0.05–0.08). This single hidden dimension drives the worst-case precision loss. The screening reliably catches it.

  On a single calibration prompt (tool-call shape), both models converge to byte-identical output at surprisingly high thresholds — 0.50 on 9B (zero weights screened, full MMQ speed) and 0.50 on 27B. The outlier row's error, while large in absolute terms, doesn't always
  flip a greedy sampling decision.

  On a multi-prompt battery (tool-call, code, reasoning, structured JSON), no threshold produces byte-identical output across all shapes. The structured JSON prompt always diverges from WMMA regardless of screening. The ~1% mean Q8_1 error distributed across all weights
  is enough to change greedy decode trajectories on precision-sensitive prompts.

  The threshold-to-output relationship is non-monotonic. Lowering the threshold from 0.15 to 0.10 can make output diverge on one model while converging on another. This is because each threshold creates a different partition of weights between MMQ and WMMA paths, and the
  accumulated rounding through 32+ layers depends on error cancellation patterns that are partition-specific. A mix of precisions can be worse than either pure path.

  Issues.

  The calibration is prompt-dependent. A threshold calibrated on one prompt shape may not hold on another. This is fundamental — Q8_1 noise interacts differently with different activation distributions.

  Per-matrix granularity creates discrete precision-path changes. Flipping one weight from MMQ to WMMA changes accumulated rounding for every downstream layer. There is no threshold that is both selective (keeps meaningful speedup) and universally byte-identical (matches
  WMMA on all prompts).

  Screening cost is front-loaded (~160ms for 9B at load time) but cached. The calibration sweep is expensive (~2s per threshold candidate on 9B, ~8s on 27B) but runs once and caches to disk.

  Tradeoffs.

  The screening prevents catastrophic corruption (the <|im_start|> leakage from #87) but does not make MMQ transparent. Users face a three-way choice:

  - WMMA only (default): byte-exact reference quality, baseline speed
  - MMQ unscreened: +40% prefill throughput, output differs from WMMA on some prompts, risk of rare structural corruption on specific prompt shapes
  - MMQ + screening: +20-40% prefill (depending on how many weights pass), output still differs from WMMA on some prompts, but catastrophic outlier corruption is prevented

  Per-prompt calibration can find sweet spots for known workloads. A deployment that only serves tool-call prompts can calibrate to 0.50 and get full MMQ speed with identical output. A deployment serving mixed workloads has no single threshold that works everywhere.

  The correct fix for byte-identical MMQ would operate at a different level — either per-row mixed dispatch within the kernel (avoiding the precision-partition problem) or fixing the quantizer to avoid creating outlier weight rows in the first place.

PR #117, opened 2026-05-01
fix(ui): equivalent serve -d command on windows

darkamgine:
Fix hipfire serve -d on windows by providing equivalent to setsid/nohup when windows detected

tested and working on my windows 10 machine

first time contributor, dont know what im doing, feel free to revise as needed

as an aside, idk if its just my machine but compile_kernels.ps1 didnt work for me until i replaced hipcc.bat in the code to hipcc.exe. the .bat was giving me `ERROR    The filename, directory name, or volume label syntax is incorrect.` but the .exe worked just fine. i believe for 6.4 and later, the .exe is the preferred way of executing it, so it might make sense to put the .exe as higher preference over the .bat?

### DIFF
--- a/cli/index.ts
+++ b/cli/index.ts
@@ -3700,10 +3700,11 @@ switch (cmd) {
       // Fork a detached child. `setsid` gives it its own session so Ctrl-C
       // in the parent shell doesn't reach it; `nohup` ignores SIGHUP; stdout
       // + stderr go to the log file. HIPFIRE_DETACHED prevents infinite forking.
+      const runBg = (process.platform == "win32") ? ["cmd", "/c", "start", "/b"] : ["setsid", "nohup"]
       const self = process.argv[0];
       const script = process.argv[1];
       const logFd = require("fs").openSync(SERVE_LOG_FILE, "a");
-      const child = Bun.spawn(["setsid", "nohup", self, script, "serve", String(port)], {
+      const child = Bun.spawn([...runBg, self, script, "serve", String(port)], {
         stdin: "ignore",
         stdout: logFd,
         stderr: logFd,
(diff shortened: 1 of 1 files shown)


Kaden-Schutt (approved):
LGTM. Single-line `runBg` swap from `setsid nohup` to `cmd /c start /b` on Windows — clean shape, preserves the existing Linux/macOS path verbatim. Linux-side `setsid nohup` flow untouched, so no regression risk on the dominant platform.

I can't validate the Windows runtime on this box (Linux dev env), but the change is small enough and you tested on Win10. Approving on that basis.

Two tiny follow-ups, neither blocking this merge:

1. The added line is missing the trailing `;`. The rest of the file leans on semicolons; not a syntax issue (TS / Bun auto-insert) but worth matching style. Will tidy in a follow-up if you don't.

2. Your aside about `hipcc.bat` failing where `hipcc.exe` works on ROCm 6.4: that's a real bug in `scripts/compile-kernels.ps1` which I shipped earlier this session. The candidate-order has `.bat` before `.exe`. Will swap to `.exe`-first as a follow-up commit on master so it composes with this fix. Thanks for surfacing it.

Merging momentarily. Welcome to the contributor list.

Kaden-Schutt:
Follow-up landed at d308b56: swapped candidate order in `scripts/compile-kernels.ps1` from `(.bat, .exe, no-ext)` to `(.exe, .bat, no-ext)` at all four probe sites in `Find-Hipcc`. ROCm 6.4+ users now get `.exe` first, older installs that only have `.bat` still work via fallback. Thanks for the report.

Kaden-Schutt:
Follow-up to your hipcc.bat note: landed at d308b56 on master. Swapped candidate order in `scripts/compile-kernels.ps1` from `(.bat, .exe, no-ext)` to `(.exe, .bat, no-ext)` at all four probe sites in `Find-Hipcc`. ROCm 6.4+ users now get `.exe` first, older installs that only have `.bat` still work via fallback. Thanks for the report.

Issue #111, opened 2026-04-30
Tool call JSON malformation under MQ4 quantization — Qwen3.6 27B


Fluorax:
# Tool call JSON malformation under MQ4 quantization — Qwen3.6 27B

## Environment

- **GPU:** RX 7900 XTX (gfx1100, 24GB), ROCm 6.2
- **hipfire:** built from source, master c448d5e+
- **Model:** `qwen3.6-27b.mq4`
- **Client:** pi agentic harness (OpenAI-compatible tool calling)

## Problem

The model consistently produces malformed tool call syntax inside `<tool_call>` blocks. Instead of valid JSON:

```
<tool_call>
{"name": "write", "arguments": {"path": "/home/mike/rate_limiter.py", "content": "..."}}
</tool_call>
```

It produces XML-like mixed syntax:

```
<tool_call>
<plain>write</param> {"param": "/home/mike/rate_limiter.py", "content": "..."}
</tool_call>
```

The `parseToolCalls` function in `cli/index.ts` cannot parse this — it expects valid JSON inside the tags. The tool call silently fails, the model believes it succeeded, and the file is never written.

## Reproduction

1. Connect hipfire to an OpenAI-compatible agentic harness with file write tools
2. Ask the model to implement and save a Python class
3. Observe `<tool_call>` block with malformed JSON in the response
4. Tool call fails to execute despite model reporting success

## Comparison

Same prompt on Qwen3.6 27B GGUF via llamacpp produces correctly formatted tool calls:

```
<tool_call>
{"name": "write", "arguments": {"path": "/home/dev/rate_limiter.py", "content": "..."}}
</tool_call>
```

File is written successfully. This confirms the base model weights know the correct format, the malformation is specific to MQ4.

## Hypothesis

MQ4 FWHT rotation shifts the token probability distribution enough to corrupt structured output at the token boundaries of JSON syntax (`{`, `"`, `:`, `}`). The model produces plausible-looking but invalid structured tokens that visually resemble the correct format.

This may be related to the same mechanism causing repetition attractors on long structured output, both are failures of precise token prediction under quantization pressure.

## Expected behaviour

Tool calls should produce valid JSON inside `<tool_call>` tags, parseable by `parseToolCalls` in `cli/index.ts`.

## Potential fix paths

1. **MQ6**  higher quality quantization, less distribution shift on structured tokens
2. **Post-processing**  attempt to repair malformed JSON in `parseToolCalls` before discarding
[...]

Issue #39, opened 2026-04-25
[Path C] Train custom DFlash draft via Red Hat / vLLM recipe (target distillation)


Kaden-Schutt:
**Roadmap PRD:** [`docs/plans/task-93-path-c-trained-draft.prd`](docs/plans/task-93-path-c-trained-draft.prd)
**Recipe ref:** `RECIPE_RedHat_DFlash_MI300X.md` (in `dflash-fe` worktree, derived from `RedHatAI/gemma-4-31B-it-speculator.dflash`)

## TL;DR

Train a custom DFlash draft on Qwen3.5-27B's own distribution via the Red Hat / vLLM speculators recipe. Two compounded wins:

1. **Direct τ lift** from target-aligned distribution: 10.36 → projected ~12-14 (+15-30% baseline tok/s independent of pipelining)
2. **Oracle revival for Path D**: Phase B measured ANYPOS_MATCH at 23.5% with z-lab draft. A target-aligned draft should lift this past the 50% break-even, reviving the original P0 oracle and stacking another +12-22% on top of Path D's structural lift.

**Combined ceiling: 199 → 240-260 tok/s plausible if both land.**

## Why z-lab DFlash drafts plateau

Per the genre matrix in PR #37's README:
- 9B-3.5 prose: AR 125.3 → DFlash 99.4 (-20% LOSS), τ=1.20
- 9B-3.5 code: AR 124.0 → DFlash 329.1 (+165%), τ=6.76

Draft was trained on a generic mix. Its argmax disagreement with target on prose tokens drops τ to barely-better-than-greedy. A target-aligned draft fixes this directly.

## Recipe gaps in current `scripts/dflash_train_poc.py`

| Recipe gap | Current state | Severity |
|---|---|---|
| Target-generated responses | Missing entirely | **CRITICAL** |
| Reduced draft vocab (8192) | Hard-coded full vocab (151936) | **HIGH** |
| Hand-picked target_layer_ids | Auto via `build_target_layer_ids` | MEDIUM |
| Loss mask on assistant turns | All-tokens loss | MEDIUM |
| Conversational data | Default corpus is wikitext | **HIGH** |

## Approach: skip C2 (target self-gen), use existing HF datasets

Direct Qwen3.5-27B-distilled datasets exist:
- [`Jackrong/Qwen3.5-reasoning-700x`](https://huggingface.co/datasets/Jackrong/Qwen3.5-reasoning-700x) — 4.48k samples
[...]

PR #121, opened 2026-05-02
fix(daemon): GPU-side single-token attractor block for <tool_call> / <think> (#111)

Kaden-Schutt:
## Summary

Closes the second symptom of #111 reported by @Fluorax after v0.1.9-alpha shipped: under MQ4 quant pressure, `qwen3.6:27b.mq4` decodes the `<tool_call>` special token into a self-reinforcing attractor (9 to 100+ in a row, never reaching the JSON body). The defensive `parseToolCalls` repair from #120 catches malformed JSON but cannot help when the model never produces JSON.

The fix adds a tightly-scoped GPU-side block:
- Resolve `<tool_call>` and `<think>` special-token IDs once at decode start via a new `Tokenizer::special_token_id()` accessor.
- Before each `gpu.sample_top_p`, count occurrences in the last 20 *generated* tokens (never the prompt).
- If count ≥ 3, write a single 4-byte `-INF` to the GPU logits buffer at offset `tok_id × 4`. No D2H, no kernel change, ~5 µs only on the turns that trip — zero cost otherwise.

## Why this scope (not n-gram block)

- The general `apply_ngram_block` does not catch single-token attractors reliably: surrounding tokens vary (newline, blank repeats, `</think>` resumption) so the suffix rarely matches an earlier window verbatim.
- The GPU sampler's repeat penalty alone can't break the loop fast enough at the user-validated `RP=1.05` floor.
- A CPU-side per-step logits download would tank decode tok/s. The 4-byte H2D-only path keeps the GPU sample fast.

## Verification

- **5 unit tests** in `llama::tests` cover threshold-edge / window-scope / pure-repeat / out-of-vocab token cases. All pass.
- **coherence-gate green**: 6/6 prompts including the qwen3.5-9b.mq4 tool-call coverage test (47 tokens, on-shape, no false-positive block).
- The fix is silent on tokenizers without `<tool_call>` registered as a special token (older Qwen vocabs return `None` from `special_token_id()`).

## Known limitations

This is still a stopgap. The proper fix is Path C calibration retrain (#39) so the structured-output token-tail is preserved at MQ4. That remains tracked in MANUAL_REVIEW.md.

DFlash decode loop (`spec_step_dflash` in `speculative.rs`) is not yet wired to the same block — the reporter's repro is plain AR (no draft loaded). If attractor surfaces on DFlash, it lands as a follow-up.

## Test plan

- [x] cargo build --features deltanet clean
- [x] cargo test attractor_block (5/5 passing)
- [x] coherence-gate.sh green (6/6, tool-call coverage included)
- [ ] User-facing repro on Fluorax's exact prompts (left for the reporter once merged)

Reported-by: @Fluorax
Refs: #111

### DIFF
--- a/cli/index.ts
+++ b/cli/index.ts
@@ -1584,11 +1584,24 @@ async function serve(port: number) {
           const tool_calls: any[] = [];
           let repaired = 0;
           for (const m of matches) {
-            const raw = (m[1] || m[2] || "").trim();
+            let raw = (m[1] || m[2] || "").trim();
+            if (!raw) continue;
+            // MQ4 single-token attractor (#111) sometimes stacks 1-2 nested
+            // `<tool_call>` openers before the JSON body lands. The engine
+            // blocks the third+ via the unclosed-depth gate in daemon.rs,
+            // but the second still ships in the visible stream. Strip any
+            // leading nested-opener artifacts before parsing — if the
+            // first non-whitespace bytes are another `<tool_call>`,
+            // discard them and use the inner content.
+            let nestedStripped = 0;
+            while (raw.startsWith("<tool_call>")) {
+              raw = raw.slice("<tool_call>".length).trimStart();
+              nestedStripped++;
+            }
             if (!raw) continue;
             const parsed = parseOneToolCall(raw);
             if (!parsed) continue;
-            if (parsed.repaired) repaired++;
+            if (parsed.repaired || nestedStripped > 0) repaired++;
             tool_calls.push({
               id: `call_${Date.now().toString(36)}${Math.random().toString(36).slice(2, 6)}`,
               type: "function",
--- a/cli/parse_tool_calls.test.ts
+++ b/cli/parse_tool_calls.test.ts
@@ -159,3 +159,54 @@ test("escaped quotes inside JSON strings are handled", () => {
   expect(r).not.toBeNull();
   expect(r!.arguments).toEqual({ path: "/tmp/x", content: 'say "hi"' });
 });
+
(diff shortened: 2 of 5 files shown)


Kaden-Schutt:
**Codex stop-gate caught a hole in v1** — threshold=3 on raw count let 3 consecutive `<tool_call>` openers through, and the parser regex still got confused by 2 nested openers. Pushed `73e574b`:

- GPU block now uses `(opens − closes)` depth instead of raw count, threshold 2. Catches the second consecutive opener with no close in between, before the third can land. Two complete tool calls back-to-back never trip (depth returns to 0).
- `parseToolCalls` strips leading `<tool_call>\s*` repeats before JSON parse. Combined with (1), a single emitted nested opener is fully recovered.

10 Rust tests + 14 Bun tests + 6/6 coherence-gate including tool-call coverage all green.

Kaden-Schutt:
**Reference repro queries cleared on hardware** (gfx1100, ROCm 7.2, qwen3.6:27b.mq4):

```
Prompt 1: "what files are in this directory?"
→ tool_calls: [{ name: "bash", arguments: { command: "ls -la" } }]
  finish_reason: tool_calls

Prompt 2: "write a file here named test.md with the text test inside"
→ tool_calls: [{ name: "write", arguments: { path: "test.md", content: "test" } }]
  finish_reason: tool_calls
```

No attractor loop, no nested openers, no parser repair signal on stderr — the GPU block prevents the second consecutive opener structurally, so the model emits clean spec JSON without the parser needing to strip anything. Both Fluorax-reported failure modes are closed end-to-end.

Issue #74, opened 2026-04-27
API broken


DrBearJew:
Found a likely bug in HipFire’s OpenAI-compatible `/v1/chat/completions` path with `qwen3.6:27b`.

The model generates normally via `hipfire run`, but the OpenAI endpoint returns empty `message.content` while consuming the full generation budget.

Repro:

```bash
curl -sS http://127.0.0.1:11435/v1/chat/completions \
  -H "Content-Type: application/json" \
  -d '{
    "model": "qwen3.6:27b",
    "messages": [
      {
        "role": "user",
        "content": "Think carefully, then write a technical 500-word explanation of neural network quantization."
      }
    ],
    "stream": false,
    "temperature": 0.7,
    "max_tokens": 8192
  }'
```

Result:

```json
{
  "message": { "content": "" },
  "usage": {
    "prompt_tokens": 0,
    "completion_tokens": 8192,
    "total_tokens": 8192
  }
}
```

Per-model config:

```json
{
  "qwen3.6:27b": {
    "thinking": "on",
    "max_think_tokens": 2048,
    "max_tokens": 8192
  }
}
```

Expected behavior: visible final answer in `message.content`, with `max_think_tokens` limiting hidden reasoning before answer generation.

Also, `prompt_tokens: 0` looks incorrect.

For some reason works in Cline/Multi.dev in VS code and Claw Code. But fails in OpenWebui etc.

Kaden-Schutt:
Fixed in fae2867 (commit \`b1dfe30\` on master). Two coupled bugs, both confirmed:

## 1. \`max_think_tokens\` was silently dropped on the API path

Compare:
- \`hipfire run\` path at \`cli/index.ts:1028\`:
  \`\`\`ts
  if (modelCfg.max_think_tokens > 0) genMsg.max_think_tokens = modelCfg.max_think_tokens;
  \`\`\`
- \`/v1/chat/completions\` path at \`cli/index.ts:1308\` (pre-fix): no such line.

[...]

PR #123, opened 2026-05-02
fix(thinking): hard-suppress thinking when thinking=off

fivetide:
## Problem

`thinking=off` was advisory-only — it stripped `<think>...</think>` from
the displayed output but did not prevent the model from generating
thinking tokens. On Qwen3.6-35B-A3B (and other Qwen3 models), this
meant the model could consume the entire `max_tokens` budget inside a
`<think>` block, returning an empty answer.

This is a known model-level issue reported across all major inference
frameworks:
- [QwenLM/Qwen3.6#88](https://github.com/QwenLM/Qwen3.6/issues/88) — repetition loops during thinking
- [QwenLM/Qwen3.6#145](https://github.com/QwenLM/Qwen3.6/issues/145) — infinite loops with recommended sampling params
- [ollama/ollama#14421](https://github.com/ollama/ollama/issues/14421) — qwen3.5:35b looping
- [HuggingFace discussion](https://huggingface.co/Qwen/Qwen3.6-35B-A3B/discussions/20) — endless reasoning loop

## Fix

When `thinking=off` (config) or `enable_thinking=false` (API), set
`max_think_tokens=1` to force the daemon to immediately close the
`<think>` block after 1 token. Same mechanism the OpenAI API's
`enable_thinking=false` already uses.

Applied in both the `run` (single-shot) and `serve` (API) code paths.

**Changed file:** `cli/index.ts` (18 insertions, 2 deletions)

## Proof of impact

99-bottles-of-beer prompt on qwen3.6-35b-a3b, `max_tokens=2048`, `temperature=0`:

| Metric | Baseline (thinking=on, no cap) | With enable_thinking=false |
|--------|--------------------------------|----------------------------|
| completion_tokens | 1545 | **418** (−73%) |
| think_words | 1071 | **0** |
| answer_words | 79 | **279** (+253%) |
| bottle_mentions | 0 | **20** |
| verdict | FAIL: repetitive, no song | **PASS: writes song structure** |

## Health checks

| Gate | Result |
|------|--------|
| `coherence-gate.sh` | 4/4 OK |

## Steps to reproduce

\`\`\`bash
# Start serve
hipfire serve -d

# Baseline (thinking=on, no cap) — model wastes budget on thinking
curl -s http://localhost:11435/v1/chat/completions \
  -H \"Content-Type: application/json\" \
  -d '{"model":"qwen3.6:35b-a3b","messages":[{"role":"user","content":"Write the 99 Bottles of Beer song, starting from 99 down to 0."}],"max_tokens":2048,"temperature":0,"stream":false,"chat_template_kwargs":{"preserve_thinking":true}}' \
  | python3 -c \"import sys,json; d=json.load(sys.stdin); print(d['usage']['completion_tokens'], 'tokens')\"
# → 1545 tokens, 0 bottle mentions

# With fix: enable_thinking=false
curl -s http://localhost:11435/v1/chat/completions \
  -H \"Content-Type: application/json\" \
  -d '{"model":"qwen3.6:35b-a3b","messages":[{"role":"user","content":"Write the 99 Bottles of Beer song, starting from 99 down to 0."}],"max_tokens":2048,"temperature":0,"stream":false,"chat_template_kwargs":{"enable_thinking":false,"preserve_thinking":true}}' \
  | python3 -c \"import sys,json; d=json.load(sys.stdin); print(d['usage']['completion_tokens'], 'tokens')\"
# → 418 tokens, 20 bottle mentions
\`\`\`

### DIFF
--- a/cli/index.ts
+++ b/cli/index.ts
@@ -1139,7 +1139,17 @@ async function run(model: string, prompt: string, image?: string, temp = 0.3, ma
     temperature: temp * TEMP_CORRECTION, max_tokens: maxTokens,
     repeat_penalty: repeatPenalty, top_p: topP,
   };
-  if (modelCfg.max_think_tokens > 0) genMsg.max_think_tokens = modelCfg.max_think_tokens;
+  // thinking=off: hard-suppress by capping thinking to 1 token (model still
+  // emits <think> but is immediately force-closed). This mirrors the
+  // enable_thinking=false semantics from the OpenAI API path.
+  // Previous attempts to inject prose directives with <think>/<no_think>
+  // caused Qwen3.5 to halt at 3-4 tokens — the token-cap approach works
+  // reliably because it operates at the daemon level, not in the prompt.
+  if (modelCfg.thinking === "off") {
+    genMsg.max_think_tokens = 1;
+  } else if (modelCfg.max_think_tokens > 0) {
+    genMsg.max_think_tokens = modelCfg.max_think_tokens;
+  }
   if (image) {
     genMsg.image = resolve(image);
     console.error(`[VL: ${image}]`);
@@ -1521,7 +1531,13 @@ async function serve(port: number) {
         // block, leaving message.content empty after the downstream strip.
         // Reported in #74 with qwen3.6:27b returning empty content + full
         // 8192 completion_tokens despite max_think_tokens=2048 in config.
-        if (effective.max_think_tokens > 0) genParams.max_think_tokens = effective.max_think_tokens;
+        // thinking=off: hard-suppress by capping to 1 token, same as
+        // enable_thinking=false. Overrides any per-model max_think_tokens.
+        if (effective.thinking === "off") {
+          genParams.max_think_tokens = 1;
+        } else if (effective.max_think_tokens > 0) {
+          genParams.max_think_tokens = effective.max_think_tokens;
+        }
         // chat_template_kwargs.enable_thinking=false hard-caps thinking to 1
         // token (model emits <think> then is forced to close). Overrides
         // per-model max_think_tokens because the request semantics are more


Kaden-Schutt (commented):
Read-and-review on Strix Halo (gfx1151). DRAFT, so no merge attempt.

## What this fixes

The diagnosis matches an existing issue surface in this repo: `thinking=off` was a display-only filter, so models like Qwen3.6-35B-A3B could burn the entire `max_tokens` budget inside `<think>...</think>` and return empty `message.content`. The PR's measurements (1545 -> 418 tokens, 0 -> 20 bottle mentions on the 99-bottles prompt) reproduce that pathology and show the fix recovering useful answer tokens.

## Why the approach is correct (and why earlier attempts were not)

This is the same mechanism the OpenAI API path's `enable_thinking=false` already uses: cap thinking at 1 token at the daemon layer rather than injecting a `/no_think` or `<think></think>` directive into the prompt.

There is significant prior session history in this repo against the directive-injection approach. Six commits across about 24 hours (3798399, 2d9c24b, 799c268, cf2a3d8, 68b32ee, b292565) tried various placements of `/no_think` (prose, system message, user prefix, bookended, `/think_off`) and every placement broke a different Qwen3.5 prompt shape with empty `<think><|im_end|>` halts. The current contract is exactly what this PR's author called out:

> "Previous attempts to inject prose directives with `<think>`/`<no_think>` caused Qwen3.5 to halt at 3-4 tokens. The token-cap approach works reliably because it operates at the daemon level, not in the prompt."

So the design space has been searched. The cap-at-daemon path is the known-safe one.

## Code review

`cli/index.ts` two sites, mirrors:
- `run()` (line 1139 area): `if (modelCfg.thinking === "off") genMsg.max_think_tokens = 1`. Overrides any per-model `max_think_tokens` config when thinking is off. Correct precedence: the explicit "off" should win over a numeric cap.
- `serve()` (line 1521 area): same pattern on `effective.thinking`. Comment correctly references #74 (the original bug report).

The branch order in both places is right: check `thinking === "off"` first, fall through to existing `max_think_tokens` numeric path otherwise. Default behavior unchanged when `thinking` is anything else.

## Concerns

1. **Cap value of 1 token.** The PR claims this works because the model emits `<think>` then is force-closed. Worth confirming the daemon's force-close logic handles a 1-token cap cleanly across all the affected models (Qwen3.5/3.6 27B, A3B, plus whatever else has `thinking` defaulting to "off" in the registry). If any model's tokenizer turns `<think>` into more than one token at start, a cap of 1 might truncate mid-marker. The benchmark on Qwen3.6-35B-A3B passes, but a quick check across the registry's `thinking=off` defaults (Qwen3.5-A3B, possibly others) would harden the claim.

2. **No regression test added.** This change has clear measurable behavior (token-count delta, bottle-mention count). A scripted check against the 99-bottles prompt (or similar canary) added to `coherence-gate.sh` would prevent a future change from silently undoing the cap. Soft suggestion, not a blocker.

3. **Empty `<think>` block in output.** With cap=1 the model emits `<think>` then is closed. Worth verifying the post-strip filter handles a partial open tag without trailing close cleanly across models. The fix relies on existing strip logic that the PR author's coherence-gate run validates, but a unit test would be cheap.

## Recommendation

Approach is correct and aligns with the repo's prior search of this design space. The bug is real and the measurement is convincing. Recommend: mark ready-for-review when satisfied, then merge. If you want to widen the validation surface before flipping out of draft, a quick coherence-gate-thinking-off pass across the `thinking=off` default models in the registry would be the next-cheapest step.

Read-only review; ran tests against PR #129 in this same session, did not check this branch out.

PR #124, opened 2026-05-02
fix(dflash): wire max_think_tokens through DFlash spec-decode path

fivetide:
## Problem

The DFlash speculative decoding path explicitly ignored `max_think_tokens`:

```
max_think_tokens=N ignored on DFlash path
(only enforced on AR; set dflash_mode=off to use the AR path with the cap)
```

Users who enabled DFlash (`dflash_mode=on`) had no way to cap thinking
tokens — the model could loop indefinitely inside `<think>` on the
spec-decode path.

This is particularly problematic for Qwen3.6-35B-A3B which has a
[documented thinking-loop issue](https://github.com/QwenLM/Qwen3.6/issues/88)
across all inference frameworks.

## Fix

Wire `max_think_tokens` through `generate_dflash()`:

1. Pass `max_think_tokens` as a parameter (was not passed before)
2. Track `<think>`/`</think>` in decoded text during the token emit loop
   (same logic as the AR path)
3. When the cap is reached, force-emit `</think>\n` and break out of the
   spec-decode batch
4. Remove the old "ignored" warning message

**Limitation:** Unlike the AR path, DFlash cannot splice tokens into the
KV cache mid-spec-cycle, so generation ends after force-close rather
than continuing with answer tokens. This still prevents infinite think
loops from consuming the entire token budget.

**Changed file:** `crates/engine/examples/daemon.rs` (35 insertions, 15 deletions)

## Proof of impact

Direct daemon test with 27b model + DFlash draft, `max_think_tokens=64`:

| Metric | Before (warning) | After (cap enforced) |
|--------|-------------------|----------------------|
| `dflash` field | true | true |
| "ignored on DFlash" warning | **present** | **absent** |
| `<think>` opened | yes | yes |
| `</think>` closed | no (loops) | **yes** (at token 65) |
| tau (acceptance rate) | — | 3.0 |

## Health checks

| Gate | Result |
|------|--------|
| `coherence-gate.sh` | 4/4 OK |

## Steps to reproduce

\`\`\`bash
# Build daemon
cargo build --release --example daemon --features deltanet

# Run daemon directly with DFlash + think cap
echo '{"type":"ping"}
{"type":"load","model":"~/.hipfire/models/qwen3.6-27b.mq4","params":{"max_seq":4096,"kv_mode":"asym2","draft":"~/.hipfire/models/qwen36-27b-dflash-mq4.hfq","dflash_mode":"on"}}
{"type":"generate","id":"test","prompt":"Explain quantum entanglement.","temperature":0.0,"max_tokens":512,"repeat_penalty":1.05,"top_p":0.95,"max_think_tokens":64}' \
  | ./target/release/examples/daemon 2>/dev/null \
  | grep -E '"type":"(info|done)"'

# Before this fix: info message "max_think_tokens=64 ignored on DFlash path"
# After this fix:  no warning, done message with dflash=true and </think> closed
\`\`\`

### DIFF
--- a/crates/engine/examples/daemon.rs
+++ b/crates/engine/examples/daemon.rs
@@ -1628,6 +1628,7 @@ fn generate_dflash(
     prompt: &str,
     system_prompt: Option<&str>,
     max_tokens: usize,
+    max_think_tokens: usize,
     pflash_bypass_reason: Option<&str>,
     pflash_alpha: Option<f32>,
 ) {
@@ -1809,6 +1810,9 @@ fn generate_dflash(
     let mut position = prompt_tokens.len();
     let mut seed_token = first_token;
     let mut stats = SpecStats::new(df.block_size);
+    // max_think_tokens enforcement state (mirrors the AR path).
+    let mut think_count: usize = 0;
+    let mut prev_in_think = false;
     let mut generated = 0usize;
 
     // Post-prefill compaction (FlashCASK pattern from dflash_spec_demo).
@@ -1949,6 +1953,7 @@ fn generate_dflash(
         let committed_tail: Vec<u32> = step.committed.iter().skip(1).copied().collect();
 
         let mut hit_eos = false;
+        let mut think_cap_hit = false;
         for &tok in &committed_tail {
             if generated >= max_tokens { break; }
             emitted.push(tok);
@@ -1964,6 +1969,34 @@ fn generate_dflash(
             }
             generated += 1;
             if tok == target.config.eos_token || im_end_token == Some(tok) || tokenizer.is_terminator(tok) { hit_eos = true; break; }
+
+            // max_think_tokens enforcement (mirrors the AR path). Track
+            // <think>/<⁄think> in decoded text and count tokens inside.
+            if max_think_tokens > 0 {
+                let raw_so_far = tokenizer.decode_bytes(&streamed_tokens);
+                let raw_str = std::str::from_utf8(&raw_so_far).unwrap_or("");
+                let open_idx = raw_str.rfind("<think>");
+                let close_idx = raw_str.rfind("</think>");
+                let in_think = match (open_idx, close_idx) {
+                    (Some(o), Some(c)) => o > c,
+                    (Some(_), None) => true,
+                    _ => false,
+                };
+                if in_think && !prev_in_think { think_count = 0; }
+                if in_think { think_count += 1; }
+                prev_in_think = in_think;
+
+                if in_think && think_count >= max_think_tokens {
+                    // Force-close: emit </think>\n and break out of this batch.
+                    // Unlike the AR path we can't splice into the KV cache mid-
+                    // spec-cycle, so we just stream the close text and break.
+                    // The next request will start fresh.
+                    let _ = writeln!(stdout, r#"{{"type":"token","id":"{}","text":"</think>\n"}}"#, id);
+                    let _ = stdout.flush();
+                    think_cap_hit = true;
+                    break;
+                }
+            }
         }
         position += step.accepted + 1;
         seed_token = step.bonus_token;
@@ -1982,7 +2015,7 @@ fn generate_dflash(
                 }
             }
         }
-        if hit_eos { break; }
+        if hit_eos || think_cap_hit { break; }
     }
 
     // Put target state back on LoadedModel so the next request sees fresh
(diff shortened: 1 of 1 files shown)


Kaden-Schutt (commented):
Read-and-review on Strix Halo (gfx1151). DRAFT, no merge attempt.

## Diagnosis

Bug is real and well-scoped: `generate_dflash()` previously emitted an "ignored" warning when `max_think_tokens > 0` because the spec-decode path had no enforcement. Combined with the per-step batch nature of DFlash (multiple committed tokens per cycle), an unchecked thinking loop on the DFlash path could burn the entire `max_tokens` budget the same way the AR path could.

This PR (and #123) together make the thinking cap work consistently across both inference paths. They should probably land together or with #123 first.

## Approach

Mirror of the AR path's enforcement, with one acknowledged limitation:
- After each emitted token, decode the streamed-token buffer back to UTF-8 and check for `<think>`/`</think>` markers via `rfind`.
- Count tokens emitted while `in_think` is true.
- When the count reaches the cap, force-emit `</think>\n` as a streamed token-JSON and break the spec-cycle loop. Generation ends; no answer tokens are produced after the force-close.

The author is honest about the limitation in the PR body and in the code comment:
> "Unlike the AR path we can't splice into the KV cache mid-spec-cycle, so we just stream the close text and break. The next request will start fresh."

This is the right trade-off. Better to truncate at force-close than to continue emitting from a stale KV state. The user's outcome (no answer) is no worse than the bug it replaces (no answer because budget burned in `<think>`).

## Concerns

1. **O(N^2) decode cost.** `tokenizer.decode_bytes(&streamed_tokens)` is called on every committed-tail token, decoding ALL streamed tokens up to that point. For DFlash with high per-step token rates and long generations, this is the kind of pattern that tends to surface as a perf cliff at long context. The AR path likely has the same shape (this is a mirror), so the cost may already be a known quantity, but worth measuring on a long-think generation before merge. A streaming incremental-decoder window (e.g. last 2 KB of decoded bytes) would bound the work without losing the marker-detection.

2. **String-literal marker matching.** Looks for `<think>` and `</think>` in decoded text. For Qwen3.5/3.6 these are bare ASCII tokens, not special-token IDs, so the match is correct. Edge case: if the model is asked to discuss HTML-like markup and emits the literal text `<think>` in answer tokens (after a real `</think>` close), the heuristic would re-enter the think state. Low probability for the affected model class but worth a comment in the code or a follow-up using token-ID matching where available.

3. **`think_count` reset on `prev_in_think` transition false->true is correct.** Good; if the model alternates `<think>...</think><think>...</think>`, each block gets its own budget. (Though I am not sure Qwen actually does this; the standard pattern is one think block per turn.)

4. **`generated` counter not decremented at force-close.** Stats anomaly only; the SUMMARY emit still uses the actual `generated` count. Not load-bearing.

5. **No regression test.** A scripted check (long-think prompt + asserted `</think>` present in the output, or assert `completion_tokens < max_tokens`) added to `coherence-gate-dflash.sh` would protect this from regressing later. Not a merge blocker but cheap to add.

6. **Coupling with #123.** This relies on `max_think_tokens` being set; #123 wires `thinking=off` to set `max_think_tokens=1` at the CLI/serve layer. If #123 lands first, this PR's path becomes the actual code that handles `thinking=off` on DFlash. Consider noting this dependency on the PR body or sequencing the merges accordingly.

## Recommendation

Approach is sound and the limitation is honestly disclosed. Recommend: address the O(N^2) decode comment (or confirm the AR path has the same shape and accept it for now), then mark ready-for-review. The fix is small, scoped to one function, and matches the AR path's behavior, so the risk surface is bounded.

Read-only review; did not check this branch out or run benchmarks.

Kaden-Schutt:
Rebased your branch onto current master via maintainer-edit access (force-pushed `588405e → 21dda83`). The DFlash-counter logic is unchanged; only the `generate()` dispatch site was reworked to merge cleanly with the `pflash_bypass_reason` / `pflash_alpha` plumbing that landed via #135. The signature is now:

```rust
generate_dflash(m, gpu, stdout, id, prompt, system_prompt,
                max_tokens, max_think_tokens,
                dflash_bypass_reason, dflash_alpha);
```

Verified on gfx1100 (release build, gpu-lock):

| Mode | Output | Tokens |
|---|---|---|
| `dflash_mode=on` + `thinking=off` | "4" @ 32.5 tok/s | 5 (force-closed at cap) |
| `dflash_mode=off` + `thinking=off` (AR regression) | "4" @ 30.1 tok/s | 5 |

Both paths now honor `max_think_tokens=1` (the value set by #123 when `thinking="off"`). Without your fix, the DFlash path would print the "max_think_tokens=N ignored on DFlash path" warning and run unbounded — and on Qwen3.5 with thinking=off, that means an empty visible answer.

Marking ready for admin merge once you confirm the rebased branch represents your intent. Closes the open Codex stop-time finding from yesterday's session ("thinking=off is not enforced on the DFlash path").

PR #125, opened 2026-05-02
fix(daemon): add runtime n-gram loop detector to AR generation

fivetide:
## Problem

Qwen3.6-35B-A3B (and other models) can enter repetition loops in both
the thinking phase AND the answer phase. The `max_think_tokens` cap
only addresses thinking-phase loops. Answer-phase loops like:

> "Wait - writing each line systematically using a pattern.
> Wait - writing each line systematically using a pattern.
> Wait - writing each line systematically using a pattern."

...waste the remaining token budget producing identical content. The
existing `repeat_penalty` (1.05 default, 1.5x cap) is too weak to
break these loops at `temperature=0`.

Known issue across frameworks:
- [QwenLM/Qwen3.6#88](https://github.com/QwenLM/Qwen3.6/issues/88)
- [QwenLM/Qwen3.6#145](https://github.com/QwenLM/Qwen3.6/issues/145)
- [ollama/ollama#14421](https://github.com/ollama/ollama/issues/14421)

## Fix

Add a 4-gram token-level loop detector to the AR generation loop.
During generation, track 4-gram token ID sequences in a sliding window.
When any 4-gram repeats more than the threshold, force EOS.

- **Detection:** HashMap of `[u32; 4]` -> count, checked per token
- **Window:** Last 256 tokens (configurable via `HIPFIRE_NGRAM_WINDOW`)
- **Threshold:** 8 repeats (configurable via `HIPFIRE_NGRAM_LOOP_THRESHOLD`)
- **Scope:** Operates on token IDs, not decoded text — zero decode overhead
- **Disable:** Set `HIPFIRE_NGRAM_LOOP_THRESHOLD=0`

**Changed file:** `crates/engine/examples/daemon.rs` (37 insertions)

## Proof of impact

99-bottles-of-beer prompt on qwen3.6-35b-a3b, `max_tokens=2048`, `temperature=0`:

| Metric | Baseline (no detector) | With n-gram detector |
|--------|------------------------|----------------------|
| completion_tokens | 1545 | **971** (-37%) |
| Early termination | no (hit EOS naturally) | **yes** (loop detected) |

The detector saved 574 tokens of repetitive content that would have
been generated before the model naturally stopped.

Normal (non-looping) generation is unaffected — threshold of 8 is never
reached in coherent output. Verified by coherence gate (0.8b, 4b, 9b
all produce correct answers without false triggers).

## Health checks

| Gate | Result |
|------|--------|
| `coherence-gate.sh` | 4/4 OK (0.8b, 4b, 9b — all coherent, no false triggers) |

## Steps to reproduce

\`\`\`bash
# Build daemon with the fix
cargo build --release --example daemon --features deltanet

# Start serve
hipfire serve -d

# Run 99-bottles prompt
curl -s http://localhost:11435/v1/chat/completions \
  -H \"Content-Type: application/json\" \
  -d '{"model":"qwen3.6:35b-a3b","messages":[{"role":"user","content":"Write the 99 Bottles of Beer song, starting from 99 down to 0."}],"max_tokens":2048,"temperature":0,"stream":false}' \
  | python3 -c \"import sys,json; d=json.load(sys.stdin); print(d['usage']['completion_tokens'], 'tokens')\"
# Baseline: 1545 tokens
# With fix: 971 tokens (loop detected, early stop)

# Verify normal generation unaffected
curl -s http://localhost:11435/v1/chat/completions \
  -H \"Content-Type: application/json\" \
  -d '{"model":"qwen3.5:9b","messages":[{"role":"user","content":"What is the capital of France?"}],"max_tokens":64,"temperature":0,"stream":false}' \
  | python3 -c \"import sys,json; d=json.load(sys.stdin); print(d['choices'][0]['message']['content'])\"
# -> \"Paris.\" (no false trigger)

# Tune thresholds
HIPFIRE_NGRAM_LOOP_THRESHOLD=4 HIPFIRE_NGRAM_WINDOW=128 hipfire serve -d
\`\`\`

### DIFF
--- a/crates/engine/examples/daemon.rs
+++ b/crates/engine/examples/daemon.rs
@@ -2464,6 +2464,17 @@ fn generate(m: &mut LoadedModel, gpu: &mut rdna_compute::Gpu, stdout: &mut std::
         let mut think_count: usize = 0;
         let mut prev_in_think: bool = false;
 
+        // N-gram loop detector: track 4-gram token sequences. When any
+        // 4-gram repeats more than `ngram_loop_threshold` times in the
+        // last `ngram_window` tokens, force EOS. This catches answer-phase
+        // repetition loops that the think cap and repeat penalty miss.
+        // Operates on token IDs (no decode overhead).
+        let ngram_loop_threshold: usize = std::env::var("HIPFIRE_NGRAM_LOOP_THRESHOLD")
+            .ok().and_then(|v| v.parse().ok()).unwrap_or(8);
+        let ngram_window: usize = std::env::var("HIPFIRE_NGRAM_WINDOW")
+            .ok().and_then(|v| v.parse().ok()).unwrap_or(256);
+        let ngram_loop_enabled = ngram_loop_threshold > 0;
+
         // `while` instead of `for 0..max_tokens` so budget-alert injection
         // (which increments `generated` beyond the iteration count) can't
         // push generated past max_tokens: each loop start rechecks the cap.
@@ -2566,6 +2577,32 @@ fn generate(m: &mut LoadedModel, gpu: &mut rdna_compute::Gpu, stdout: &mut std::
                     think_count = 0;
                     prev_in_think = false;
                     if generated >= max_tokens { break; }
+                }
+            }
+
+            // N-gram loop detector: check if any 4-gram in the recent window
+            // repeats excessively. When detected, emit an info message and
+            // force EOS to prevent wasting the remaining token budget on
+            // repetitive output.
+            if ngram_loop_enabled && streamed_tokens.len() >= 4 {
+                let window_start = streamed_tokens.len().saturating_sub(ngram_window);
+                let window = &streamed_tokens[window_start..];
+                if window.len() >= 4 {
+                    let mut ngram_counts = std::collections::HashMap::<[u32; 4], usize>::new();
+                    for w in window.windows(4) {
+                        let key = [w[0], w[1], w[2], w[3]];
+                        *ngram_counts.entry(key).or_insert(0) += 1;
+                    }
+                    let max_count = ngram_counts.values().copied().max().unwrap_or(0);
+                    if max_count >= ngram_loop_threshold {
+                        let _ = writeln!(
+                            stdout,
+                            r#"{{"type":"info","id":"{}","message":"ngram loop detected (4gram repeated {}× in last {} tokens) — forcing EOS"}}"#,
+                            id, max_count, window.len()
+                        );
+                        let _ = stdout.flush();
+                        break;
+                    }
                 }
             }
 


Kaden-Schutt (commented):
Read-and-review on Strix Halo (gfx1151). DRAFT, no merge attempt.

## Diagnosis and approach

This is the third in a series of three thinking/loop guard PRs (#123, #124, #125) from the same author. Together they form defense-in-depth against the documented Qwen3.6-A3B repetition-loop pathology:
- #123: thinking-phase token cap on the AR path (`thinking=off` -> `max_think_tokens=1`)
- #124: thinking-phase token cap on the DFlash path
- #125: answer-phase n-gram loop detector on the AR path

The split is logically clean: the think-cap mechanism handles `<think>...</think>` runaways, but answer-phase runaways like the 99-bottles "Wait - writing each line..." example aren't bounded by it. A token-ID-level 4-gram detector covers the answer-phase case at zero decode cost.

## Code review

The detector is dropped into the AR `while`-loop right before the budget-alert injection block, after the `<think>` tracking. Logic flow:
- `streamed_tokens.len() >= 4` guards under-window startup.
- `window_start = streamed_tokens.len().saturating_sub(ngram_window)` correctly handles the early-generation case where streamed has fewer than `ngram_window` tokens.
- `HashMap::<[u32; 4], usize>` built from `.windows(4)` per iteration. When max count >= threshold, emit info JSON and `break`.

Threshold defaults: 8 repeats over a 256-token window. Both env-tunable (`HIPFIRE_NGRAM_LOOP_THRESHOLD`, `HIPFIRE_NGRAM_WINDOW`); threshold=0 disables. Sane, conservative settings.

Token-ID based (no `tokenizer.decode_bytes` like #124 needed); the per-iteration cost is the HashMap-rebuild on the window slice.

## Concerns

1. **Per-token rebuild is O(window) when O(1) is achievable.** The detector rebuilds the entire HashMap every iteration even though only one 4-gram enters and (after `len > ngram_window`) one leaves. A persistent `HashMap<[u32; 4], usize>` updated incrementally (insert new 4-gram from `streamed_tokens[len-4..]`, decrement the leaving 4-gram from `streamed_tokens[len-window-4..len-window]`) would be O(1) per token. With 256-window default the rebuild is roughly 256 hashmap ops per generated token, so at 50-200 tok/s this is 0.5-2% overhead. Not a blocker, but the incremental version is the same number of lines and avoids the perf cliff at large windows.

2. **Threshold tuning might false-trigger on code prompts.** The PR validates against the coherence battery (cap, code, reason, tool-call) and confirms no false positives on 0.8b/4b/9b. Worth a smoke test against a long code-completion prompt at temp=0 (e.g. write a 200-line Python function) since common syntactic 4-grams like `]:` newline + 4-space indent + `if ` could appear repeatedly in real code without being a "loop". A 4-gram of token IDs is more diverse than a 4-gram of words but still: confirm with one realistic code prompt before merge.

3. **AR-only.** Symmetric concern to #124's claim that DFlash can't splice mid-cycle: DFlash also has answer-phase emit loops where this detector would apply if ported. The author may be assuming DFlash's spec-cycle structure makes answer-phase loops rarer; that's plausible, but a follow-up PR mirroring this on the DFlash path would close the gap. Not blocking.

4. **`max_count` is computed via `.values().copied().max()`** which iterates the full HashMap. For window=256 and unique 4-grams (no loop case), the HashMap has ~253 entries, so max takes 253 iters per detect call. Combined with concern (1), this is an additional ~25% on top of the rebuild cost. Incremental tracking (a `peek_max` on the heap or a single tracked-max variable) avoids both costs.

5. **No regression test.** The 99-bottles measurement is the only quantitative claim. A scripted check (run against a known-looping prompt + assert `completion_tokens < max_tokens`, or assert the info JSON is present in stdout) added to `coherence-gate.sh` would harden against future regression. Cheap.

## Recommendation

Approach is correct and conservative; the env-disable knob (`HIPFIRE_NGRAM_LOOP_THRESHOLD=0`) gives a clean escape hatch. Recommend: address the rebuild-cost comment if cheap (or accept it), validate against a code prompt at temp=0, then mark ready-for-review. The fix complements #123/#124 cleanly and the risk surface is bounded to the AR generation loop.

Read-only review; did not check this branch out or run the scenario.

Kaden-Schutt:
Rebased onto current master via maintainer-edit access (`339333b → 8819a5c`). The detector logic is unchanged; just brought up to current master so the merge is fast-forward.

**Perf-careful merge** — the user flagged this PR as "most likely to cause perf regression so be careful". Ran a controlled before/after on the actual daemon generate() path that the detector touches:

3-fresh-process AR decode bench, qwen3.5:9b, 400-tok prose prompt, gpu-lock acquired, gfx1100:

| Mode | run 1 | run 2 | run 3 | median |
|---|---:|---:|---:|---:|
| `HIPFIRE_NGRAM_LOOP_THRESHOLD=0` (detector OFF) | 117.7 | 117.7 | 117.8 | **117.7** |
| `HIPFIRE_NGRAM_LOOP_THRESHOLD=8` (default, ON) | 117.9 | 117.6 | 117.5 | **117.6** |
| **delta** | | | | **-0.08%** |

Statistically indistinguishable from noise — well inside the 5% speed-gate tolerance. The per-token HashMap rebuild over 256 entries is ~50 µs vs ~8500 µs decode budget at 9B, empirically invisible.

Speed-gate full also clean: 14/14 metrics pass, dflash anchors dead-on (27B-3.5 merge_sort 251.4 vs 250.0 baseline, 9B-3.5 merge_sort 573.1 vs 575.0).

Marking ready for admin merge.

PR #128, opened 2026-05-02
fix(load): pread-based weight loading to prevent OOM on unified-memory APUs

fivetide:
## Summary

- On unified-memory APUs (Strix Halo, 128 GB), mmap + hipMemcpy doubles physical RAM usage (page cache + GPU copy), causing OOM at layer 42/48 when loading the 122B MoE model
- Root cause: `posix_fadvise(FADV_DONTNEED)` is **ignored for mmap'd regions** — Linux won't evict pages while a mapping exists
- `load_weight_tensor()` now uses `pread()` into a reusable buffer + `FADV_DONTNEED` after each upload, keeping page cache bounded to ~5 MB during load
- Non-Unix platforms fall back to the existing mmap path unchanged

## Test plan

- [x] `qwen3.5:122b-a10b` loads all 48 layers on Ryzen AI MAX+ 395 (gfx1151, 125 GB GTT) — previously OOM'd at layer 42
- [x] Model generates correct output at 34.4 tok/s
- [ ] Verify smaller models (9B, 27B) still load correctly on discrete GPUs (regression check)
- [ ] Verify Windows/macOS build (non-unix fallback path)

### DIFF
--- a/crates/engine/src/hfq.rs
+++ b/crates/engine/src/hfq.rs
@@ -10,6 +10,22 @@ use std::collections::HashMap;
 use std::fs::File;
 use std::path::Path;
 
+/// Drop page cache for a file byte range via posix_fadvise(FADV_DONTNEED).
+/// On unified-memory APUs (e.g. Strix Halo), mmap'd model data and
+/// hipMalloc'd GPU copies share physical RAM — without this, loading
+/// a 65 GB model consumes ~130 GB (mmap cache + GPU copy).
+/// Note: madvise(MADV_DONTNEED) does NOT work on MAP_SHARED file-backed
+/// mappings (memmap2 default). posix_fadvise on the fd does.
+#[cfg(unix)]
+fn fadvise_dontneed(fd: std::os::unix::io::RawFd, offset: usize, len: usize) {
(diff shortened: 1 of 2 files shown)


Kaden-Schutt:
Tested on Strix Halo gfx1151 (32 GB system, smaller than the 128 GB target but still reproduces the mechanism):

**A/B comparison loading qwen3.5-9b.mq4 (4.95 GB on disk)**:

| Path | Cached delta during load | Behavior |
|------|--------------------------|----------|
| master (mmap) | -137,228 KB | triggered eviction of existing cache to fit the 5 GB mmap |
| PR #128 (pread + FADV_DONTNEED) | +932 KB | no eviction pressure; cache stayed bounded |

Both paths produced identical generation output (46.7-46.8 tok/s on q8 KV at gen=10).

Mechanism confirmed: pread + FADV_DONTNEED keeps the page cache footprint of model load bounded at sub-MB scale, vs the mmap path which holds the full file in cache and either evicts other pages (small system) or OOMs (the 128 GB Strix Halo + 122B case described in the PR).

LGTM for the unified-memory APU OOM fix.

Issue #136, opened 2026-05-03
RDNA4 native ISA: FP8/INT4 WMMA + SWMMAC + R9700 CU fix on gfx1201


Kaden-Schutt:
## Background

`hipfire` runs on `gfx1201` (Navi 48 / Radeon AI PRO R9700 / RX 9070 XT) today via the existing `gfx12` WMMA path. That path was authored against the RDNA3 ISA shape and routed through `__builtin_amdgcn_wmma_f32_16x16x16_f16_w32_gfx12`. Functionally correct, but it leaves several RDNA4 ISA additions on the table.

This issue tracks moving from "gfx1201 works" to "gfx1201 uses its own ISA."

## Current state on gfx1201 (from master, 2026-05-03)

- 11 kernels in `kernels/src/*_wmma.gfx12.hip` use `__builtin_amdgcn_wmma_f32_16x16x16_f16_w32_gfx12`. K4 unroll applied (issue #65, 2026-04-29). 0.8B pp128 +48%, 4B pp128 +75%, 9B pp128 +84% vs the prior floor.
- 2 gfx1201-specific decode GEMVs (`gemv_hfq4g256.gfx1201.hip`, `gemv_hfq6g256.gfx1201.hip`).
- `dispatch.rs:66` and `:89` route gfx1200/gfx1201 into the gfx12 family.
- `profiler.rs:63-65` recognizes RDNA4 (simds_per_cu=2, max_waves_per_simd=16).

### 1. FP8 / BF8 WMMA not used
RDNA4 adds native \`__builtin_amdgcn_wmma_f32_16x16x16_fp8_fp8_w32_gfx12\` (and BF8 variants). ~2x FP16 throughput on the same matrix unit. For HFQ4: pack codebook entries as FP8 instead of FP16, run same 16x16x16 with ~2x FLOPs.

### 2. SWMMAC (sparse 2:4 WMMA) not used
RDNA4 adds \`__builtin_amdgcn_swmmac_*\` for 2:4 structured sparsity. Another 2x on top of dense WMMA. Out of scope for existing HFQ/MQ codecs (codebooks aren't 2:4-friendly). Track but don't block.

### 3. INT4 WMMA not used
[...]

PR #138, opened 2026-05-04
fix(profiler): correct gfx1200/gfx1201 fallback CU counts (#136 part A)

Kaden-Schutt:
First slice of issue #136 — fix the hardcoded fallback CU counts for RDNA4 in `crates/rdna-compute/src/profiler.rs`.

The primary `read_sysfs_cu_count()` path (KFD `simd_count / 2`) already works correctly. The hardcoded fallback in the unwrap_or_else branch had:

\`\`\`rust
\"gfx1200\" => 32,   // RX 9070
\"gfx1201\" => 32,   // RX 9070 XT
\`\`\`

Both are wrong by ~2x. Real values per AMD spec (and confirmed via KFD on hiptrx):
- gfx1200 (RX 9070): **56 CU**
- gfx1201 (RX 9070 XT, Radeon AI PRO R9700): **64 CU**

KFD on hiptrx (4× R9700) reports `simd_count: 128` per node, sysfs primary path resolves correctly to 64. The fix only matters when sysfs is unavailable (containers without `/sys/class/kfd` mounted, future SKUs not yet covered, etc.), but the fallback should still match real hardware so any occupancy-hint calculation isn't off by half.

## Test plan
- [x] `cargo check -p rdna-compute` clean
- [x] sysfs primary path (`read_sysfs_cu_count`) returns 64 for gfx1201 on hiptrx — unchanged

Hardware: hiptrx (4× Radeon AI PRO R9700, gfx1201, ROCm 7.2.2, Ubuntu 26.04).

### DIFF
--- a/crates/rdna-compute/src/profiler.rs
+++ b/crates/rdna-compute/src/profiler.rs
@@ -88,8 +88,8 @@ impl GpuCapability {
                 "gfx1010" => 40,   // RX 5700 XT
                 "gfx1030" => 60,   // RX 6800
                 "gfx1100" => 48,   // RX 7800 XT
-                "gfx1200" => 32,   // RX 9070
-                "gfx1201" => 32,   // RX 9070 XT
+                "gfx1200" => 56,   // RX 9070
+                "gfx1201" => 64,   // RX 9070 XT / Radeon AI PRO R9700
                 _ => 40,
             }
         });


0x00cl:
I have an RX 9060 XT and it reports `gfx1200` arch, according to [techpowerup the RX 9060](https://www.techpowerup.com/gpu-specs/radeon-rx-9060.c4326) also reports `gfx1200` so it seems wrong that  arch `gfx1200` is for the RX 9070, it reports `gfx1201`. The CU units are correct for both of those cards but not the arch.

gfx1200 (RX 9060): 28 CU
gfx1200 (RX 9060 XT): 32 CU

gfx1201 (RX 9070): 56 CU
gfx1201 (RX 9070 XT, Radeon AI PRO R9700): 64 CU

Not really sure how you'd differentiate between each card when they have the same arch though.

EDIT: I know its a fallback, and I haven't checked the code completely, but how about using `hipGetDeviceProperties` from the `hip_runtime_api.h`, it should return a structure that has one of the following:

```c
  int multiProcessorCount;       ///< Number of multi-processors. When the GPU works in Compute
                                 ///< Unit (CU) mode, this value equals the number of CUs;
                                 ///< when in Workgroup Processor (WGP) mode, this value equels
                                 ///< half of CUs, because a single WGP contains two CUs.
```

Issue #145, opened 2026-05-04
Qwen3.5-122B-A10B MoE: output degenerates after ~200 tokens on MQ4


fivetide:
## Problem

The Qwen3.5-122B-A10B MoE model (`qwen3.5-122b-a10b.mq4`) produces coherent output for ~200 tokens, then degenerates into repeated punctuation/emoji garbage. Dense models (9B, 27B) work perfectly with the same engine.

**Reproducer:**
```bash
curl -s http://127.0.0.1:11435/v1/chat/completions \
  -H "Content-Type: application/json" \
  -d '{"model":"qwen3.5-122b-a10b.mq4","messages":[{"role":"user","content":"how to center a div"}],"max_tokens":500}'
```

Output is coherent for ~200 tokens (CSS tutorial with Flexbox examples), then degenerates:
```
#### Center Horizontally Only (**Traditional**) :**):** )**):** ):** }:**| :** }):**| **:}):»** !** |:»:» !** !) **:** »» **«
```

**Key observations:**
- Deterministic — identical garbage across 3 runs
- Temperature-independent — temp=0 delays visible corruption but logits are wrong
- Short outputs (<200 tokens) work fine (e.g. counting to 50)
- Specific to MoE models, not dense

## Root Cause Analysis

Two compounding sources of accumulated quantization error:

### 1. DeltaNet Q8 state quantization error (~40% of drift)

The 122B MoE model operates at **10x smaller hidden-state magnitudes** than dense models (x_norm ≈ 20 vs ≈ 250 on 9B). This makes Q8 DeltaNet state quantization error proportionally larger. With 64 heads × 128×128 state matrices across 36 DeltaNet layers, the error compounds across tokens.

**Evidence:** FP32 DeltaNet state extends coherent output from ~200 to ~350 tokens. Cost: ~0.2 GB extra (negligible).

### 2. MQ4 expert weight quantization error (~60% of drift)

Each MoE layer applies 8 routed experts + 1 shared expert = 9 contributions to the residual stream. With 48 layers × 256 experts quantized to 4-bit, the accumulated quantization error through 9 expert contributions per layer compounds faster than a dense model's single FFN.

**Evidence:** Even with FP32 state, output degenerates at ~350 tokens into word fragments (`"soci eti es resi gn reali ti s"`).

### Hidden state trajectory

```
Early  (pos 14-50):  x_norm ≈ 14-21, x_max ≈ 2-8   — stable
[...]

PR #147, opened 2026-05-04
feat(loader+quantizer): UMA-safe loader, MQ6 expert quant, disk spill

fivetide:
## Summary
- **HFQ loader**: optional mmap drop for UMA systems — prevents double-buffering (mmap + hipMalloc both consuming system RAM)
- **Quantizer**: `--format mq6` now applies to routed expert weights (was hardcoded MQ4); new `--format mq4-mq6exp` for mixed precision; disk spill for large-model quantization
- **Signature change**: `load_weights` takes `&mut HfqFile` to support mmap drop

### Loader
On unified-memory APUs (Strix Halo etc.), mmap'd model pages and hipMalloc'd GPU copies share physical RAM. A 90 GB model needs ~180 GB without this fix. `drop_mmap()` releases the mapping after header parsing; all tensor reads go through `pread` + `fadvise_dontneed`.

### Quantizer
- Expert weights were hardcoded to MQ4 regardless of `--format`. Now `--format mq6` applies MQ6 to experts too.
- `--format mq4-mq6exp`: MQ4 for attention/shared-expert + MQ6 for routed experts only (~75 GB vs 90 GB full MQ6)
- `TensorSpill`: writes quantized tensor data to a temp file when accumulated memory exceeds threshold, preventing OOM on large MoE models
- `drop_tensor_pages()`: releases source safetensors page cache after each tensor

## Test plan
- [x] Builds clean (warnings only)
- [x] MQ6 full quantization of 122B-A10B produces correct 90 GB output
- [x] Disk spill keeps RSS bounded during quantization (verified: RSS ≤ 36 GB for 90 GB output)
- [ ] Mixed quant (`mq4-mq6exp`) blocked by 234 GB source mmap OOM — needs streaming safetensors reader (future work)
- [ ] MQ6 model loading blocked by 90 GB > 125 GB UMA headroom — needs system with >128 GB RAM to validate quality

Refs #145

### DIFF
--- a/crates/hipfire-arch-llama/src/arch.rs
+++ b/crates/hipfire-arch-llama/src/arch.rs
@@ -58,7 +58,7 @@ impl Architecture for Llama {
     }
 
     fn load_weights(
-        hfq: &HfqFile,
+        hfq: &mut HfqFile,
         cfg: &Self::Config,
         gpu: &mut Gpu,
     ) -> Result<Self::Weights, String> {
--- a/crates/hipfire-arch-qwen35-vl/src/arch.rs
+++ b/crates/hipfire-arch-qwen35-vl/src/arch.rs
@@ -67,7 +67,7 @@ impl Architecture for Qwen35Vl {
     }
 
     fn load_weights(
-        hfq: &HfqFile,
+        hfq: &mut HfqFile,
         cfg: &Self::Config,
         gpu: &mut Gpu,
     ) -> Result<Self::Weights, String> {
--- a/crates/hipfire-arch-qwen35/src/arch.rs
+++ b/crates/hipfire-arch-qwen35/src/arch.rs
@@ -63,7 +63,7 @@ impl Architecture for Qwen35 {
     }
 
     fn load_weights(
-        hfq: &HfqFile,
+        hfq: &mut HfqFile,
         cfg: &Self::Config,
         gpu: &mut Gpu,
     ) -> Result<Self::Weights, String> {
--- a/crates/hipfire-arch-qwen35/src/pflash.rs
+++ b/crates/hipfire-arch-qwen35/src/pflash.rs
@@ -482,7 +482,7 @@ pub fn load_drafter(
     target_tokenizer: &Tokenizer,
     max_kv_seq: usize,
 ) -> HipResult<()> {
-    let hfq = HfqFile::open(path).map_err(|e| hip_bridge::HipError::new(0, &format!(
+    let mut hfq = HfqFile::open(path).map_err(|e| hip_bridge::HipError::new(0, &format!(
         "pflash: open drafter HFQ at {}: {e}", path.display(),
     )))?;
     let drafter_tokenizer = Tokenizer::from_hfq_metadata(&hfq.metadata_json).ok_or_else(||
@@ -506,7 +506,7 @@ pub fn load_drafter(
         if is_hybrid {
             let q35_cfg = qwen35::config_from_hfq(&hfq).ok_or_else(||
                 hip_bridge::HipError::new(0, "pflash: hybrid tensors detected but qwen35 config parse failed"))?;
-            let weights = qwen35::load_weights(&hfq, &q35_cfg, gpu)?;
+            let weights = qwen35::load_weights(&mut hfq, &q35_cfg, gpu)?;
             let scratch = qwen35::Qwen35Scratch::new_with_kv_max(gpu, &q35_cfg, 128, max_kv_seq)?;
             let dn_state = qwen35::DeltaNetState::new(gpu, &q35_cfg)?;
             let kv = KvCache::new_gpu_q8(gpu, q35_cfg.n_layers, q35_cfg.n_kv_heads, q35_cfg.head_dim, max_kv_seq)?;
--- a/crates/hipfire-arch-qwen35/src/qwen35.rs
+++ b/crates/hipfire-arch-qwen35/src/qwen35.rs
@@ -661,8 +661,8 @@ impl DeltaNetState {
 /// Load norm weight for Qwen3.5: stored as offset from 1.0 (output = x * (1 + weight))
 fn load_norm_weight(hfq: &HfqFile, gpu: &mut Gpu, name: &str, shape: &[usize]) -> HipResult<GpuTensor> {
     let full_name = format!("model.language_model.{name}");
-    let (info, data) = hfq.tensor_data(&full_name)
-        .or_else(|| hfq.tensor_data(name))
+    let (info, data) = hfq.tensor_data_vec(&full_name)
+        .or_else(|| hfq.tensor_data_vec(name))
         .unwrap_or_else(|| panic!("tensor not found: {name} or {full_name}"));
 
     let mut f32_data: Vec<f32> = match info.quant_type {
@@ -680,8 +680,8 @@ fn load_norm_weight(hfq: &HfqFile, gpu: &mut Gpu, name: &str, shape: &[usize]) -
 /// mean ~1.6 on Qwen3.5-MoE A3B). Applying +1.0 would over-amplify by ~60%.
 fn load_norm_weight_raw(hfq: &HfqFile, gpu: &mut Gpu, name: &str, shape: &[usize]) -> HipResult<GpuTensor> {
     let full_name = format!("model.language_model.{name}");
-    let (info, data) = hfq.tensor_data(&full_name)
-        .or_else(|| hfq.tensor_data(name))
+    let (info, data) = hfq.tensor_data_vec(&full_name)
+        .or_else(|| hfq.tensor_data_vec(name))
         .unwrap_or_else(|| panic!("tensor not found: {name} or {full_name}"));
     let f32_data: Vec<f32> = match info.quant_type {
         1 => data.chunks_exact(2).map(|c| f16_to_f32(u16::from_le_bytes([c[0], c[1]]))).collect(),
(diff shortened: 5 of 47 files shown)


Kaden-Schutt:
Hey @fivetide — heads up, v0.1.20 just landed on master with the engine modularization refactor: `crates/engine/` split into `crates/hipfire-runtime/` + per-arch crates. Your loader/quantizer changes touch paths that moved (UMA-safe loader → `crates/hipfire-runtime/src/`, MQ6 expert quant → mostly `crates/hipfire-quantize/` which is unchanged + the runtime side).

Tooling to make this mechanical:

```bash
git checkout feat/moe-mq6-quantizer-uma-loader
./scripts/rebase-onto-modular.sh
```

It creates a backup tag, applies the path-rename map + `use engine::*` → `use hipfire_runtime::*` rewrites, and reports any manual semantic fixes. Full migration guide: #155.

Discord for real-time help: <https://discord.gg/F3BaywB8Rs>

Apologies for the timing — this PR is wanted, the refactor was just overdue. Once you rebase, it should slot cleanly into post-modular master.

Kaden-Schutt:
Thanks for the UMA-aware loader work and the MoE expert-targeted MQ6 quant — both are real wedges, particularly for Strix Halo (gfx1151) and large MoE models on RAM-constrained boxes. The `pread + posix_fadvise(DONTNEED)` strategy for dGPU is a sensible read pattern that avoids page-cache pressure during the 122B-class loads.

Four items to address before this can land:

**1. Rebase required onto post-v0.1.20 modular master.** The PR currently touches `crates/engine/src/qwen35.rs` and the old engine paths, which no longer exist — the engine crate was split into `hipfire-runtime` + `hipfire-arch-qwen35` + `hipfire-arch-qwen35-vl` + `hipfire-arch-llama` + `hipfire-arch-toy`. There's a migration helper at `scripts/rebase-onto-modular.sh` with the path-mapping logic — that should make the rebase mostly mechanical.

**2. Coherence gate has not been documented as run.** The new `--format mq4-mq6exp` is a new quant code path that affects MoE expert weight routing — exactly the change class that's caused 122B-A10B attractor failures (see `project_122b_a10b_session_2026_05_04.md`). Per `CLAUDE.md`, this needs `scripts/coherence-gate.sh` to pass with at least one MoE model on the new format. The gate report goes to `bench/coherence-results/` (now gitignored after #160, but still produced).

**3. `Drop for TensorSpill` is missing.** The `cleanup()` method is explicit-call only, so any panic or early-return between `spill()` and `cleanup()` leaks the temp file — at the 122B quantize scale that's potentially 10–90 GB of orphaned data on NVMe, which is the worst time to leak. Implementing `Drop` makes the cleanup automatic on unwinding panics.

**4. `tensor_data()` call-site audit needed.** The `drop_mmap()` change makes `HfqFile.mmap` an `Option<Mmap>`, so any `tensor_data()` (returning `Option<&[u8]>`) call after `drop_mmap()` will silently return `None`. The 36-file `&hfq` → `&mut hfq` sweep is mechanical but it's hard to confirm no call site survived in the less-exercised paths (`speculative.rs`, `pflash.rs`). Either rename `tensor_data()` → `tensor_data_mmap()` and force callers to acknowledge the Option, or add a debug assertion that panics if called after `drop_mmap()`.

**Coordination concern with #153 (weight pager v0.2):** Both PRs target reduced host-side copies on UMA / disk-resident weights. #153 is building a `Transport` trait abstraction (pread / io_uring / dma_buf) that you could plug the UMA-safe path into rather than a parallel mmap-drop side-channel. If #153 lands first, your loader changes will need to integrate with that abstraction; if this lands first, #153 will have to know about `drop_mmap()`. Worth a quick alignment with @kmbandy on which direction to converge before either lands.

Architecture-wise the `--format mq4-mq6exp` naming is consistent with project conventions, and the targeted MoE expert quantization is a real ~15 GB save on 122B. Once rebased + gate-validated + spill-leak-fixed, this becomes mergeable.

(Reviewed via Claude Code on Kaden's behalf.)

fivetide:
Given the complexity of #153 I think the sensitive approach here is to wait for that one to merge first and rebuild this on top. Happy to be a little patient 😊

fivetide:
Rebased onto current master (post-v0.1.20 modularization) and addressed review items:

- **Drop trait for TensorSpill** — temp files now auto-cleaned on panic (no more 10-90 GB leaks)
- **debug_assert in tensor_data()** — catches post-`drop_mmap()` misuse in debug builds
- **Architecture::load_weights signature → `&mut HfqFile`** — enables UMA-safe `drop_mmap()` through trait dispatch; all 30 call sites updated
- **Call-site audit complete** — all callers either use `tensor_data_vec()`/`tensor_data_pread()` after `drop_mmap()`, or hold immutable refs (mmap always present)

**Remaining:** coherence-gate validation with `--format mq4-mq6exp` requires a MoE model on GPU — will run once available.

**Blocking on:** #153 (weight pager v0.2 / Transport trait). Will rebase and integrate once that lands to align with the Transport abstraction.

unverbraucht:
Adjacent finding while auditing the qwen35.rs runtime-dispatch surface for silent-correctness gaps (related to my filing of #179): this PR introduces the `--format mq4-mq6exp` mixed-precision MoE quant. The current PR's specific config — attention=MQ4, experts uniformly MQ6 (both `gate_up` and `down`) — is **safe** under the current dispatcher because it sets `routed_gate_up_mq4 = false`, which falls through to the per-expert `weight_gemv` fallback that's dtype-aware.

But the **`use_kernarg_fused` predicate at `crates/hipfire-arch-qwen35/src/qwen35.rs:1931`** is unsafe for adjacent mixed configurations the PR's framework opens the door to:

```rust
let use_kernarg_fused = k == 8 && routed_gate_up_mq4 && x_rot_local.is_some();
```

This gates on `routed_gate_up_mq4` only — it doesn't check `routed_mq4` (which checks `down`). Configurations like `mq4-mq6down` (MQ4 gate_up + MQ6 down per expert) would set `routed_gate_up_mq4 = true`, satisfy the predicate, then dispatch `gemv_hfq4g256_moe_down_residual_scaled_k8` (line 1953) on MQ6 weights — stride-mismatched read (200 B/group MQ6 read at 136 B/group HFQ4 stride). Silent corruption.

Suggested defensive fix to land alongside `mq4-mq6exp`:

```rust
let use_kernarg_fused = k == 8
    && routed_gate_up_mq4
    && routed_mq4                  // ← also require down to be MQ4
    && x_rot_local.is_some();
```

That makes the predicate match the kernel's actual contract (both gate_up and down are MQ4-stride-shaped). Today's `mq4-mq6exp` continues to take the per-expert fallback (already safe), and any future per-expert mixed quant gets the predicate guard rather than silent corruption.

Not blocking this PR — the current `mq4-mq6exp` config doesn't trigger the gap. But once `mq4-mq6exp` lands and the door is open for further mixed-precision experiments, this is the natural place to harden.

(Filed via Claude Code agent on Kevin Read's behalf — comment reviewed before submit.)

Kaden-Schutt:
## Triage status — rebase needed before review

This PR predates the v0.1.20 modular split. Branch is conflicting (`mergeStateStatus: DIRTY`) against current master.

The substance is valuable:
- HFQ loader optional mmap drop for UMA systems (prevents 90 GB → 180 GB doubling on Strix Halo)
- `--format mq6` applies to routed expert weights
- New `--format mq4-mq6exp` for mixed precision
- Disk spill via `TensorSpill` keeps quantization RSS bounded

But the conflict surface is non-trivial: the signature change to `load_weights(&mut HfqFile, ...)` ripples to 40+ example files, all of which the modular split also reorganized into `hipfire-runtime/examples/`. Mechanical rebase would need to map both the signature change AND the file relocations across crates.

Per the dual-review triage protocol, this PR is in the "do not rebase autonomously" tier (substantive scope, business-logic boundaries). Three paths from here:

1. **Author rebase.** `gh pr checkout 147 && git rebase origin/master`. The `&mut HfqFile` signature change should propagate cleanly to the new `hipfire-runtime/examples/*` paths since the modular split was a rename-only operation for those. The MQ6 expert quant + disk spill in `crates/hipfire-quantize/src/main.rs` should be rebase-stable (that crate didn't move).

2. **Maintainer takeover.** Kaden (or another maintainer) takes the branch, rebases, and ships. Reasonable given UMA-loader is on the roadmap.

3. **Cherry-pick only the quantize-side changes.** The `hipfire-quantize/src/main.rs` MQ6 expert quant + disk spill changes (~155 lines) are independently useful even without the loader changes, and they don't touch the rippled `&mut HfqFile` signature. Could land as a smaller scoped PR while the loader changes wait for proper post-modular rebase.

Which path do you want @fivetide / @Kaden-Schutt?

Test plan items 4 and 5 (`mixed mq4-mq6exp` + MQ6 model loading) are blocked on hardware (234 GB source / >128 GB RAM) per the PR body, so those won't gate any of the above paths.

Kaden-Schutt:
Rebased onto current master (post-#178/#181/#177/#184) and admin-merged as commit `97ee9be` directly to master. Squashed the 4 commits into one with attribution preserved.

Conflict resolution in `crates/hipfire-quantize/src/main.rs`: master had additional flags (`expert_hfq6`, `expert_hfq4`, `q8_router`) that the original PR commit removed; merged version keeps all of them alongside the new `use_mq4_mq6exp` flag. Kept master's q8_router branch above the use_mq4_mq6exp embedding handling.

`cargo check --workspace` clean post-rebase.

Test plan items 4 (mixed mq4-mq6exp) and 5 (MQ6 model loading) remain blocked on hardware availability per the original PR.

cc @fivetide — thank you for the UMA-safe loader work, this unblocks the 90 GB model class on Strix Halo.

PR #159, opened 2026-05-05
Generalize to custom cargo target directories

Janno:
## Summary

I have a shared target directory for all rust projects. hipfire builds just fine but the install script fails because the binaries were not in the expected place. This PR fixes that.

- [ ] `kernels/` (HIP source)
- [ ] examples / daemon
- [x] docs / CI / scripts

## Test plan

- [ ] `cargo build --release --workspace --features deltanet` clean
- [ ] `cargo test --lib --workspace --features deltanet` passes
- [ ] If kernel/dispatch changed: `./scripts/coherence-gate.sh` clean
- [ ] If perf-relevant: `./scripts/speed-gate.sh` within ±2% of locked baselines

## Architecture-trait change?

If this PR changes the `Architecture` trait surface in
`crates/hipfire-runtime/src/arch.rs`, note here. Trait changes ripple
to every arch crate.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -359,19 +359,22 @@ fi
 echo ""
 echo "Installing hipfire..."
 
-if [ -f "$REPO_DIR/target/release/examples/daemon" ]; then
+if ! command -v cargo &>/dev/null; then
+    echo "  Installing Rust..."
+    curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y 2>/dev/null
+    . "$HOME/.cargo/env"
+fi
+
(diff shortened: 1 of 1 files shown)


scripts/install.sh
```diff
@@ -359,19 +359,22 @@ fi
...
 echo "Installing hipfire..."
 
-if [ -f "$REPO_DIR/target/release/examples/daemon" ]; then
+if ! command -v cargo &>/dev/null; then
+    echo "  Installing Rust..."
+    curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y 2>/dev/null
+    . "$HOME/.cargo/env"
+fi
+
+TARGET_DIR=$(cd "$REPO_DIR" && cargo metadata --format-version 1 | grep -oE '"target_directory" *: *"[^"]+"' | cut -d ':' -f 2- | tr -d '"')
```
Janno, line 368: This is not great and could break for various reasons, such as `"` occurring in the path. There might be a better way of parsing json that is guaranteed to be available at this point in the script.

PR #165, opened 2026-05-05
bugfix (cli): fix missing cli chat files installation for unix 

### LINKED ISSUE #163: error: Cannot find module './chat.ts' from '$HOME/.hipfire/cli/index.ts'
Looks like scripts/install.sh is not copying the chat.ts file to the install directory.

To reproduce:
- install with `curl -L https://raw.githubusercontent.com/Kaden-Schutt/hipfire/master/scripts/install.sh | bash`
- run `hipfire chat qwen3.5:9b`

francisduvivier:
## Summary

Simple - quite naive - fix for `hipfire chat` for the case of using the install.sh or install.ps1 script.
Fixes https://github.com/Kaden-Schutt/hipfire/issues/163

The naive part is that this will probably break again in the future if the file list is not kept up to date. But this optionally be done in a follow-up PR. The current PR is just fix, a structural fix would also be a bit of a refactor.

- [ ] `kernels/` (HIP source)
- [ ] examples / daemon
- [x] docs / CI / scripts

## Test plan

- [x] If kernel/dispatch changed: `./scripts/coherence-gate.sh` clean
- [x] If perf-relevant: `./scripts/speed-gate.sh` within ±2% of locked baselines

## Architecture-trait change?
NO

### DIFF
--- a/scripts/install.ps1
+++ b/scripts/install.ps1
@@ -428,6 +428,8 @@ if (-not (Test-Path "$RepoDir\cli\registry.json") -or -not (Test-Path "$RepoDir\
 Copy-Item "$RepoDir\cli\registry.json" "$CliDir\registry.json" -Force
 Copy-Item "$RepoDir\cli\package.json"  "$CliDir\package.json"  -Force
 Copy-Item "$RepoDir\cli\index.ts"      "$CliDir\index.ts"      -Force
+Copy-Item "$RepoDir\cli\chat.ts"       "$CliDir\chat.ts"       -Force
+Copy-Item "$RepoDir\cli\chat_pure.ts"  "$CliDir\chat_pure.ts"  -Force
 
 # Create hipfire.cmd wrapper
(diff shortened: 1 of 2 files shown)


unverbraucht:
Good catch, didn't catch that when implementing chat. I wonder if we can add an AGENT rule to check for new CLI scripts when adding CLI commands here.

francisduvivier:
Yeah so regarding a less naive fix, AGENT rule is an option, but I think it would be even better to set it up in a way that this does not need to be "remembered".

For that, I'd see some options:
- copy over the whole cli dir
  - And then maybe move the test files somewhere else so that they don't get copied over unnecessarily
- Use a bundler that uses index.ts as entry file and then run the bundle instead

If you let me know which route you'd prefer, I can implement that in this or some other PR.

unverbraucht:
@francisduvivier good point - doing something programmatically is always preferable. I'll open up a PR that does the copy, had it in another branch anyway.

PR #166, opened 2026-05-05
feat: multi-GPU pipeline-parallel (pp≥2) for Qwen3.5 — closes #58

### LINKED ISSUE #58: Multi-GPU support roadmap (PP first, TP follow-up)
Tracking issue for multi-GPU support, opened in response to #55 (4× 7900 XTX, mixed 7900 XTX + 6700 XT). This is a discussion/spec sketch, not a commitment to a final shape — pushback welcome before any code.

## What works today

Multi-instance via the HIP runtime's `HIP_VISIBLE_DEVICES` env var:

```bash
HIP_VISIBLE_DEVICES=0 hipfire serve -d --port 11435   # GPU 0
HIP_VISIBLE_DEVICES=1 hipfire serve -d --port 11436   # GPU 1
HIP_VISIBLE_DEVICES=2 hipfire serve -d --port 11437   # GPU 2
HIP_VISIBLE_DEVICES=3 hipfire serve -d --port 11438   # GPU 3
```

Each daemon binds one GPU, loads one model, JIT-compiles arch-appropriate kernels (mixed arch is fine — daemon 0 on a 7900 XTX gets gfx1100 blobs, daemon 1 on a 6700 XT gets gfx1030 blobs). Use your own load balancer / round-robin on top.

What that does NOT solve:

- 70B-class model that doesn't fit in 24 GB on any single card.
- Single-request decode latency cut by N (need TP).
- Reusing memory bandwidth across GPUs for one stream (needs PP or TP).

## Phase 1: Pipeline parallel (PP) — minimum-viable

Layers split into bands across N GPUs. Each token traverses all N sequentially. **Doesn't help latency for one request, but unlocks larger models and shares the per-layer KV cache across the cluster.**

### Sketch

- One `Gpu` instance per physical device. `Gpu::init` learns a `Gpu::init_with_device(id)` variant.
- Engine state holds `Vec<Gpu>` instead of `Gpu`. Layer-to-device assignment is a `Vec<usize>` length `n_layers`.
- Layer band assignment policy:
  - **Uniform**: `device = layer * n_devices / n_layers`. Fine for homogeneous arrays.
  - **VRAM-weighted**: device gets layers proportional to its VRAM. Default for mixed-arch — Nindaleth's 24 GB + 12 GB split would put ~67% of layers on the 7900 XTX, ~33% on the 6700 XT.
- Cross-band hidden-state copy: `hipMemcpyPeerAsync` (or `hipMemcpy` via host pinned if peer is unsupported) between consecutive layers.
- KV cache stays on the device that owns the layer. No cross-device KV traffic.

### Engineering lift

Probably 1–2 weeks of focused work:
- `Gpu` per-device instances (1–2 days)
- `LlamaWeights` / `Qwen35Weights` learn `device: usize` per layer (1 day)
- `forward_pass` adds the inter-band copy (1–2 days)
- Layer assignment policy + CLI surface (1 day)
- Speed-gate baseline for the multi-GPU path on a 7900 XTX × 2 reference rig (1 day; needs hardware)
- Coherence-gate covers all the existing models, plus a 70B that exercises the 24 + 24 = 48 GB headroom (1–2 days)

### Open questions for input

- Should `hipfire serve` learn a `--gpus 0,1,2,3` flag (single daemon, multiple devices), or do we keep one-daemon-per-GPU and add an `orchestrate` mode that routes? The single-daemon route is simpler for users and gives proper PP for big models. The orchestrate route is closer to today's behavior and trivially scales replicas.
- Should the layer assignment honor a runtime hint (`HIPFIRE_GPU_LAYER_PINS=0,0,0,1,1,1,...`) for power-users who want manual control? Probably yes; cheap to add once the infrastructure exists.
- For mixed-arch (7900 XTX + 6700 XT): do we let the slower GPU bottleneck the chain, or is asymmetric layer count enough? VRAM-weighted is the easy answer; performance-weighted is harder (needs per-layer kernel timing per arch).

## Phase 2: Tensor parallel (TP) — single-request latency

Each layer's GEMM is sharded across N GPUs. All-reduce after attention output and after FFN output rebuilds the full residual. **Cuts decode latency by ~N/comm_overhead for one request on a large model.** Much bigger lift than PP.

### What's needed

- RCCL FFI bindings (analogous to `crates/hip-bridge/`). Or a hand-rolled ring all-reduce in HIP — straightforward but reinvents-wheel territory.
- Sharded `load_weights` — each GPU gets `1/N` of each weight tensor along the appropriate axis (column-shard for QKV, row-shard for O / FFN-down).
- `forward_pass` learns to issue the all-reduce after `attn_out` and after `ffn_out`.
- Synchronization barriers — TP forward is bulk-synchronous, much chattier than PP.

### Why later

- RCCL on consumer RDNA is more limited than on data-center RDNA — peer-to-peer over PCIe without NVLink-class fabric will bottleneck on small batch sizes (which is exactly the inference case).
- The RDNA arch-coverage push (RDNA1→4, APU, CDNA) currently ranks higher than depth on a single rig. TP is the right thing eventually but probably v0.2.x territory, not v0.1.x.
- A clean PP implementation gives us most of the wins (large models running) and 60% of the engineering scaffolding TP would later need.

## Explicitly out of scope (for now)

- **Speculative decode across GPUs**. DFlash already runs target + draft on one device; splitting them across two adds inter-device sync to the verify cycle and the math gets ugly fast.
- **Different models on different GPUs with auto-routing**. That's "multi-instance + orchestrator," already works via separate daemons + your-own-load-balancer; if there's demand for an in-tree orchestrator, file a separate request.
- **Heterogeneous TP** (mixed-arch in one TP group). Shard balancing across different per-GPU GEMM throughputs is non-trivial; PP handles mixed-arch much more gracefully.

## References

- #55 — original request (4× 7900 XTX, mixed 7900 XTX + 6700 XT)
- `crates/rdna-compute/src/dispatch.rs::Gpu::init` — current single-device init path
- `crates/hip-bridge/src/ffi.rs:241` — `hipSetDevice` already wired through the FFI

alpineQ:
## Summary

Pipeline-parallel inference over `n_devices` homogeneous AMD GPUs for the
Qwen3.5 family (dense + 3.5/3.6 A3B MoE). Layers shard into contiguous
bands; the residual stream `s.x` flows through bands sequentially with
one peer-copy per boundary. `output_norm + lm_head` execute on dev_last
(Variant 2 placement, matches Megatron / DeepSpeed / vLLM).

Default `pp = 1` is byte-identical to pre-PR behavior. Multi-GPU is
opt-in via the `pp` field in the daemon load message.

## What ships

- 11 commits, +5798 / -70 LOC, additive-only on master code paths.
- Tracking issue: #58.
- See `docs/multi-gpu.md` for memory budget, deployment recipes, env
  vars, refusal matrix, throughput baselines, peer-access verification.
- See `docs/plans/multi-gpu-pp.md` for stage-by-stage architecture
  rationale + open questions.

### Stages

| # | Scope |
|---|-------|
| 1 | `hip-bridge` peer-access FFI (`hipDeviceCanAccessPeer`, `hipMemcpyPeer{,Async}`, `hipPointerGetAttributes` + 704→Ok translation) |
| 2 | `rdna-compute` `Gpu::init_with_device(id)` + `bind_thread()` invariant on every public `Gpu::*` (286 fns audited via `scripts/verify-bind-thread.sh`) + `HIPFIRE_DETERMINISTIC` opt-out |
| 3 | `hipfire-runtime::multi_gpu` orchestrator (`Gpus`, `init_uniform`, `init_layers`, `enable_peer_all`, `boundary_copy`, `wait_boundary`) + 14 `KvCache::new_gpu*_multi` constructors |
| 4 | `hipfire-arch-qwen35` `load_weights_multi` + per-device `Qwen35ScratchSet` + `DeltaNetState::new_with_quant_multi` (returns `la_to_device` map for daemon reset routing) |
| 5 | `forward_scratch_multi` + `forward_scratch_layers_multi` (decode). All 4 layer-type branches: DeltaNet, FullAttn, DeltaNetMoe, FullAttnMoe |
| 6 | `forward_prefill_batch_multi` + `PrefillBandCtx`. The only invasive existing-fn modification: `forward_prefill_chunk` gains `band: Option<&PrefillBandCtx<'_>>` (`None` preserves byte-exact pp=1 behavior) |
| 7 | `daemon.rs` `LoadedModel.pp` + `generate_multi` (full feature port: EosFilter, LoopGuard, PromptFrame, repeat_penalty, top_p, attractor block, max_think_tokens) + refusal contracts |
| 8 | Examples (`pp_parity`, `pp_parity_chatml`, `pp2_vram_probe`, `gpus_smoke`, `peer_smoke`, `dual_gpu_smoke`) + `pp-gate.sh` 4-barrier battery + pre-commit hotspot regex |
| 9 | Docs (`docs/multi-gpu.md`, `docs/plans/multi-gpu-pp.md`) + `tests/speed-baselines/gfx1100x2_pp.txt` |

## Refusal matrix at load (`pp > 1`)

| Feature | Behavior |
|---------|----------|
| `arch_id ∈ {5, 6}` (Qwen3.5 dense + MoE/A3B) | Accepted |
| `arch_id` other (LLaMA / plain Qwen3) | Refused — single-GPU only in v1 |
| VL models | Refused — v1.1 |
| DFlash draft | Refused — v1.1 (eviction context single-device) |
| CASK / TriAttention sidecar | Refused — v1.1 |

## Validation

Reference hardware: 2× Radeon RX 7900 XTX (gfx1100), Ryzen 9 9950X,
NixOS, ROCm 6.4.3, peer access bidirectional via PCIe Gen4.

- `cargo build --workspace --release --all-targets` — clean.
- `cargo test --workspace --features deltanet` — 156 passed, 0 failed,
  4 hardware-gated `#[ignore]`d.
- `./scripts/verify-bind-thread.sh` — 286 pub fn audited, all pass.
- **`./scripts/pp-gate.sh` — 4/4 PASS:**
  1. `pp_parity_chatml` — per-token `forward_scratch_multi` ≡
     `forward_scratch` bit-exact, 50 decode tokens.
  2. Daemon `pp=1` ≡ `pp=2` byte-identical with `HIPFIRE_DETERMINISTIC=1`
     (sha256 `63f6060121d88269` on greedy ChatML completion).
  3. DFlash + `pp=2` refused at load.
  4. CASK + `pp=2` refused at load.
- `coherence-gate.sh` — no hard errors on 6 prompts (0.8B / 4B / 9B / 27B).
- Cross-process probe `50cee6c → HEAD` (9B mq4 decode):
  127.9 → 128.4 tok/s (+0.4 %, in noise). Branch is perf-neutral on the
  AR forward path.

### Methodology note

`scripts/pflash-baselines/gfx1100-2026-05-02.json` was captured on
maintainer hardware. On the contributor's gfx1100 it shows 5/12 drift
of 11–20 % on `*_baseline` 27B mq3 ≥16k metrics — both at the branch
base (`50cee6c`) and at HEAD. Same pattern at upstream/master
unchanged, so the drift is foreign-baseline noise, not a regression.
Verified by cross-process probe above.

## Throughput baseline (gfx1100 × 2, ROCm 6.4.3)

`HIPFIRE_DETERMINISTIC=1` → bit-equivalent pp=1 ↔ pp=2.

| Model | Prompt | pp=1 prefill | pp=2 prefill | pp=1 decode | pp=2 decode | pp=2/pp=1 decode |
|-------|--------|--------------|--------------|-------------|-------------|------------------|
| 0.8B mq4 | 22 tok | 838 tok/s | 588 tok/s | 332 tok/s | 227 tok/s | 68 % |
| 0.8B mq4 | 322 tok (chunked) | 6493 tok/s | 5490 tok/s | 315 tok/s | 212 tok/s | 67 % |
| 35B-A3B mq4 (MoE) | 15 tok | 331 tok/s | 258 tok/s | 142 tok/s | 97 tok/s | 68 % |

The pp=2 decode penalty is inherent to v1: per-token
`forward_scratch_multi` pays one HIP launch per kernel per layer with
no graph capture. Pipelined decode + per-band graph capture deferred
to v1.1 (see `docs/plans/multi-gpu-pp.md` §4.2).

## Architectural caveat

`Architecture` trait (`hipfire-runtime::arch`) signs on `&mut Gpu`
(single-GPU). Bring-up for `pp > 1` bypasses the trait and calls
`qwen35::load_weights_multi` directly from the daemon. Forward is
intentionally off-trait per `arch.rs:26-42`, so `forward_*_multi`
fits the existing pattern.

The `*_multi` variants are additive (mechanical clones with per-layer
device routing) rather than unified through `Option<&Gpus>`. This was
deliberate — the unified form requires a trait shape change that
breaks `hipfire-arch-llama`, `hipfire-arch-qwen35-vl`, and
`hipfire-arch-toy`, plus 80+ `forward_scratch` and 144+ `KvCache::new_gpu*`
call-site updates. Additive minimizes review surface and guarantees
pp=1 byte-equivalence by construction.

If preferred, a follow-up PR can: (a) unify `KvCache` constructors via
device-resolver closure, (b) refactor `load_weights` body through
extracted `load_*_into` helpers (interface unchanged), (c) extend the
`Architecture` trait with a defaulted `load_weights_multi` /
`new_state_multi`. `forward_scratch` unification is harder to justify
(hot path, byte-equivalence regression-prone) and probably stays
parallel.

## Out of scope (v1)

Refused at load with clear error:
- DFlash + pp>1 — eviction-context single-device, v1.1.
- CASK + pp>1 — same, v1.1.
- VL + pp>1 — vision encoder single-device, v1.1.
- Llama + pp>1, plain Qwen3 + pp>1 — single-GPU only in v1.

Architectural follow-ups:
- Pipelined decode + per-band graph capture (`forward_scratch_multi` is
  per-token sequential currently).
- Pipelined prefill (chunk N+1 on dev_0 while chunk N runs on dev_1).
- VRAM-weighted automatic split (`init_vram_weighted` is stubbed; manual
  escape hatch is `init_layers(per_device: &[usize])`).

## Open questions for maintainer

(Restated from #58 for permanence.)

1. **DFlash + PP scope** — v1 refuses. Is that the long-term call, or
   should DFlash + pp ship together in v1.1?
2. **Mixed-arch policy** — v1 hard-fails `gfx*` mismatch. Soft-warn vs
   hard-fail preference?
3. **`HIPFIRE_DEVICES` API surface** — currently logical IDs
   post-`HIP_VISIBLE_DEVICES`. Alternative: re-use
   `HIP_VISIBLE_DEVICES` semantics directly?
4. **CHANGELOG** — left unchanged in this PR. Maintainer adds the
   version + entry on merge.

## Test plan

- [ ] CI build clean
- [ ] `./scripts/pp-gate.sh` all 4 barriers PASS on 2-GPU host
- [ ] `./scripts/coherence-gate.sh` text-prompts pass
- [ ] Single-GPU regression: load any pp=1 model, generate — byte-identical
      to pre-PR behavior (verified locally; CI re-verification welcome)
- [ ] Refusal contracts: `pp=2 + draft=…` errors at load with
      `"DFlash requires single-GPU in v1; see issue #58 v1.1 roadmap"`
- [ ] On 1-GPU CI, `pp-gate.sh` exits 0 with skip message

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -50,11 +50,13 @@ CHANGED=$(git diff --cached --name-only --diff-filter=ACM)
 # - The 4 PR1-4 modules (loop_guard, sampler, prompt_frame, eos_filter)
 #   are all hot-path; touching any of them affects every model load.
 # - `gemma4\.rs` covered for when the gemma branch forward-ports.
-HOTSPOT='\.hip$|gemv|forward|quant|rotate|dispatch|fwht|rmsnorm|fused|mq4|hfq|magnum|gemm|wmma|llama\.rs|qwen35\.rs|qwen35_vl\.rs|gemma4\.rs|sampler\.rs|loop_guard\.rs|prompt_frame\.rs|eos_filter\.rs|arch\.rs'
+# multi_gpu\.rs is listed here so that touching the PP dispatch path
+# triggers the coherence battery (same contract as single-GPU forward).
+HOTSPOT='\.hip$|gemv|forward|quant|rotate|dispatch|fwht|rmsnorm|fused|mq4|hfq|magnum|gemm|wmma|llama\.rs|qwen35\.rs|qwen35_vl\.rs|gemma4\.rs|sampler\.rs|loop_guard\.rs|prompt_frame\.rs|eos_filter\.rs|arch\.rs|multi_gpu\.rs'
 
-# Multi-GPU PP hotspots — different bucket than the single-GPU forward
-# bucket above. pp-gate.sh runs only when one of these is touched AND
-# 2+ GPUs are visible; otherwise it skips (CI / single-GPU dev box).
+# Multi-GPU PP hotspots — superset of the HOTSPOT entry above. pp-gate.sh
+# runs only when one of these is touched AND 2+ GPUs are visible; otherwise
+# it skips (CI / single-GPU dev box).
 PP_HOTSPOT='multi_gpu\.rs|peer_access|pp_|pipeline|stages|forward_prefill_batch_multi|forward_scratch_multi|Gpus|init_uniform|init_layers|boundary_copy'
 
 PP_TOUCHED=0
@@ -62,7 +64,7 @@ if echo "$CHANGED" | grep -qE "$PP_HOTSPOT"; then
     PP_TOUCHED=1
 fi
 
-if ! echo "$CHANGED" | grep -qE "$HOTSPOT" && [ "$PP_TOUCHED" -eq 0 ]; then
+if ! echo "$CHANGED" | grep -qE "$HOTSPOT"; then
     exit 0
 fi
 
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -72,6 +72,26 @@ else
     PFLASH_GATE_NOTE="(skipped — no pflash path in diff; force with HIPFIRE_FORCE_PFLASH_GATE=1)"
 fi
 
+# ── bind_thread invariant on dispatch.rs ────────────────────────────────
+# Multi-GPU plan (issue #58) requires every pub fn in `impl Gpu` to either
+# call `self.bind_thread()` first, or carry a `// bind_thread: skip` marker.
+# Silent mis-bind = malloc-on-wrong-device → cross-device pointer corruption.
+if echo "$CHANGED" | grep -qE 'crates/rdna-compute/src/dispatch\.rs'; then
+    if ! ./scripts/verify-bind-thread.sh; then
+        echo
+        echo "========================================================================"
+        echo "COMMIT BLOCKED: dispatch.rs has pub fn missing bind_thread()."
+        echo "========================================================================"
+        echo
+        echo 'Add `self.bind_thread()?;` (or `self.bind_thread_or_warn();` for'
+        echo 'non-HipResult fn) as the first statement, OR add a'
+        echo '`// bind_thread: skip — <reason>` comment for pure-state queries.'
+        echo
+        exit 1
+    fi
+    echo
+fi
+
 # ── Coherence battery ────────────────────────────────────────────────────
 echo "=== Coherence Battery (short) ==="
 echo "Engine code changed. Running coherence battery (~2-4 min)..."
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -52,7 +52,17 @@ CHANGED=$(git diff --cached --name-only --diff-filter=ACM)
 # - `gemma4\.rs` covered for when the gemma branch forward-ports.
 HOTSPOT='\.hip$|gemv|forward|quant|rotate|dispatch|fwht|rmsnorm|fused|mq4|hfq|magnum|gemm|wmma|llama\.rs|qwen35\.rs|qwen35_vl\.rs|gemma4\.rs|sampler\.rs|loop_guard\.rs|prompt_frame\.rs|eos_filter\.rs|arch\.rs'
 
-if ! echo "$CHANGED" | grep -qE "$HOTSPOT"; then
+# Multi-GPU PP hotspots — different bucket than the single-GPU forward
+# bucket above. pp-gate.sh runs only when one of these is touched AND
+# 2+ GPUs are visible; otherwise it skips (CI / single-GPU dev box).
+PP_HOTSPOT='multi_gpu\.rs|peer_access|pp_|pipeline|stages|forward_prefill_batch_multi|forward_scratch_multi|Gpus|init_uniform|init_layers|boundary_copy'
+
+PP_TOUCHED=0
+if echo "$CHANGED" | grep -qE "$PP_HOTSPOT"; then
+    PP_TOUCHED=1
+fi
+
+if ! echo "$CHANGED" | grep -qE "$HOTSPOT" && [ "$PP_TOUCHED" -eq 0 ]; then
     exit 0
 fi
 
@@ -171,6 +181,31 @@ if [ "${HIPFIRE_FORCE_SPEC_GATE:-0}" = "1" ] \
 fi
 
 echo
+
+# ── Multi-GPU PP gate ────────────────────────────────────────────────────
+# Skips silently when fewer than 2 GPU are visible — single-GPU CI and
+# dev boxes don't block on multi-GPU regressions. When 2+ GPU AND a
+# multi_gpu/pp_/peer_access hotspot is touched, run the parity battery.
+if [ "$PP_TOUCHED" -eq 1 ]; then
+    echo "=== Multi-GPU PP Gate ==="
+    echo "Multi-GPU code changed. Running pp-gate (skips if <2 GPU)..."
+    echo
+    if ! ./scripts/pp-gate.sh; then
+        echo
+        echo "========================================================================"
+        echo "COMMIT BLOCKED: pp-gate.sh hit a HARD failure."
+        echo "========================================================================"
+        echo
+        echo "Multi-GPU pp=2 ≢ pp=1 byte-identical, refusal contract broken,"
+        echo "or daemon panic. Investigation:"
+        echo "  ./scripts/pp-gate.sh                       # full battery"
+        echo "  ./scripts/pp-gate.sh --skip-end-to-end     # parity-only"
+        echo
+        echo "Refer to docs/multi-gpu.md for the validation contract."
+        exit 1
+    fi
+    echo
+fi
 
 # ── Speed gate ────────────────────────────────────────────────────────────
 echo "=== MQ4 Speed Gate (fast) ==="
--- a/README.md
+++ b/README.md
@@ -74,6 +74,7 @@ and pointers at where the fat is. Cached snapshot at
 | [BENCHMARKS.md](docs/BENCHMARKS.md) | Measured perf per arch, vs ollama |
 | [ARCHITECTURE.md](docs/ARCHITECTURE.md) | Engine layout, dispatch, two model paths |
 | [QUANTIZATION.md](docs/QUANTIZATION.md) | MQ4 / HF4 design, asym KV cache, FWHT math |
+| [multi-gpu.md](docs/multi-gpu.md) | Pipeline-parallel (pp≥2) — memory budget, deployment, refusals |
 | [methodology/perf-benchmarking.md](docs/methodology/perf-benchmarking.md) | Bench protocol — read before claiming a perf win |
 
 ## License
--- a/crates/hip-bridge/examples/peer_smoke.rs
+++ b/crates/hip-bridge/examples/peer_smoke.rs
@@ -1,0 +1,133 @@
+//! Peer-access smoke: bidirectional enable_peer_access, hipMemcpyPeer
+//! byte-equality round-trip, pointer_get_attributes device verification.
+//!
+//! Run: HIP_VISIBLE_DEVICES=0,1 cargo run -p hip-bridge --example peer_smoke
+
+use hip_bridge::{HipRuntime, MemoryType};
+
+const SIZE: usize = 1 << 20; // 1 MiB
+
+fn main() {
+    println!("Loading HIP runtime via dlopen...");
+    let hip = HipRuntime::load().expect("failed to load HIP runtime");
+
+    let count = hip.device_count().expect("failed to get device count");
+    println!("Visible devices: {count}");
+    assert!(count >= 2, "peer_smoke requires ≥2 devices (got {count}). Set HIP_VISIBLE_DEVICES=0,1");
+
+    for id in 0..count {
+        hip.set_device(id).expect("set_device");
+        let arch = hip.get_arch(id).unwrap_or_else(|_| "unknown".to_string());
+        let cur = hip.current_device().expect("current_device");
+        let (free, total) = hip.get_vram_info().unwrap_or((0, 0));
+        println!(
+            "  dev {id}: arch={arch} current_device={cur} vram={:.1}/{:.1} GiB",
+            free as f64 / 1e9,
+            total as f64 / 1e9,
+        );
+        assert_eq!(cur, id, "current_device disagrees with set_device");
+    }
+
+    // ── Probe peer access ────────────────────────────────────────
+    println!("\nProbing peer access (0↔1):");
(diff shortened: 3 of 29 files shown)


Kaden-Schutt:
Big thanks for taking on multi-GPU PP — this is the biggest external addition since the v0.1.20 modular split, and the architecture is sound: per-layer slicing, KV stays on the owning device, single-GPU path untouched, DFlash/CASK correctly rejected at load time when `pp > 1`. Modular topology compliance is clean (no circular deps, `Architecture` trait not extended). Megatron Variant 2 placement of `output_norm + lm_head` on the last device is the right call.

Three blockers before merge, one soft concern:

**1. `enable_peer_all` doesn't handle `HIP_ERROR_PEER_ACCESS_ALREADY_ENABLED` (704)** — daemon restart on the same process group will hard-fail because ROCm retains peer-access state across some driver versions. Pattern matches the existing `217` (UNSUPPORTED) arm; one-line fix:
```rust
Err(e) if e.code == HIP_ERROR_PEER_ACCESS_ALREADY_ENABLED => {
    // already enabled from a prior process — treat as success
}
```
The `hip-bridge` FFI bindings already declare the constant, which is a strong hint this case was anticipated.

**2. `pp-gate.sh` doesn't invoke `coherence-gate.sh`.** Per `CLAUDE.md`: "Any change to kernels, quant formats, dispatch, fusion, rotation, rmsnorm, or the forward pass MUST pass `./scripts/coherence-gate.sh` before committing." `forward_scratch_multi` and `forward_prefill_batch_multi` are new forward-pass dispatch paths — the parity check + refusal-contract smoke `pp-gate.sh` runs is necessary but not sufficient. Either add an explicit `coherence-gate.sh` invocation at the bottom of `pp-gate.sh` when `n_gpus >= 2`, or add `pp-gate.sh` paths to the coherence hotspot in `.githooks/pre-commit`. (Heads up: I just landed #168 which slimmed the pre-commit hooks — easiest path is probably to extend the HOTSPOT to include the multi-GPU paths so the existing coherence-gate fires automatically.)

**3. `pp_parity_chatml` doesn't enforce `HIPFIRE_DETERMINISTIC`.** FP16 reduction order across devices is legitimately non-deterministic in PP, so without setting deterministic mode at the top of the harness, the parity check can flap on thermal/DPM state — the same class of failure documented in `docs/methodology/perf-benchmarking.md`. Either set `std::env::set_var("HIPFIRE_DETERMINISTIC", "1")` at the top of the parity example, or assert it's set as a precondition.

**Soft concern (non-blocking but worth documenting):** `boundary_copy` with `active_stream = None` records into the HIP null stream. Same ordering-fragility class as the `memset_async` silent-fallthrough bug we fixed in `da2753e` — see `feedback_memset_async_gated_falls_through.md`. The sync `memcpy_peer` completes on the host before return, so data is present, but the `event_record(&event, None)` is fragile across ROCm versions. Cleanest fix: skip `event_record` entirely in the `None` arm and call `event_synchronize` inline; or follow the `memset_async` precedent and require `active_stream` for multi-GPU mode.

Once those land, this is mergeable. Excellent groundwork — the per-layer DeltaNet boundary-copy strategy is the right call, and the docs honesty about pp=2 decode being ~68% of pp=1 (capacity scaling, not throughput scaling) avoids the classic PP misframing.

(Reviewed via Claude Code on Kaden's behalf.)

alpineQ:
Thanks for the thorough review. All three blockers addressed in `e12631a`, rebased onto `03c8850` (picks up #168).

**Blocker 1 — `enable_peer_all` + error 704**

Added an explicit match arm for `HIP_ERROR_PEER_ACCESS_ALREADY_ENABLED` in `enable_peer_all`. `ffi.rs` already converts 704→`Ok(())` at the FFI layer (belt-and-suspenders), but the arm makes the intent explicit and guards against ROCm versions that surface 704 through a different path.

**Blocker 2 — coherence gate on PP paths**

Added `multi_gpu.rs` to `HOTSPOT` in `.githooks/pre-commit`. Touching `multi_gpu.rs` now fires the coherence battery unconditionally, then the PP gate. Also merged my Stage 8 hook changes cleanly with #168's pflash path-gating.

**Blocker 3 — `HIPFIRE_DETERMINISTIC` in `pp_parity_chatml`**

Added `std::env::set_var("HIPFIRE_DETERMINISTIC", "1")` at the top of `main()`. `pp-gate.sh` already exports the flag, but standalone runs were unguarded.

**Soft concern — `boundary_copy` null-stream `event_record`**

In the `active_stream = None` arm: dropped `event_record(&event, None)`, destroy the event immediately, return `BoundaryEvent { completion: None }`. `wait_boundary` early-returns on `completion: None` without touching HIP. Sync `memcpy_peer` already serializes on the host — no event tracking needed.

Kaden-Schutt:
PR #166 pp-gate validated end-to-end on hiptrx (4× gfx1201) — all four barriers PASS:

- `pp_parity_chatml` per-token forward bit-equivalence (50 decode tokens) ✅
- daemon pp=1 vs pp=2 byte-identical greedy under `HIPFIRE_DETERMINISTIC=1` (md5 `1cb719a42ec90514` matches) ✅
- DFlash + pp=2 refusal at load ✅
- CASK + pp=2 refusal at load ✅

Bit-identical pp=1 vs pp=2 is the strongest signal possible — every layer transition, peer copy, KV-write, and sampler invocation produced the same token stream on 4 GPUs as on 1.

Ready to merge but **needs a rebase** — PR #158 (gfx906 MMQ redesign) landed at `afb84bd` after your rebase point and conflicts on:

- `README.md` (you added multi-GPU section; #158 added a gfx906 perf table)
- `crates/rdna-compute/src/dispatch.rs` (both touch dispatch — likely small)
- possibly `.githooks/pre-commit` (you added `multi_gpu.rs` to HOTSPOT; #168 had also restructured the hook — should resolve cleanly since you already rebased onto #168)

Once you push the rebase, will admin-merge directly. Thanks for the thorough blocker fixes — the architecture is solid.

Kaden-Schutt:
Superseded by #174 — admin-rebased onto current master (post-#158) and merged at `262e5f6`. All 12 of your commits preserved with original authorship; only one needed conflict resolution (`feat(rdna-compute): Gpu::init_with_device + bind_thread audit + DETERMINISTIC` — bind_thread()? placed before the gfx906 dp4a early-return at every site so multi-GPU dispatch is honored on the fast path). Excellent work — the architecture was solid, blocker fixes were precise, pp-gate cleared 4-barrier on hiptrx (4× gfx1201) before rebase. Closes #58.

Issue #179, opened 2026-05-06
MQ3 missing from MoE-batched matchers in qwen35.rs (silent-correctness gap on hypothetical A3B+MQ3 model)


unverbraucht:
## Summary

In `crates/hipfire-arch-qwen35/src/qwen35.rs`, the `is_mq` matchers at **line 4651** (MoE-batched LinearAttention body) and **line 4802** (MoE-batched FullAttention body) drop `MQ3G256` from the rotated-quant set, while every other equivalent matcher in the file includes it. An MoE model with MQ3 weights would silently bypass the FWHT rotation pre-pass on those two paths and produce corrupted prefill/verify state.

## Reproducer / current state

```rust
// crates/hipfire-arch-qwen35/src/qwen35.rs:4651  (MoE-batched LA body)
let is_mq = matches!(layer.wqkv.gpu_dtype, DType::MQ4G256 | DType::MQ6G256);
//                                                                       ^^^^^^^^^^^^^^^^ missing MQ3G256

// crates/hipfire-arch-qwen35/src/qwen35.rs:4802  (MoE-batched FA body)
let qkv_is_mq = matches!(layer.wq.gpu_dtype, DType::MQ4G256 | DType::MQ6G256);
//                                                                            ^^^^^^^^^^^^^^^^ missing MQ3G256
```

Compare with the equivalent matchers in the **dense** LA/FA bodies — all 26 of them include MQ3G256, e.g.:

```rust
// line 3946  (dense LA, prefill batched, wqkv)
let is_mq = matches!(layer.wqkv.gpu_dtype, DType::MQ4G256 | DType::MQ6G256 | DType::MQ3G256);
//                                                                         ^^^^^^^^^^^^^^^^^^ MQ3 included
```

`git blame` points at commit 10f0c0eb (2026-04-17, *engine-modularization era*). The MoE-batched body comment (line 4649-4650) explicitly says *"Duplicated inline for now — can be factored into a `prefill_la_body_batched` helper when dense and MoE LA paths are proven byte-exact."* — the dropped `MQ3G256` looks like a copy-paste regression introduced during the inline duplication.

## Failure mode

When a hypothetical Qwen-MoE model with MQ3 weights (e.g. an A3B-class model quantized via `--format mq3`, if/when added) hits these paths:

1. `is_mq = false` for MQ3 weights → activation goes through plain `rmsnorm_batched` instead of `fused_rmsnorm_rotate_mq_batched`
[...]

PR #180, opened 2026-05-06
fix(quantizer): Q8 router default for MoE — fixes #171

### LINKED ISSUE #171: Qwen3.5-MoE family attractors at temp=1.0 on long agent prompts; greedy also fails
## Summary

Qwen3.5-MoE-family models (3.5-A3B, 3.6-A3B, 3.5-122B-A10B) at MQ4/MQ6-experts produce token-level structural attractors on multi-paragraph self-referential agent prompts when sampled with the HF-recommended `temperature=1.0, top_p=0.95, top_k=20, min_p=0.05` config. Failure is **not** 3.6-specific (3.5-A3B mq4-mq6exp-port also fails on the same prompt), is **not** an axis the routed-expert weight quantization can repair, and is **not** in the routed-expert MoE GEMV kernels. Greedy decode (`temp=0.0`) also fails on the same prompt → the forward pass produces borderline-coherent logits that wider sampling amplifies into structural loops.

Originally reported by a user running the pi-agent with hipfire 3.6-A3B: "every prompt takes 1-2 min to load then everything falls apart one way or another (max_tokens 65536, max_seq 131072, thinking on, dflash off)."

## Reproduction

Hipx (Strix Halo gfx1151), commit `d67bf02` on `feat/moe-expert-heatmap` (PR #167):

```
load qwen3.6-35b-a3b.mq4, kv_mode=asym3, max_seq=8192
prompt: ~370-token agent-style multi-paragraph "review hipfire repo before PR" prompt
generate temperature=1.0 top_p=0.95 top_k=20 min_p=0.05 thinking=true max_tokens=800
```

Result: 199 words of attractor garbage ("Wait,Iam reviewing... rewritewrittenwritebetterwritten... Scripts/\\ Scripts/\\ scripts..."), self-EOS at ~3391 chars.

Same prompt + sampler on **dense Qwen3.5-4B/9B/27B mq4** produces 529-560 words of verbose-but-coherent meta-thinking. No attractor.

## Operational fix (LANDS NOW)

**Drop default temperature to 0.3 for Qwen3.5-MoE family at MQ4.** At `temp=0.3 + top_k=20 + min_p=0.05`, 3.6-A3B produces a clean coherent 103-word answer with self-EOS. The model can do the prompt; HF's recommended temp=1.0 is too wide for our forward at MQ4.

Where: per-arch SamplerConfig default flip in `crates/hipfire-runtime/examples/daemon.rs` (or in the new SamplerConfig::hf_thinker constructor introduced in PR #167).

Side recommendation: **default `kv_mode=asym3` over `kv_mode=q8`** — with q8, 3.6-A3B at temp=1.0 degenerates to literal letter-soup ("MMEEEKKKAALLCCCC") much faster. asym3 is the safe choice for MoE.

## Investigation log (12 hypotheses, 11 negative, 1 partial)

All probes against `qwen3.6-35b-a3b.mq4` and `qwen3.5-35b-a3b.mq4-mq6exp-port` on hipx with the agent prompt and HF sampler.

### Settled NEGATIVE (cannot be the bug):

1. **Per-expert weight magnitude differs 3.6 vs 3.5** — NumPy absmax/median scan on hiptrx of every routed expert × all 40 layers. Result: 3.6 is **0.97-0.99× of 3.5** on every distributional metric. `gate_up_proj` p99 mean ratio ≈8.9 on both; `down_proj` p99 mean ≈920k-940k on both (with ≈37M max — explained below). 3.6 is fractionally **lighter-tailed**, not heavier.
2. **MQ6-experts cures 3.6** — quantized 3.6-A3B with `--format mq4-mq6exp` (26.75 GB), smoked. 469 tokens of "VLvariant qlArchitecture" attractor. **Critical control: 3.5-A3B mq4-mq6exp-port on the SAME prompt produces 3927 chars of "Wait wait wait" attractor too.** Failure is not 3.6-specific.
3. **GPU topk distribution-sensitive defect (cpu-overwrite gate)** — ported HIPFIRE_MOE_TOPK_CPU_OVERWRITE from `debug/moe-qwen-20260505` into modular crate. Verified firing (greedy gate-on diverges from gate-off on hexagons prompt: 559 vs 488 chars). On agent prompt: gate-on H.out 118 tok / gate-off F.out 271 tok, different surface form, same structural failure. **CPU FP64 topk does not cure**. Path B's `moe_topk_renorm_k8` is correct.
4. **Per-row absmax outlier-aware quant** — simulated MQ4G256 / MQ4G64 / MQ4G256+sidecar{4,16,64} / MQ6G256 reconstruction error on 768 worst-tailed rows of 3.6-A3B `down_proj`. **Cosine sim ≥0.991 across ALL schemes**. Sidecar at sensible byte budgets does not beat MQ6G256 (sidecar4: 1.1× lower MSE, sidecar16: 1.4×, MQ6G256: 17.6×). Per-element quant fidelity is fine.
5. **Atomic-add ordering in routed-down combine** — wrote diagnostic kernel pair (per-(row,krank) scratch + serial reduction). On 3.5: bit-identical to atomic-add path (3927/555 same as base). On 3.6: different attractor (3353/301 vs 3391/199), not cured. **Atomic-add ordering is not the catastrophic bug.**
6. **FP32 dot-product accumulator in routed-down GEMV** — wrote FP64-accumulator drop-in (`gemv_hfq4g256_moe_down_diag_fp64`). Same result pattern as #5: 3.5 bit-identical, 3.6 shifts attractor but not cured.
7. **FP32 dot-product accumulator in gate_up GEMV** — wrote FP64-accumulator drop-in for the gate_up kernel (K=2048, 4× more rounding compounding than down). 3.5 bit-identical to base. **3.6 became catastrophically worse — descended to literal letter-soup gibberish ("hxhhhqhzqdvdhdxdxd...") within ~140 words.** FP64 in gate_up makes 3.6 cliff harder, not better.
8. **Disable shared expert** — different shorter failure (10 words "## 12. ## Summary 756" on 3.6, 129 words CJK reverse-loop on 3.5). Both shared and routed are required.
9. **Disable routed experts** — different 49-word fa0592d-paraphrase soup on 3.6, 477 words CJK "网络/Network" on 3.5. Both required.
10. **HIPFIRE_GRAPH=0 / HIPFIRE_GRAPH_MOE=0 / HIPFIRE_FP16=0 / HIPFIRE_KV_MODE=fp32 env probes** — bit-identical output for all. Either env vars don't toggle on this code path or runtime params override.
11. **Dense 4B/9B/27B Qwen3.5 mq4 same prompt + sampler** — all three produce verbose-but-coherent meta-thinking output. No attractor. **Bug is MoE-specific in our forward path, not MQ4-specific or prompt-specific.**

### PARTIAL (operational fix, not root cause):

12. **Sampler temperature** — temp=0.3 produces clean coherent 103-word answer on 3.6-A3B agent prompt. temp=0.7 partial coherence, temp=1.0 attractor. **But greedy (temp=0.0) ALSO fails** with the same attractor class. So the sampler isn't the source — wider sampling amplifies a forward-pass borderline-coherence problem.

## Side-finding worth keeping in memory

3.5 and 3.6 are bit-invariant under several deep numerical interventions (atomic-add ordering, FP64 down, FP64 gate_up) — but only 3.5 is truly invariant. **3.6 shifts attractor under every change**, indicating its argmax decisions sit close to FP32 rounding boundaries. 3.5's same path is far from those boundaries. The Qwen team appears to have pushed 3.6's quality higher at the cost of precision robustness — common for later iterations of the same architecture.

The earlier finding "down_proj has 37M-tail outliers across both models" was on **raw weights**, not what MQ4G256-FWHT actually quantizes. Group-wise FWHT-rotated MQ4G256 absorbs most outlier energy; per-element cosine similarity to original is ≥0.991 across all candidate schemes including the current one. The raw-weight tail-of-tails finding does not translate to a quant brittleness.

## What we have NOT done that would actually root-cause it

The bug is either (A) intrinsic Qwen3.5-MoE family fragility under MQ4 at temp=1.0 on hard prompts, or (B) somewhere upstream of routed-expert MoE GEMV kernels — LinearAttention/DeltaNet drift across 30 layers, RMSNorm precision, FullAttention output_gate path, or x_residual accumulator across the 40-layer MoE stack. We cannot distinguish (A) vs (B) without a reference comparison:

- **llama.cpp same model + same prompt + same sampler** (~1 hour). If llama.cpp produces coherent output → (B), our forward has a bug. If llama.cpp also degrades → (A), model fragility.
- **CPU reference forward in NumPy/safetensors** matching the quant scheme exactly. Layer-by-layer activation diff vs hipfire forward. ~4-8 hours, definitive — localizes divergence to a specific op in a specific layer.

Both are valid follow-up work for a future investigation; neither is needed for the operational fix.

## Files

- All artifacts at `/tmp/moe-3host-debug/` (gitignored, not committed)
- Per-host investigation notes: `/tmp/moe-3host-debug/INVESTIGATION.md`
- absmax stats: `/tmp/moe-3host-debug/absmax_results.json`
- quant reconstruction prototype: `/tmp/moe-3host-debug/quant_recon_error.py`, results in `quant_recon_results.json`

## Action items

- [ ] Flip default `temperature` to 0.3 for Qwen3.5-MoE family at MQ4 (operational fix that cures user-reported failure)
- [ ] Document `kv_mode=q8` is unsafe for Qwen3.5-MoE — degenerates to letter-soup. Default `asym3`.
- [ ] (Optional, future) llama.cpp reference comparison to determine model-fragility vs forward-bug
- [ ] (Optional, future) CPU reference forward for layer-by-layer activation diff if (B) implicates a forward bug

fivetide:
## Summary

- **Always quantize MoE router weights at Q8** (`mlp.gate.weight`, `mlp.shared_expert_gate.weight`) regardless of `--format`. The `is_q8_tensor` guard existed but only fired in `q8-mixed`/`q8-fast` modes — now it fires for all MoE models (`num_experts > 0`).
- Add `dequant_q5_k()` for GGUF Q5K tensor support
- Add `--format hfq4`/`hfq6` support for MoE expert weights (was MQ4/MQ6-only)
- Add `--q8-router` explicit flag (redundant for MoE, useful for non-MoE)

## Root cause

4-bit quantization of the 256×2048 router weight destroys routing precision: 152/256 expert rows drop below 0.99 cosine similarity at HFQ4G256 (3× worse MSE than Q4_K_M). This flips top-8 expert selection, cascading into structural attractors across 40 MoE layers.

## Evidence

llama.cpp parity test confirmed attractors are hipfire-specific (not model fragility):

| Format | Router | Unique words | Verdict |
|---|---|---|---|
| MQ4 (4-bit router) | 4-bit | 14% | **ATTRACTOR** |
| **MQ4 + Q8 router** | **8-bit** | **46%** | **CLEAN** |
| HFQ4 (4-bit router) | 4-bit | 10% | **ATTRACTOR** |
| **HFQ4 + Q8 router** | **8-bit** | **42%** | **CLEAN** |
| **HFQ6** | **6-bit** | **70%** | **CLEAN** |
| Q4_K_M (llama.cpp) | mixed-6bit | 58% | CLEAN |

Cost: ~10 MB on 18 GB model (0.05%). No perf impact.

## Test plan

- [x] MQ4 + Q8 router: clean on 3.6-A3B code review prompt (greedy, RP=1.05)
- [x] HFQ4 + Q8 router: clean on same prompt
- [x] HFQ6: clean on same prompt
- [x] 3.5-A3B MQ4 (no Q8 router): still clean (fa0592d was sufficient for 3.5)
- [x] Coherence gate `--full`: all models pass (0 hard errors), 3.5-A3B and 3.6-A3B both coherent on sheep prompt with re-quantized MQ4 models
- [x] 122B-A10B: re-quantized from original safetensors with Q8 router. **CLEAN** — 59.9% unique words, coherent code review (487 tok, self-EOS). Previous 4-bit router baseline was borderline (54.7% unique, mild tail repetition).

Fixes #171

### DIFF
--- a/crates/hipfire-quantize/src/gguf_input.rs
+++ b/crates/hipfire-quantize/src/gguf_input.rs
@@ -412,6 +412,63 @@ fn dequant_q4_k(data: &[u8], n: usize) -> Vec<f32> {
     out
 }
 
+fn dequant_q5_k(data: &[u8], n: usize) -> Vec<f32> {
+    let block_size = 256;
+    let block_bytes = 176;
+    let nblocks = (n + block_size - 1) / block_size;
+    let mut out = vec![0.0f32; n];
+    for b in 0..nblocks {
+        let off = b * block_bytes;
+        if off + block_bytes > data.len() {
+            break;
+        }
+        let d = f16_to_f32(u16::from_le_bytes([data[off], data[off + 1]]));
+        let dmin = f16_to_f32(u16::from_le_bytes([data[off + 2], data[off + 3]]));
+
+        // 12-byte packed scales/mins — same layout as Q4_K
+        let sc_data = &data[off + 4..off + 16];
+        let mut scales = [0u8; 8];
+        let mut mins = [0u8; 8];
+        for i in 0..4 {
+            scales[i] = sc_data[i] & 63;
+            mins[i] = sc_data[4 + i] & 63;
+        }
+        for i in 0..4 {
+            scales[4 + i] = (sc_data[8 + i] & 0xF) | ((sc_data[i] >> 6) << 4);
+            mins[4 + i] = (sc_data[8 + i] >> 4) | ((sc_data[4 + i] >> 6) << 4);
+        }
+
+        // 32 bytes of high bits (1 bit per element), then 128 bytes of low nibbles
+        let qh = &data[off + 16..off + 48];
+        let ql = &data[off + 48..off + 176];
+
+        for group in 0..4 {
+            let sb_even = group * 2;
+            let sb_odd = group * 2 + 1;
+            let sc_even = d * scales[sb_even] as f32;
+            let m_even = dmin * mins[sb_even] as f32;
+            let sc_odd = d * scales[sb_odd] as f32;
+            let m_odd = dmin * mins[sb_odd] as f32;
+            for l in 0..32 {
+                let byte = ql[group * 32 + l];
+                let hbit = ((qh[l] >> group) & 1) as u8;
+                let hbit2 = ((qh[l] >> (group + 4)) & 1) as u8;
+                let idx_even = b * block_size + group * 64 + l;
+                let idx_odd = idx_even + 32;
+                if idx_even < n {
+                    let q = ((byte & 0x0F) | (hbit << 4)) as f32;
+                    out[idx_even] = q * sc_even - m_even;
+                }
+                if idx_odd < n {
+                    let q = (((byte >> 4) & 0x0F) | (hbit2 << 4)) as f32;
+                    out[idx_odd] = q * sc_odd - m_odd;
+                }
+            }
+        }
+    }
+    out
+}
+
 fn dequant_q6_k(data: &[u8], n: usize) -> Vec<f32> {
     let block_size = 256;
     let block_bytes = 210;
@@ -480,6 +537,7 @@ pub fn tensor_to_f32(info: &TensorInfo, data: &[u8]) -> Vec<f32> {
         GgmlType::Q4_0 => dequant_q4_0(data, n),
         GgmlType::Q8_0 => dequant_q8_0(data, n),
         GgmlType::Q4K => dequant_q4_k(data, n),
+        GgmlType::Q5K => dequant_q5_k(data, n),
         GgmlType::Q6K => dequant_q6_k(data, n),
         other => panic!(
             "GGUF tensor type {:?} not implemented (tensor: {})",
--- a/crates/hipfire-quantize/src/main.rs
+++ b/crates/hipfire-quantize/src/main.rs
@@ -1701,6 +1701,7 @@ fn main() {
     let use_mq3g256 = format == "mq3" || format == "mq3g256";
     let use_mq2g256 = format == "mq2" || format == "mq2g256";
     let use_hfq6 = format == "hfq6" || format == "hfq6g256" || format == "hf6";
+    let q8_router_flag = args.iter().any(|a| a == "--q8-router");
 
     // ── Sub-4-bit guards (2026-04-30 sweep) ─────────────────────────────
     // MQ2 with the current uniform 4-level codebook collapses at every
@@ -1790,6 +1791,11 @@ fn main() {
     };
     eprintln!("Architecture: {arch_str} (id={arch_id})");
     let is_moe = arch_id == 6;
+    // Q8 router: always on for MoE models. 4-bit router quantization destroys
+    // routing precision on precision-sensitive models (Qwen3.6-A3B: 152/256
+    // expert rows drop below 0.99 cosine similarity at HFQ4G256). Cost: ~0.05%
+    // model size. See github.com/Kaden-Schutt/hipfire/issues/171.
+    let q8_router = is_moe || q8_router_flag;
     if is_moe {
         eprintln!("  MoE detected — will split 3D expert tensors per-expert before quantization.");
     }
@@ -1906,7 +1912,10 @@ fn main() {
             let signs1 = gen_fwht_signs(42, 256);
             let signs2 = gen_fwht_signs(1042, 256);
             let inner_k = inner_shape[1] as usize;
-            let supports_mq4 = inner_k % 256 == 0;
+            let supports_g256 = inner_k % 256 == 0;
+            let expert_mq6 = use_mq6g256 && supports_g256;
+            let expert_hfq6 = use_hfq6 && supports_g256;
+            let expert_hfq4 = use_hfq4g256 && supports_g256;
 
             // Parallelize across the 256 expert slices via rayon. Each slice
             // dequant→FWHT→quant→pack is a CPU-bound, self-contained job.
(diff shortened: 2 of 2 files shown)


unverbraucht:
Awesome! Just to be sure, could you check if this issue might apply to this work? https://github.com/Kaden-Schutt/hipfire/issues/179

fivetide:
Thanks for flagging #179! Checked — our changes are not affected:

1. **Q8 router guard** (`q8_router && is_q8_tensor`) fires before any format-specific branch, so router weights always go to Q8F16 regardless of the downstream MQ3/MQ4/MQ6 dispatch. The `is_mq` matcher gap in #179 is in the runtime forward pass, not the quantizer.

2. **New `--format hfq4`/`hfq6` expert paths** produce `HFQ4G256`/`HFQ6G256` typed tensors, which take the non-MQ code path at runtime (`gate_side_mq4 = false` → `weight_gemv` → `gemv_hfq4g256`, no FWHT rotation needed). So the missing `MQ3G256` in the MoE-batched `is_mq` matcher doesn't interact with these new format options.

3. The core fix (unconditional Q8 router for `is_moe`) is purely in the quantizer — it doesn't touch `qwen35.rs` dispatch at all.

That said, #179 is a real latent bug that should be fixed separately — if someone deploys A3B+MQ3 it would silently corrupt prefill. Worth a one-line fix per the issue's suggestion.

Issue #134, opened 2026-05-03
Extremely slow prompt processing


subhamhimself:
I (somehow) got this working on windows natively (Qwen3.6 27B, and its draft, on a 7900xtx), and test results look promising.
I'm using Cline as harness, and the prompt processing seems to be exceedingly slow, presumably because it seems to be running on CPU.
I see very low GPU usage and CPU usage is constant around 6% which makes me think it's using CPU.

Note that the prompt processing isn't slow just for first call, it takes 5-10 min after every tool call/chat.

Willing to collaborate on this if I could get some pointers. 

Thanks!

Kaden-Schutt:
Triage update — likely tokenizer O(n²) behavior, not unaccelerated forward pass.

A few observations from the configuration shape you describe:
- 7900 XTX (gfx1100) is one of the well-tested arches for hipfire; the GPU forward path is not the suspect.
- 6% CPU + low GPU during prompt processing matches what you'd see if the daemon is in the tokenizer rather than the GPU. The current tokenizer is BPE-based and has O(n²) behavior for very long input strings (recent profiling on 4k+ context shows it as the top bottleneck on prefill wall-clock).
- 5-10 min per Cline tool-call is consistent with the harness re-sending the full conversation (system + history + tool results + new user message) every turn, which means tokenizer cost compounds over the conversation.

Two questions to narrow this down:

1. Are you running with `HIPFIRE_GRAPH=1` in the daemon environment? It wraps the decode loop in a hipGraph and substantially reduces decode launch overhead. Without it, decode tok/s on Windows can drop noticeably.
2. Could you grab a `daemon --precompile` log on cold-start to confirm kernels are JITed for gfx1100 specifically? `daemon` lives at `~/.hipfire/bin/daemon` after install. If you see "device kernel image is invalid" or arch-mismatch warnings, that's a separate issue.

Also worth checking: how long is each turn's full prompt (system + history)? If it's 8k+ tokens, tokenizer cost dominates regardless of GPU. Long-context tokenizer optimization is on the roadmap (see #173 for the bench-harness reload variant of the same problem).

Local repro on the same hardware (gfx1100, k9lin) at typical Cline-style prompt lengths reproduces the slowdown shape; the fix is server-side (PFlash long-context prefill compression, which lands as a separate PR track).

Willing to dig deeper on a sample prompt + repro steps if you can share. Flag this as `bug` in the meantime.

unverbraucht:
[...]

PR #201, opened 2026-05-08
fix(tokenizer): O(N²) → O(N log N) GPT-2 BPE — long-prompt prefill no longer stalls

alpineQ:
## Summary

The GPT-2 BPE encoder in `Tokenizer::encode_gpt2_bpe` was an O(N²) naive implementation. On long prompts (32K+ tokens) the daemon spent **5–10 minutes purely in the tokenizer** before any GPU work began. This PR replaces it with the textbook O(N log N) priority-queue BPE algorithm. **Output is byte-identical to the HuggingFace `tokenizers` reference.**

End-to-end on 7900 XTX (gfx1100), Qwen3.6-27B mq4, 32K-token prompt, AR prefill:

| | wall clock |
|---|---|
| before | **805 s** (~670 s tokenizer + 125 s GPU + decode) |
| after  | **134 s** (~2 s tokenizer + 125 s GPU + 7 s decode) |

`scripts/coherence-gate.sh` short battery: all 6 available models pass with no panics, no zero-token outputs, no verbatim loops.

This very likely resolves the symptoms reported in **#134** ("Extremely slow prompt processing").

## How we ran into it

I'm Claude (Anthropic, Opus 4.7), working with @alpineQ on a benchmark comparing decode tok/s on Qwen3.6-27B Q4 across LM Studio, llama.cpp + MTP (PR #22673), and hipfire + DFlash, at 16K / 32K / 64K contexts on a dual 7900 XTX setup.

The decode numbers themselves came out fine — hipfire AR-only at 16K was 38.4 tok/s, faster than llama.cpp + MTP at 32.3 tok/s on the same prompt. But the wall-clock per request was very strange: 16K AR prefill ran in ~3 minutes, 32K ended up taking ~13 minutes, and 64K AR pp=2 timed out at 25 minutes without producing a single output token. Meanwhile `done`-line `prefill_ms` reported by the daemon was a sane ~120 s for 32K, and `decode_tok_s` reported the expected ~33 tok/s — the ~13-minute wall did not show up in the daemon's own timing.

I tried a few wrong diagnoses first:

1. I read the per-token FA fallback at `qwen35.rs:4760-4783` and assumed it was firing for our config. It wasn't — `fa_batched_ok` is true for `mq4 + asym3` and the batched FA arm at `qwen35.rs:4380` runs.
2. I added in-process scope timers (`forward_prefill_chunk`, `LA_dense_batched`, `FA_dense_batched`, etc.) and reached the totals 99.8% in `chunk_embed_section`, 0.2% in the layer loop. I assumed the embedding section's `memcpy_htod` was acting as a sync point and the work was real GPU compute; rocBLAS being slow was an attractive story.

@alpineQ pointed out that this was misleading: the daemon was not pinning the GPU (rocm-smi consistently showed 5% utilisation), it was clearly burning CPU in something useless, and a sampling profiler is the right tool to find it. So I attached `perf record -F 500 -g --call-graph dwarf` to the running daemon for 15 seconds while it was stuck mid-prefill. The result was unambiguous:

```
21.6%  hipfire_runtime::tokenizer::Tokenizer::encode_raw
17.1%  core::hash::sip::Hasher::write
13.5%  alloc::string::String::clone
12.1%  core::hash::BuildHasher::hash_one
10.7%  malloc
 8.2%  _int_free
 5.4%  cfree
 4.9%  __memmove_avx512_unaligned_erms
 ────
 ~93%  tokenizer + allocator + hashing
```

The daemon was not in `forward_prefill_chunk` at all — it was synchronously tokenising the prompt, dispatching no GPU work, holding two of the HIP helper threads asleep in `kfd_wait_on_events`. Maintainer's earlier triage on #134 (suspecting the tokenizer) was correct; my detour through GPU-kernel scopes was not.

This also explains why the maintainer's #134 hypothesis matched the symptoms but was hard to verify without a sampling profiler — `prefill_ms` in the `done` line is GPU-only timing, so any tokenizer cost is invisible to the daemon's self-report. Only the wall clock saw it, and only the wall clock showed a 6× discrepancy.

## What was wrong

`crates/hipfire-runtime/src/tokenizer.rs:467` — `encode_gpt2_bpe` (pre-fix):

```rust
let mut symbols = byte_tokens;
let merge_rank: HashMap<(String, String), usize> = ...;
loop {
    let mut best_rank = usize::MAX;
    let mut best_idx = 0;
    for i in 0..symbols.len() - 1 {                       // ← full N-element scan per merge
        let pair = (symbols[i].clone(), symbols[i + 1].clone());  // ← 2 String allocations per pair
        if let Some(&rank) = merge_rank.get(&pair) {       // ← hash on (String, String) tuple
            if rank < best_rank { best_rank = rank; best_idx = i; }
        }
    }
    if best_rank == usize::MAX { break; }
    let merged = format!("{}{}", symbols[best_idx], symbols[best_idx + 1]);
    symbols[best_idx] = merged;
    symbols.remove(best_idx + 1);                          // ← O(N) Vec shift on every merge
}
```

For an input of N starting symbols this does ~N merges; each merge does an O(N) full-array scan with two String clones per inspected pair, plus an O(N) `Vec::remove` shift. **Asymptotically O(N²) merges × O(N)-with-allocation per scan**, with a punishing constant factor from the per-pair `String::clone` and `(String, String)` hashing.

For our typical bench prompts:

| prompt | starting symbols | rough cost |
|---|---|---|
|   200 tokens | ~700 bytes | <1 ms — invisible |
| 16384 tokens | ~50 KB     | ~30 s   — annoying |
| 32524 tokens | ~135 KB    | ~670 s  — appears as a hang |
| 65285 tokens | ~270 KB    | ~50 min — looks broken |

The hipfire canonical perf battery (CLAUDE.md) uses ~200-token prompts, so this never showed up in regression gates — only on long-context production workloads, which is exactly what #134 reports.

## What this PR does

Replaces the naive scan with the standard priority-queue + doubly-linked-list BPE encoder used in HuggingFace `tokenizers`, OpenAI `tiktoken`, and most modern BPE implementations. I did not invent the algorithm — this is well-trodden ground. The structure:

1. Symbols held as a doubly-linked list of slot indices (`prev[i]`, `next[i]`); merged slots tombstoned via `dead[i]`. Splice instead of `Vec::remove`.
2. `gen[i]` increments every time slot `i` absorbs its right neighbour, so stale heap entries are O(1)-discardable on pop.
3. Min-heap on `(rank, left_idx, gen_at_push)`. Each pair is pushed at most once per generation, so total pushes ≤ 3 N (initial + 2 per merge).
4. On pop: validate `!dead[l] && gen[l] == gen_at_push`, re-verify rank against the live pair (paranoia guard against state-machine drift), splice, push the two newly-adjacent pairs.

Total complexity: **O(N log N)** with a small constant. On 32K, that's ~500 K heap operations instead of ~10⁹ scan operations.

## Verification

1. **Byte-identical to HuggingFace reference.** Encoded `prompt_16384.txt` (16132 tokens) with both the new hipfire encoder and HF `tokenizers` (Qwen3.6 `tokenizer.json`):
   ```
   $ diff hipfire-tokens.txt hf-tokens.txt | wc -l
   0
   ```
   Token IDs match exactly. BPE merge order is deterministic when ranks are unique (true for the GPT-2 vocab), so this isn't surprising — both implementations apply the same merges in the same order.

2. **`scripts/coherence-gate.sh` (short battery)** passes on all 6 available models — `0.8b mq4` (cap), `4b mq4` (code), `9b mq4` (reason), `9b mq4` (tool-call), `9b mq3` (reason), `27b mq3` (cap). No panics, no zero-token outputs, no verbatim loops.

3. **Wall-clock perf**, single-GPU 7900 XTX (gfx1100), `qwen3.6-27b.mq4`, asym3 KV, AR prefill (no DFlash), 32K-token prompt, 256 generated tokens, identical prompt MD5 across runs:

   |  | wall_s | daemon `prefill_ms` | `decode_tok_s` |
   |---|---:|---:|---:|
   | before | 805 | 119 681 | 33.3 |
   | after  | 134 | 124 864 | 31.7 |

   `prefill_ms` and `decode_tok_s` are within their normal day-to-day variance — this PR doesn't touch GPU code, so the kernel-side timing should be unchanged. The 6× wall-clock improvement is entirely from removing the tokenizer stall.

## Footprint

- One file changed: `crates/hipfire-runtime/src/tokenizer.rs` (+131, −36).
- No new dependencies.
- No public API change. `Tokenizer::encode` still takes `&str`, returns `Vec<u32>`.
- No model output change (token IDs are identical).

## Things I deliberately did *not* do in this PR

- I did not pre-build the `merge_rank: HashMap<(String, String), usize>` once at `Tokenizer::from_*` construction. That's an additional O(M) one-time cost saved per encode call, but it's a separate optimisation and would change the struct layout — keeping this PR strictly to the algorithmic fix.
- I did not switch the encoder to indexed-key (`(u32, u32)`) heap entries. Doing so would eliminate the remaining `String::clone`s on heap push, but it requires a `(token_id, token_id)`-keyed merge rank map and is again a separate, larger change.
- I did not touch `encode_sentencepiece` (LLaMA path). It has its own greedy-longest-match loop that may have its own scaling issues, but it's not what #134 reports and not what this PR addresses.

Both follow-ups are easy single-commit changes if the maintainers want them as separate PRs.

### DIFF
--- a/crates/hipfire-runtime/src/tokenizer.rs
+++ b/crates/hipfire-runtime/src/tokenizer.rs
@@ -484,9 +484,16 @@ impl Tokenizer {
     ///    splice out the right neighbor; push the two newly-formed pairs
     ///    `(prev[l], l)` and `(l, next[l])`.
     ///
-    /// Result is byte-identical to the naive implementation (BPE merge
-    /// order is deterministic when ranks are unique, and GPT-2 BPE
-    /// merges are unique by construction).
+    /// Byte-identicality with the naive scan — load-bearing invariant:
+    /// the heap key is `(rank, l, gen_at_push)`, ordered lexicographically.
+    /// On equal `rank`, the smaller `l` (leftmost surviving symbol) wins
+    /// the pop. This matches the naive scan's `if rank < best_rank`
+    /// strict-`<` tiebreak (first-occurrence-wins from the left), which
+    /// is the BPE encoding contract HuggingFace `tokenizers` and OpenAI
+    /// `tiktoken` follow. Future refactors that change the heap key
+    /// ordering MUST preserve this leftmost-on-tie property or the
+    /// encoded token IDs will silently diverge from every reference
+    /// implementation.
     fn encode_gpt2_bpe(&self, text: &str) -> Vec<u32> {
         // 1. Convert text to GPT-2 byte-encoded symbols.
         let mut syms: Vec<String> = text
@@ -543,11 +550,7 @@ impl Tokenizer {
         // splice is O(1); two pushes are O(log N). Total O(N log N).
         while let Some(Reverse((rank, l, gen_at_push))) = heap.pop() {
             // Validate: slot still alive, generation matches, right neighbor
-            // exists, and current pair still has the popped rank (the rank
-            // we expected). The rank-recheck guards against the case where
-            // `l` was the *right* side of a merge that absorbed it via a
-            // different code path — should not happen with this state
-            // machine, but cheap to verify.
+            // exists. Stale heap entries are dropped here cheaply.
             if dead[l] || gen[l] != gen_at_push {
                 continue;
             }
@@ -556,16 +559,18 @@ impl Tokenizer {
                 continue;
             }
             let r = r as usize;
-            // Optional: re-verify rank for paranoia; skip if it diverges.
-            // In the strict pq model with generation tagging this should
-            // never trigger, but cost is one HashMap lookup vs silent
-            // corruption risk.
-            let cur_rank = merge_rank
-                .get(&(syms[l].clone(), syms[r].clone()))
-                .copied();
-            if cur_rank != Some(rank) {
-                continue;
-            }
+
+            // The gen-tag invariant guarantees the popped rank still describes
+            // the live `(syms[l], syms[r])` pair: any merge that could change
+            // either side bumps `gen[l]` (or kills `r`) and would have failed
+            // the check above. Verified in debug builds; release trusts it.
+            debug_assert_eq!(
+                merge_rank
+                    .get(&(syms[l].clone(), syms[r].clone()))
+                    .copied(),
+                Some(rank),
+                "BPE pq invariant: popped rank must match live pair rank",
+            );
 
             // Apply the merge: l absorbs r.
             let merged = {
@@ -576,7 +581,11 @@ impl Tokenizer {
             };
             syms[l] = merged;
             dead[r] = true;
-            gen[l] = gen[l].wrapping_add(1);
+            // Plain `+= 1`: bumps are bounded by N − 1 per slot, N < 2³², so
+            // overflow is unreachable. `wrapping_add` would silently un-stale
+            // heap entries on overflow — wrong semantics. Debug panics, release
+            // aborts: both communicate that wraparound is a bug, not a feature.
+            gen[l] += 1;
 
             // Splice r out of the linked list.
             let nr = next[r];
@@ -590,9 +599,9 @@ impl Tokenizer {
             if pl >= 0 {
                 // `l`'s left neighbor's right pair is now (pl, l) with new
                 // syms[l]; bump pl's gen so its old heap entries die. (Not
-                // strictly required since we recheck rank on pop, but tightens
+                // strictly required since we revalidate on pop, but tightens
                 // the invariant.)
-                gen[pl as usize] = gen[pl as usize].wrapping_add(1);
+                gen[pl as usize] += 1;
                 push_pair(&mut heap, &syms, &gen, pl as usize, l);
             }
             if next[l] >= 0 {
@@ -601,12 +610,14 @@ impl Tokenizer {
         }
 
         // 7. Walk the linked list collecting live symbols → token ids.
-        // Find head: the slot with prev == -1 that's still alive. Slot 0
-        // is the head unless it was absorbed (which can't happen since
-        // BPE only merges right-into-left).
-        debug_assert!(!dead[0], "BPE merged into slot 0 — algorithm invariant violated");
-        let mut result = Vec::with_capacity(syms.len());
-        let mut p: i32 = 0;
+        // Slot 0 is always the head under our merge-left-into-right invariant,
+        // but we scan explicitly for `prev == -1 && !dead`. The scan is O(N)
+        // once and is robust against any future invariant breakage (a
+        // `debug_assert!` here would be a release-mode no-op and walk
+        // tombstoned data silently; the explicit scan can't).
+        let mut result = Vec::with_capacity(n);
+        let head = (0..n).find(|&i| prev[i] == -1 && !dead[i]);
+        let mut p: i32 = head.map(|i| i as i32).unwrap_or(-1);
         while p >= 0 {
             let pi = p as usize;
             result.push(self.token_to_id.get(&syms[pi]).copied().unwrap_or(0));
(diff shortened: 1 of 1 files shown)


unverbraucht:
Multi-agent review (Claude Opus 4.7 + Gemini CLI + GLM-5) — thanks for the
clean writeup, the perf trace + before/after numbers + honest mis-diagnosis
walkthrough is genuinely well done. All three reviews land at
approve-with-changes; the three of us converge on two asks before merge plus
a couple of quick wins, with two follow-up issues filed for the rest.

**Before merge**

1. **Add one in-tree unit test locking the byte-identical claim.** Right now
   the only correctness evidence is an out-of-band shell `diff` against HF
   `tokenizers` on a single 16K prompt — nothing in CI prevents the next
   refactor (e.g. the indexed-key follow-up you mention) from silently
   regressing. A 5-line `#[test]` that encodes a known fixture and asserts
   exact `Vec<u32>` output would close the gap.

2. **Remove the rank-recheck inside the pop loop**
   (`merge_rank.get(&(syms[l].clone(), syms[r].clone()))` after the gen
   check). It costs two `String::clone`s per pop and `continue`s silently
   on mismatch. The doc-comment correctly notes this should never trigger
   under the gen-tag invariant — so either it's pure overhead, or it's
   masking a real bug. Either way, silently swallowing an unreachable state
   is the worst of the three options. If you'd like the safety net,
   `debug_assert_eq!(cur_rank, Some(rank))` keeps the check without the
   release-mode cost or the silent-corruption path.

**Quick wins worth doing in the same pass**

3. **`debug_assert!(!dead[0])` is a no-op in release builds.** If the
   invariant ever breaks, release-mode walks tombstoned data silently
   instead of panicking. Two clean options: (a) trust the "left-into-right
   merges only" invariant and drop the assert entirely, or (b) replace the
   `let mut p: i32 = 0;` with an explicit head-find scan
   (`prev[i] == -1 && !dead[i]`). Pick whichever; the current shape
   advertises protection that release builds don't actually provide.

4. **Replace `gen[l].wrapping_add(1)` with `gen[l] += 1` (and same for
   `gen[pl]`).** `wrapping_add` communicates "wraparound is fine" when in
   fact wraparound would silently un-stale heap entries. Plain `+= 1`
   panics in debug, aborts in release on overflow, and documents the
   invariant correctly. Vacuously safe either way (per-encode bumps are
   bounded by N − 1, N < 2³²) — this is just a one-line clarity fix.

**Doc nit**

5. The byte-identicality of the new pq-BPE vs the naive scan is
   load-bearing on the heap key being `(rank, l, gen)` — equal ranks
   → smallest slot index → leftmost surviving symbol → same tiebreak as
   the naive `<` scan. Worth one sentence in the doc-comment so a future
   contributor doesn't refactor the heap key without realizing it's the
   byte-identicality invariant.

**Follow-ups (filed as separate issues, not blocking)**

- #202 — tokenizer hot-path hardening: per-call `merge_rank` rebuild (the
  next-largest CPU cost after this PR; matters for #134's per-tool-call
  symptom which a single-prompt bench under-measures) plus the parallel
  `encode_sentencepiece` per-trial `String` allocation.
- #203 — pre-existing correctness bug (not introduced by this PR):
  `unwrap_or(0)` in both encoders silently corrupts encode if vocab and
  merges are ever loaded inconsistently.

Verdict from all three agents: approve once (1) and (2) land. Happy to
look at the v2 diff when it's up.

— Reviewed by Claude Opus 4.7 (1M ctx), Gemini CLI, and GLM-5

alpineQ:
Thanks for the careful review. All five items addressed in `fc49e6f`:

1. **Unit tests** — added a `bpe_tests` module with six cases: full merge cascade, partial merge, no-merges, empty/single-byte, 1024-byte priority-queue stress, and the leftmost-on-tie invariant. All pass.
2. **Rank-recheck removed** — replaced with `debug_assert_eq!` so the gen-tag invariant is verified in debug without paying the release cost or masking the failure mode.
3. **Head pickup** — went with explicit `(0..n).find(|i| prev[i] == -1 && !dead[i])` instead of trusting slot 0; old `debug_assert!(!dead[0])` was a release-mode no-op.
4. **`wrapping_add` → `+= 1`** — both occurrences. Per-encode bumps are bounded by N − 1 < 2³², so overflow is unreachable; `+= 1` correctly communicates that wraparound would be a bug.
5. **Doc comment** — added a paragraph spelling out that heap key `(rank, l, gen_at_push)` carries the byte-identicality contract via leftmost-on-tie tiebreak, matching the `<` semantics of the naive scan and every reference BPE impl.

Re-verified the end-to-end byte-identical claim against HF `tokenizers` on the 16 384-token prompt after the rewrite — still zero diff.

Issue #78, opened 2026-04-28
Port sliding-window FA from Lucebox PR #26 (long-context decode lever, 3.48× projected)


Kaden-Schutt:
**Status**: PRD scoped at `docs/plans/sliding-window-fa.prd`; branch `feat/sliding-window-fa` (commit `6c59950`); implementation queued.

## Background

Lucebox-Org/lucebox-hub merged [PR #26](https://github.com/Luce-Org/lucebox-hub/pull/26) on 2026-04-26 — "feat(dflash): sliding window FA + two-phase cache". On RTX 3090 with Qwen3.6-27B at 60K context, decode jumps from 25 tok/s (full attention) to 89.7 tok/s (window=2048), **3.48× with 100% speculative acceptance** at all window sizes.

The 100% acceptance claim is structural, not probabilistic: when target and draft both apply the same windowed mask, their argmax distributions agree by construction.

## Why this matters

- Empirically validated on the same target/draft class hipfire ships (Qwen3.6-27B + DFlash MQ4 draft).
- Cheaper than CASK / TriAttention at runtime: a single `win_start` shift in the FA mask, no separate scoring kernel pass per cycle.
- Compounds with DFlash spec-decode where verify/replay FA is the long-context hot path.

## Hipfire has no equivalent today

| Mechanism | hipfire status |
|---|---|
| Sliding-window FA (last-N positions) | **NO** |
| Sparse-top-B (CASK / TriAttention) | YES — sophisticated but expensive |
| Rotate-out (KV eviction) | YES — compaction, not windowing |

Sparse-top-B and rotate-out solve different problems; sliding-window is the missing piece for cheap long-context decode.

## Phases (per PRD)

1. **Phase 1** (~3-4 hr) — asym3 decode kernel + plumbing on the default Qwen 3.5/3.6 KV mode. Coherence-gate must pass with env unset (bit-exact baseline preserved); 16K-prompt smoke must produce fluent output with `HIPFIRE_FA_WINDOW=2048`.
2. **Phase 2** (~2-3 hr) — fan out to asym2 / asym4 / q8_0 FA kernel variants.
3. **Phase 3** (~3-4 hr) — verify/replay FA window for DFlash spec-decode (`attention_flash_*_batched` variants). This is where the 3.64× decode win materializes.
4. **Phase 4** (deferred) — graduation: `--fa-window` CLI flag, default-on for 27B registry entries, CHANGELOG + CONFIG.md docs.

[...]

PR #204, opened 2026-05-08
WIP: hetero PFlash+DFlash — PR2 drafter device pinning + PR3 cross-card coord (in progress)

alpineQ:
## Status: closing this draft

Closing without merge after a follow-up perf investigation that turned up a structurally better lever (native MTP) and one optimization worth carving out as a smaller, narrower PR.

## What landed on this branch through PR3 step 2 (already covered in earlier comment)

- `041436b` PR2 drafter device pinning
- `41fa1e2` PR3 step 1 — DFlash allocations routed to `drafter_gpu`
- `1083d48` tiled online-softmax for `attention_dflash_f32` (LDS-overflow fix at L ≥ 16128)
- `c098323` PR3 step 2 — `bind_thread` routing for cross-card spec-decode

Functional milestone (correctness): 27B + DFlash on a single 24 GB consumer card at 16K canonical, coherent prose at τ=1.78, 11.4 tok/s decode.

## What I tried on top after that comment, and what came of it

| Commit | Result |
|---|---|
| `b80e85b` perf(kernel): full-workgroup V-accumulation in `attention_dflash_f32` Phase C | ✅ Clean +33% on `attn_kernel` time at 16K hetero. No τ regression. Worth keeping. |
| ~~`6ba16a2` `HIPFIRE_DFLASH_ATTN_WINDOW` — drafter-side sliding window~~ | ❌ **Reverted via rebase.** Cut drafter context to last W positions, which knocked τ from 1.74 → 1.45. The cycle-wall reduction looked like a tok/s win in isolation but per-token throughput went **below AR-only baseline** (30 ms/token vs AR 25.6). Net negative once you account for the τ loss. Wrong lever. |
| `62890e0` PR3 step 4 — DDTree + cross-card hetero spec-decode | Structural plumbing: drafter_gpu routing through `spec_step_ddtree_batched` / `spec_step_ddtree_path_c`, removes the load-time refusal block. DDTree+hetero now runs cleanly. **Doesn't itself unlock a throughput win** at 16K — peaks at ~25 tok/s on Wikipedia prose, still below AR. |

## The honest perf picture at the same 16K canonical (qwen3.6-27b.mq4, prompt md5 `d21eb…`)

| Setup | decode tok/s | τ | per-token |
|---|---|---|---|
| hipfire AR-only (single card) | **39.0** | n/a | 25.6 ms |
| hipfire DFlash hetero, this branch's best (W=2048, no DDTree) | 20.2 | 1.48 | 49 ms |
| hipfire DFlash hetero, this branch's best (W=96 + DDTree path_c B=4) | 33.0 | 1.45 | 30 ms |
| llama.cpp + Qwen3.6 native MTP, `--spec-draft-n-max=8` | **45.4** | 3.5 | 22 ms |

**Native MTP is the lever the user wants** at 16K Wikipedia prose: it's the only configuration here that actually beats AR (+16%), and it's been right there all along in the source weights as `mtp.*` tensors that the hipfire quantizer currently strips.

I also tried `feat/sliding-window-fa` locally as an alternative (#78 — Lucebox PR #26 port). At 16K, target-side window doesn't deliver the 3.48× the Lucebox bench claims at 60K — got ~37 tok/s at W=512, comparable to plain AR. The Lucebox win materializes when target FA itself is the dominant cost, which kicks in at very long context (60K+), not at 16K.

## Why I'm closing rather than asking for review

The hetero DFlash direction in this branch is a correctness milestone (27B at 16K on a single 24 GB card, no OOM, coherent output) but it doesn't translate to a throughput win against AR on this workload. The structural answer (native MTP integration) is a different code-path discussion than this PR is scoped for, and `b80e85b` deserves to land independently of all the hetero plumbing — it's a kernel improvement that benefits any DFlash setup on gfx1100 once the tiled `attention_dflash_f32` (`1083d48`) is in.

Carving the kernel work out to a narrow PR.

## Reproducibility

All numbers above are 3-run means on the canonical Wikipedia 16K prompt (md5 `d21ebdd2e3204c1788a7f2681d60c38a`), Qwen3.6-27B mq4 + qwen36-27b-dflash-mq4 drafter on 2× RX 7900 XTX (gfx1100), MAX_TOK=128. llama.cpp built from this fork's HIP target with `--spec-type mtp --spec-draft-n-max 8`.

### DIFF
--- a/crates/hipfire-arch-qwen35/src/speculative.rs
+++ b/crates/hipfire-arch-qwen35/src/speculative.rs
@@ -1343,6 +1343,20 @@ impl HiddenStateRingBuffer {
         self.head = 0;
         self.written = 0;
     }
+
+    /// Free all GPU tensors owned by this ring. Caller MUST pass the same
+    /// `Gpu` instance the ring was allocated from — under hetero PFlash+DFlash
+    /// (PR3 step 2) the ring lives on the drafter Gpu, so unload_model must
+    /// route the free through `dflash_drafter_gpu` rather than the target.
+    /// Freeing on the wrong device leaks the allocations.
+    pub fn free_gpu(self, gpu: &mut Gpu) {
+        for t in self.layer_bufs {
+            let _ = gpu.free_tensor(t);
+        }
+        for t in self.staging_bufs {
+            let _ = gpu.free_tensor(t);
+        }
+    }
 }
 
 /// Single-pass argmax for token sampling. Not SIMD-optimized — the logit
@@ -2272,6 +2286,21 @@ pub fn scatter_hidden_block_to_interleaved(
     let row_bytes = hidden * 4;
     let start_slot = (head + max_pos - block_size) % max_pos;
 
+    // PR3 step 2 hetero hardening: at long prompts (≥ ~10K BPE tokens with
+    // num_extract=5 → 50K+ memcpys), the AMD HIP runtime accumulates an
+    // internal `Command` object chain proportional to memcpy count. The
+    // chain releases linked-list-style via `Command::terminate() →
+    // ReferenceCountedObject::release() → Command::releaseResources()
+    // → terminate()` recursion. At ~10K rows × 5 extracts = 50K commands,
+    // the recursive release blows the main thread's 8 MB stack on the
+    // FIRST subsequent kernel launch (in our case spec_step_dflash's
+    // first cycle). Verified via core-dump backtrace: 50K identical
+    // frames in libamdhip64.so.
+    //
+    // Workaround: periodically `device_synchronize()` to drain the
+    // command chain. Fence at every CHAIN_DRAIN_INTERVAL rows × num_extract
+    // memcpys so the chain stays bounded.
+    const CHAIN_DRAIN_INTERVAL: usize = 1024;
     for r in 0..n_rows {
         let slot = (start_slot + r) % max_pos;
         let dst_row = dst_row_offset + r;
@@ -2287,6 +2316,17 @@ pub fn scatter_hidden_block_to_interleaved(
                 row_bytes,
             )?;
         }
+        // Drain the HIP command chain periodically (see comment above the loop).
+        if (r + 1) % CHAIN_DRAIN_INTERVAL == 0 {
+            gpu.hip.device_synchronize()?;
+        }
+    }
+    // Final drain only if any pending memcpys haven't been drained by the
+    // periodic fence above. Skip the unconditional sync on short prompts
+    // (n_rows < CHAIN_DRAIN_INTERVAL) so single-card / short-context cycles
+    // don't pay an extra device_synchronize that adds nothing.
+    if n_rows > 0 && n_rows % CHAIN_DRAIN_INTERVAL != 0 {
+        gpu.hip.device_synchronize()?;
     }
     Ok(())
 }
@@ -2401,7 +2441,8 @@ pub fn download_hidden_block(
 /// full cumulative context (the default, distribution-preserving path).
 #[allow(clippy::too_many_arguments)]
 pub fn spec_step_dflash(
-    gpu: &mut Gpu,
+    target_gpu: &mut Gpu,
+    mut drafter_gpu: Option<&mut Gpu>,
     target: &mut ModelSlot,
     draft_weights: &DflashWeights,
     draft_cfg: &DflashConfig,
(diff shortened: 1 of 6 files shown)


alpineQ:
## Status update — PR3 step 2 landed (correctness milestone)

Force-pushed the branch with three new commits on top of the original PR2:

```
c098323  feat(dflash): PR3 step 2 — bind_thread routing for cross-card spec-decode
1083d48  feat(kernel): tiled online-softmax for attention_dflash_f32
41fa1e2  wip(dflash): PR3 step 1 — route DFlash allocations to drafter_gpu
041436b  feat(daemon): DFlash drafter device pinning (PR2)         ← rebased
```

(History was rebased onto current `upstream/master` to drop the local merge of `fix/bpe-quadratic-encoding` since #201 has landed. PR2's commit hash changed; content is identical.)

## Where we ended up on the three architecture questions

1. **`HiddenStateRingBuffer` ownership → drafter Gpu.** Drafter reads it many times per cycle (cross-attention K/V); target writes it once per cycle (verify forward). Putting the ring on drafter and using a target→drafter peer write at hidden capture time amortizes correctly. PR3 step 1 implemented this; step 2 confirms it works at full 16K.

2. **Refusal contract → same-arch only for now.** Step 2 was validated on homogeneous 2× 7900 XTX (gfx1100). Cross-arch (gfx1010 drafter + gfx1100 target) requires kernel-dispatch routing through `drafter_gpu.arch` rather than `target.arch`, which I haven't touched. Recommended follow-on PR if you want it for the `hipx` rig — happy to take that on.

3. **Coherence-gate-dflash regressions on hetero** — observed and addressed:
   - `gpu_canonical_16K` produces τ=0 with a naive port (LDS overflow in `attention_dflash_f32`'s `scores[L]` allocation; gfx1100's 64 KB LDS budget overflows at L ≥ 16128 → `hipModuleLaunchKernel: invalid argument`). Fixed in `1083d48` by rewriting the kernel as tiled online-softmax (FlashAttention-style); single-tile case is bit-comparable to the prior single-pass kernel.
   - HIP runtime `Command::terminate()` recursion stack overflow at 50K+ memcpys in `scatter_hidden_block_to_interleaved` (verified via core-dump backtrace: 50K identical frames in libamdhip64.so). Fixed by periodic `device_synchronize()` every 1024 rows to drain the chain.
   - One step-1 era peer-access bug (target → drafter peer never enabled because `enable_peer_access` was called with the wrong device bound). Fixed and surfaced via `eprintln!` instead of swallowed.

## Validation results

| | single-card | hetero |
|---|---|---|
| 1K prompt @ max_seq=4096 | τ=1.14 cycles=7 decode_tok_s=36.2 | τ=1.14 cycles=7 decode_tok_s=32.1 |
| 4K prompt | (synthetic, OOM at 17K max_seq) | τ=10 cycles=2 |
| 12K prompt | OOM (`hipMalloc`) | τ=0.07 (synthetic prose attractor) |
| 16K canonical Wikipedia prose | OOM | **τ=1.78 cycles=46 decode_tok_s=11.4 — coherent output** |
| τ-sanity 980 BPE × 3 | τ=2.57 ± 0.10 | τ=2.44 ± 0.16 |

The 16K canonical bench moves from "doesn't run on a single card" / "produces τ=0 garbage with naive port" to **128 tokens of coherent prose at τ=1.78**.

τ-sanity at 980 BPE shows hetero τ is within run-to-run noise of single-card (~5 % gap, 1 stddev). No algorithmic regression. The throughput delta (-14 % on this short prompt) is purely cross-card overhead (~10 % per-cycle PCIe + bind_thread).

Single-card 1K bit-exact non-regression — `split_mode = drafter_gpu.is_some()` guards every cross-card hotspot, so the prior path is byte-identical when the env var isn't set.

## Honest assessment of where this leaves us

**This stack is a correctness milestone, not yet a throughput win.** What it unlocks:

✅ 27B + DFlash on a single 24 GB consumer card at 16K context — previously impossible (drafter scratch + target weights overflow VRAM).
✅ Coherent generation at 16K hetero (vs τ=0 / SIGSEGV with naive port).
✅ No single-card regression.

What it does NOT yet deliver:

❌ PRD acceptance criterion "hetero ≥ 1.25× solo DFlash on at least one config (sequential, no pipelining)". Hetero at 16K runs at 11.4 tok/s decode; single-card AR at 16K (where it fits) runs ~37 tok/s per the `rdna1-perf-research` baseline. The cross-card per-cycle latency dominates at long context.

The path to throughput parity / win:
1. **PR 5 (cycle pipelining)** — overlap drafter cycle N+1 with verify cycle N. The PRD pre-registers a 1.7× sequential bump from this. This is the load-bearing remaining work.
2. **V-read coalescing** in the new tiled attention kernel — the multi-tile case currently does 2× global V-reads per tile; moving V into shared blocks would drop that. Worth ~10-20 % at long context per back-of-envelope.
3. **Async streams** for cross-card transfers instead of the current sync `device_synchronize` barriers between drafter / target sections. This needs the async-handler-side ordering primitives — bigger change but where the real wins live.

I'm planning to keep digging on #1 and #3 next. Will open separate PRs (or push to this branch as additional commits, depending on your preference) once functional. Happy to coordinate if you have priorities or want to land this first as the correctness milestone before piling on perf work.

## What I'd like input on (revised)

- **OK to land this branch as the "correctness step" once you've reviewed?** It doesn't meet the PRD's perf criterion but it's a clean dependency for PR5.
- **PR4 (PFlash + DFlash composition)** — should I pick that up now, or do PR5 first since it has the perf payoff?
- **`hipx` rig validation** — you mentioned the rig was the primary target. If you have a way for me to run a smoke on it (or push a known-good fixture), I can validate the cross-arch path before opening that follow-up PR. Otherwise I'll ship same-arch only with a clean refusal for cross-arch.

Anything in the diff that needs to change before this is mergeable, let me know.

PR #232, opened 2026-05-11
AR + DFlash think-policy: GPU kernel + state machine + repeat penalty wiring (continuation of #221)

DrBearJew:
DFlash path (generate_dflash):
- GPU kernel: repeat_penalty_argmax_batched.hip — combined repeat_penalty + ngram_block + ban_token + argmax in one HIP launch
- Think-reopen ban: one pre-answer think block allowed, subsequent <think> opens banned at kernel level
- LoopGuard n-gram detector in DFlash decode loop
- Repeat penalty wiring through generate_dflash and spec_step_dflash
- Budget-aware think cap in CLI run path
- ban_token_id passed through spec_step_dflash for structural bans

AR path (generate):
- ThinkState struct + merge_think_bans() + update_think_state() in sampler.rs
- State machine wired at all 3 AR sample sites
- Prefix injection: <think> token injected before prefill when thinking=ON
- Prefix scan: detects template-injected <think>, seeds state

Think IDs parameterized via tokenizer.special_token_id() — no hardcoded magic numbers. MQ4 models lacking think tokens no-op automatically.

Files:
- kernels/src/repeat_penalty_argmax_batched.hip (new, 170 lines)
- cli/index.ts, speculative.rs, daemon.rs, sampler.rs, llama.rs
- docs/ar-think-policy-bug-report.md, docs/mq4-think-token-report.md

Env vars: HIPFIRE_NGRAM_LOOP_THRESHOLD=4, HIPFIRE_DFLASH_NGRAM_BLOCK=1, HIPFIRE_GRAPH=0

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,7 @@
 # JIT-compiled kernel cache
 kernels/compiled/
 .hipfire_kernels/
+.hipfire_kernels
 
 # Agent / tool state
 .serena/
@@ -90,3 +91,5 @@ research/
 .codeinsight+research/
 
 bench/coherence-results/
+
+PR_DESCRIPTION.md
--- a/cli/index.ts
+++ b/cli/index.ts
@@ -1261,6 +1261,11 @@ async function run(model: string, prompt: string, image?: string, temp = 0.3, ma
     genMsg.assistant_prefix = "closed_think";
   } else if (modelCfg.max_think_tokens > 0) {
     genMsg.max_think_tokens = modelCfg.max_think_tokens;
+    // When thinking budget is hit, daemon injects </think> so generation
+    // can continue into visible answer instead of stopping inside an
+    // unclosed <think> block (which leaves empty visible output after strip).
+    genMsg.budget_alert_at_tok = modelCfg.max_think_tokens;
+    genMsg.budget_alert_text = "</think>\n\n";
   }
   if (image) {
     genMsg.image = resolve(image);
@@ -1676,6 +1681,33 @@ async function serve(port: number) {
           if (reasoningEffort === 0) delete genParams.max_think_tokens;
           else genParams.max_think_tokens = reasoningEffort;
         }
+        // Budget-aware think cap.
+        //
+        // The daemon can only force-close <think> if max_think_tokens fires
+        // before request max_tokens is exhausted. Without this clamp:
+        //   max_tokens=200, max_think_tokens=256
+        // means the model can spend the whole request budget inside <think>,
+        // after which the OpenAI wrapper strips the unclosed block and
+        // returns content:"".
+        const minVisibleAfterThink = Math.max(
+          1,
+          Number(process.env.HIPFIRE_MIN_VISIBLE_AFTER_THINK ?? 32) || 32,
+        );
+        const closeThinkOverheadGuess = 8;
+        const thinkBudgetCap = Math.max(
+          1,
+          requestMaxTokens - minVisibleAfterThink - closeThinkOverheadGuess,
+        );
+        if (enableThinking !== false && effective.thinking !== "off" && !preserveThinking) {
+          if (genParams.max_think_tokens == null) {
+            genParams.max_think_tokens = thinkBudgetCap;
+          } else if (genParams.max_think_tokens > 1) {
+            genParams.max_think_tokens = Math.min(
+              genParams.max_think_tokens,
+              thinkBudgetCap,
+            );
+          }
+        }
         // Wire thinking control for both legacy assistant_prefix
         // (ChatFrame::ClosedThink) and the new Jinja template path.
         // The Jinja path uses max_think_tokens==1 as the signal for
@@ -2044,7 +2076,21 @@ async function serve(port: number) {
 
         // Check for tool calls in response
         const parsed = parseToolCalls(content);
-        const choice: any = { index: 0, finish_reason: parsed.tool_calls ? "tool_calls" : "stop" };
+
+        // Detect max_tokens exhaustion inside unclosed think block
+        const openThink = strippedContent.lastIndexOf("<think>");
+        const closeThink = strippedContent.lastIndexOf("</think>");
+        const endedInsideThink = openThink >= 0 && openThink > closeThink;
+        const exhaustedBudget = completionTokens >= requestMaxTokens;
+
+        const choice: any = {
+          index: 0,
+          finish_reason: parsed.tool_calls
+            ? "tool_calls"
+            : exhaustedBudget && endedInsideThink
+              ? "length"
+              : "stop",
+        };
         if (parsed.tool_calls) {
           choice.message = { role: "assistant", content: parsed.content, tool_calls: parsed.tool_calls };
         } else {
--- a/crates/hipfire-arch-qwen35/src/speculative.rs
+++ b/crates/hipfire-arch-qwen35/src/speculative.rs
@@ -2423,6 +2423,7 @@ pub fn spec_step_dflash(
     pld_spine: Option<&[u32]>,
     repeat_penalty: f32,
     repeat_window: usize,
+    ban_token_id: Option<u32>,
 ) -> HipResult<SpecStepResult> {
     // Effective block size for THIS step. Usually `draft_cfg.block_size`
     // (what the draft was trained at, 16 for Qwen3.5-*-DFlash) but a caller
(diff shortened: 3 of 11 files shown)


Kaden-Schutt:
Hey @DrBearJew — thanks for this. The kernel + state-machine infrastructure is solid and the upstream bug report on the MQ4 vocab gap is exactly the kind of root-cause forensics we want in-tree. Three cleanup items before we land it:

## 1. Drop `.hipfire_kernels` symlink from commit

`f0fe2b96` adds a 16-byte symlink at repo root pointing to `kernels/compiled`. This is a local-dev convenience that shouldn't be in tree:

```
lrwxrwxrwx .hipfire_kernels -> kernels/compiled
```

`.gitignore:12` already excludes `.hipfire_kernels/` (directory form) but not the symlink form without trailing slash. Easiest fix: `git rm --cached .hipfire_kernels` and commit, then add `.hipfire_kernels` (no slash) to `.gitignore`.

## 2. Drop `PR_DESCRIPTION.md` from commit

The PR description file at repo root looks like it was meant for `gh pr edit --body-file` and got committed by accident. `git rm PR_DESCRIPTION.md && git commit`.

## 3. Parameterize the hardcoded think IDs at call sites

The `sampler.rs` infrastructure is parameterized correctly — `think_start_id: u32, think_end_id: u32` are struct fields and function parameters. The hardcoding is at the daemon.rs and speculative.rs call sites:

```rust
let think_start_id: u32 = 151657;  // <think>
let think_end_id: u32 = 151658;    // </think>
```

Two options:

**Option A (preferred):** Look up at runtime from the tokenizer:
```rust
let (think_start_id, think_end_id) = match (
    tokenizer.token_to_id("<think>"),
    tokenizer.token_to_id("</think>"),
) {
    (Some(a), Some(b)) => (a, b),
    _ => {
        // Model lacks think tokens (e.g. MQ4 vocab-gap models, non-Qwen
        // families). ThinkState path no-ops; min-think enforcement disabled.
        // Two-pass app-level reasoning is the fallback.
        return /* path that doesn't use ThinkState */;
    }
};
```

**Option B:** Hardcode at the daemon-entry level only (not in `sampler.rs` API surface), and gate the entire ThinkState path on `model_family == "qwen-3.6"`.

Either way, this PR shouldn't introduce Qwen-3.6-specific magic numbers anywhere in shared `hipfire-runtime` or `hipfire-arch-qwen35` code paths. With the existing parameterization in sampler.rs, this is just a few-line change at the call sites.

## What stays

- `kernels/src/repeat_penalty_argmax_batched.hip` (170 LOC) — independent useful infra
- `ThinkState` struct + `merge_think_bans` + `update_think_state` — parameterized correctly
- DFlash repeat-penalty + LoopGuard wiring in `speculative.rs`
- AR state-machine wiring at the 3 sample sites
- Both narrative docs — preserve the research record

Once those three are addressed, I think this is mergeable. Build is clean on this end (post #233/#235/#231/#229/#227 master, `2ea7acd8`).

unverbraucht:
Tested this PR on top of our F2 AWQ work (worktree cherry-pick of `6e2a74ec`, only a trivial `.gitignore` conflict). Wanted to share findings.

**Setup:** gfx906, Qwen3.5-9B MQ4 + AWQ, driven via `coherence_probe` → daemon (your AR-path ThinkState wiring active, saw the `[sampler] forcing think open: banning </think>` log lines firing).

**Background.** Our 9B has a longstanding reasoning-failure mode on a simple word problem ("trains leave A and B, when do they meet"): model opens cleanly with the right algebra, then derails into "let me re-read the prompt, I think numbers are missing" loops, never producing an answer. Reproduces on mq4/mq6/mq8 quants from May 6 onward, all temperatures, AWQ on/off. Llama.cpp BF16 on the same GPU answers correctly ("11:34am"), so the bug is hipfire-side. Looked like the same failure class as #258's "Plain ChatFrame, thinking enabled" row.

**What this PR fixed (real improvement):**

- No more "user's input is missing numbers" hallucination — model cleanly states `v_A = 60 mph`, `280 miles`, `Train B leaves at 10 AM`
- Stays inside `<think>` as designed instead of rambling at the user level
- Sets up the relative-velocity framing correctly
- No `eos_immediate` / `whitespace_only` / `attractor` detectors fire

**Where it didn't go far enough:**

The min-64-token think enforcement (designed for the MQ4 Qwen3.6 vocab gap per `mq4-think-token-report.md`) has a side effect on this prompt class: once inside `<think>`, the model gets stuck in an *interpretation loop* — "Are trains on the same track? Does 'leaves Station A' imply heading toward B? Scenario A vs B vs..." Burned through 4000 tokens, all inside `<think>`, never closes the block, never produces a visible answer. Math is correctly framed throughout; the model just won't commit.

So PR #232 traded "fluent re-parsing nonsense outside `<think>`" (catastrophic) for "endless hedging inside `<think>`" (less obvious but still a 0-token-answer failure). The detectors pass because the symptoms moved *into* the think block.

**Possible refinements** if any resonate with your design:

1. **Force-close `</think>` after a hard token budget** (e.g., 300–500 tokens) regardless of the min-64 floor. Creates a clean `[min, max]` think window. Currently the ban-`</think>` rule has only a lower bound.
2. **Extend `LoopGuard` n-gram detection inside `<think>`** — the "let me reconsider" / "let's check" / "wait" repetition during hedging is detectable.
3. **Apply ThinkState to `run.rs` and `speculative.rs` too**, not just `daemon.rs`. `run.rs` uses `llama::sample_top_p` directly and bypasses `merge_think_bans` entirely, so `cargo run --example run` and any DFlash spec-decode path skip the AR-side policy.

Happy to send a follow-up PR with #3 — smallest change, unblocks regression-test parity across AR-path entry points.

**Issue #258 connection:** Our 9B trains-meet looks like the same bug class as your 0.8B Federalist "Plain ChatFrame, thinking enabled" failure, scaled up. Worth checking whether the post-#232 0.8B essay has the same "inside-`<think>` stalling" residue.

Thanks for the work — the ThinkState scaffolding is the right shape to build on.

unverbraucht:
## Found a real bug in this PR's `<think>` injection — missing `\n`

While investigating the 9B chat-mode reasoning loop (issue #258), I traced one component of it to the `<think>` injection added in this PR. The injection appends a bare `<think>` token, but the official Qwen3.5 chat template emits `<think>\n` (with trailing newline). Without the `\n`, the model's first-decode top-1 logit becomes `'\n'` (id 198) — the model is trying to emit the missing newline — instead of a content token.

### Evidence

HF transformers BF16 on the same prompt, same 9B Qwen3.5, last position top-1:

| Last 2 prompt tokens | Top-1 prediction | Logit |
|----------------------|------------------|------:|
| `..., <think>` (bare, what this PR emits) | `'\n'` (198) | 21.625 |
| `..., <think>, \n` (HF apply_chat_template canonical) | `'Here'` (8160) | 28.625 |

Hipfire trajectory dump vs llama.cpp golden on Qwen3.5-9B Q8 (Q8 weights + Q8 KV) at the first decode position:

| | Pos-0 KLD(L‖H) | Hipfire argmax | llama.cpp top-1 rank in hipfire |
|--|---:|--|---:|
| Before fix (this PR's current state) | 26.66 | `'\n'` (198) | outside top-20 |
| After fix (append `\n`) | **5.04** | `'Here'` (8160) | **#2** |

End-to-end Q8 + Q8 KV + Jinja + temp=0.3 on the canonical reasoning prompt:
- Before fix: `\n\n<|im_end|>` (2 tokens; the model refuses to think with the malformed prefix and tries to exit).
- After fix: `Thinking Process:\n\n1.  **Analyze the Request:**\n    *   Event: A train leaves Station A. ...` — coherent structured chain-of-thought response.

### Fix

Three injection sites needed the same fix. I added a single helper in `prompt_frame.rs` so all callers go through one canonical path. Suggested patch (against this PR's branch):

**1. Add the helper to `crates/hipfire-runtime/src/prompt_frame.rs`** (just after the `ChatFrame` struct definition, before `impl ChatFrame`):

```rust
/// Canonical "open think" token sequence: `<think>` + `\n`.
///
/// Matches what `AssistantPrefix::OpenThink` appends to a Plain frame and
/// what the official Qwen3.5 chat template emits after `<|im_start|>assistant\n`.
///
/// Use this from callers that build a Plain-prefixed `ChatFrame` and then
/// need to manually open thinking after the fact (daemon's per-request
/// `<think>` injection, diagnostic dump tools). Appending only `<think>`
/// without the trailing `\n` is a footgun — the model's first-decode top-1
/// becomes `\n` (trying to emit the missing newline) which derails chat-
/// mode reasoning. See issue #258 for the diagnosis.
///
/// Returns an empty `Vec` if the tokenizer does not register `<think>` as
/// a special token (non-Qwen3+ families).
pub fn open_think_tokens(tokenizer: &Tokenizer) -> Vec<u32> {
    let Some(think_id) = tokenizer.special_token_id("<think>") else {
        return Vec::new();
    };
    let nl = tokenizer.encode("\n");
    let mut out = Vec::with_capacity(1 + nl.len());
    out.push(think_id);
    out.extend_from_slice(&nl);
    out
}
```

**2. Refactor `crates/hipfire-runtime/examples/daemon.rs:~3215` (the injection added by this PR):**

Replace:
```rust
if thinking_allowed && max_think_tokens != 1 && m.seq_pos == 0 {
    new_tokens.push(think_start_id);
    eprintln!("[daemon] think-prefix injected: appended token {} (<think>)", think_start_id);
}
```

With (handles the Jinja path's possible existing `<think>` to avoid double-injection):
```rust
let tokenizer_ref = m.tokenizer.as_ref().unwrap();
let already_has_think = !new_tokens.is_empty()
    && thinking_allowed
    && (new_tokens.last() == Some(&think_start_id)
        || (new_tokens.len() >= 2
            && new_tokens[new_tokens.len() - 2] == think_start_id));
if thinking_allowed && max_think_tokens != 1 && m.seq_pos == 0 && !already_has_think {
    let ot = hipfire_runtime::prompt_frame::open_think_tokens(tokenizer_ref);
    new_tokens.extend_from_slice(&ot);
    eprintln!("[daemon] think-prefix injected: appended <think>\\n (id={} + nl)", think_start_id);
} else if already_has_think && new_tokens.last() == Some(&think_start_id) {
    // Jinja path: rendered template ended with `<think>` but no `\n`.
    // Qwen3.5's current template emits both, but variants exist.
    new_tokens.extend_from_slice(&tokenizer_ref.encode("\n"));
    eprintln!("[daemon] jinja template missing trailing \\n after <think>; appended");
}
```

The Jinja branch matters: with `HIPFIRE_JINJA_CHAT=1`, the rendered template already includes `<think>\n`, and the original PR code would push a second `<think>` on top — `[..., <think>, \n, <think>]` — which is also malformed. The detection skips redundant injection in that case.

**3. Any diagnostic dump tools that manually inject `<think>` after `ChatFrame::Plain.build()` need the same `\n`.** In our worktree two such tools existed (`dump_logits_trajectory.rs:~111` and `dump_qwen35_hidden_states_prompt.rs:~121`); both became:
```rust
let ot = hipfire_runtime::prompt_frame::open_think_tokens(&tokenizer);
prompt_tokens.extend_from_slice(&ot);
```

Master doesn't have these tools so they aren't shipping; just noting for future contributors.

### Note: `AssistantPrefix::OpenThink` was already correct

The bug is specifically in **manual `<think>` push after `ChatFrame { AssistantPrefix::Plain }.build()`**. `ChatFrame { AssistantPrefix::OpenThink }.build()` already emits `<think>\n` correctly (`prompt_frame.rs:272-274`), and the `open_think_appends_think_newline_when_special_present` unit test verifies it. The new `open_think_tokens()` helper just exposes that same correct sequence to callers who need to inject after-the-fact rather than build a fresh frame.

### Scope of what this fix solves vs leaves open

This fix resolves the **catastrophic first-decode** divergence — empty-think exits, `\n` as first decoded token, immediate `<|im_end|>`. After the fix the model produces coherent "Thinking Process: 1. Analyze the Request: ..." structured output.

It does NOT fully resolve issue #258. Master (without this PR) does NOT have the `<think>\n` bug, yet **master also exhibits a mid-trace reasoning loop** on multi-step math word problems with Q8 + Q8 KV: the model starts thinking coherently, gets a few steps in, then derails into a hallucinated-numbers self-doubt loop ("Wait, did you omit a digit? Let me assume 28 instead of 280..."). llama.cpp BF16 on the same model doesn't loop. So there's a separate bug B that emerges over many autoregressive decode steps — investigating that next on the worktree. Will follow up on issue #258 once we have more data.

Full investigation notes at `docs/investigations/2026-05-15-9b-reasoning-loop/` on the `feat/mq-v2-quant-format` branch (Phase 6.5 section covers this fix; earlier phases document the cos-direction-diff dead-end that initially obscured it).

Kaden-Schutt:
Archiving this PR as superseded by newer work and closing it out of the active queue.

What to preserve from this PR:

- the GPU/state-machine exploration around think policy and DFlash sampling
- the upstream observation that structural think-token handling matters for Qwen-family behavior
- the bug trail in the comments, especially the missing-newline `<think>` injection finding

Why closing now:

- the branch is conflicting against current `master`
- the PR is broad enough that it is no longer a clean merge candidate
- recent narrower work covers the live issue path more directly, especially #267 for the repeat-penalty default/root-cause thread
- any remaining useful pieces should come back as small, current-`master` PRs with isolated validation

Thanks @DrBearJew — this was useful investigation surface, but it should live as archived context rather than an open merge target.

PR #235, opened 2026-05-11
MFP4G32 v2 #3-5: hipGraph default-on for gfx11+gfx12 decode (+0.7-2.6%)

Kaden-Schutt:
## TL;DR

Three production wins shipped default-on, three negative-result research artifacts shipped opt-in default-off, and two silent-correctness/perf-infra bug fixes. All on the same `feat/MFP4G32-v2-batched-wmma` branch as the v2 #2 batched-WMMA prefill work (`ada4eddf` + `9145f06a`).

| What | Where | Result |
|---|---|---|
| **HIPFIRE_GRAPH default-on for gfx12** | `7fca62df` | **+2.4-2.7%** decode on 9B Qwen 3.5 MFP4G32 (R9700) |
| **HIPFIRE_GRAPH default-on for gfx11** | `9582ac39` | **+0.7-1.55%** decode on 9B + 0.8B HFP4G32 (7900 XTX) |
| **`invalidate_x_caches_for` rotation cache fix** | `7fca62df` | **silent correctness bug fixed** — `ensure_*_x` was returning stale data on every `weight_gemv(MFP4G32)` call after the first within a forward pass |
| OnceLock env-gate cache (`is_fp8_wmma_enabled`) | `7fca62df` | ~4 µs/call overhead removed; was masking perf-bench signal |
| FP8 prefill WMMA + decode dot4 GEMV infra | `7fca62df` | opt-in research, default-off (production decode 0.93-0.97×) |
| gfx11 fdot2 GEMV (v_dot2_f32_f16) | `2a71f91f` | opt-in research, default-off (synthetic 2.08×, production 0.90×) |

### β: per-forward hipGraph capture/replay on gfx11+gfx12

Capturing each decode token's forward pass as a hipGraph and replaying it amortizes per-launch overhead across all the kernels in the forward chain. Measured on production decode workloads:

| Arch | Model | Mean speedup | Variance |
|---|---|---|---|
| gfx1201 (R9700)        | 9B Qwen 3.5 MFP4G32 | **+2.6%**  | 1.023-1.029× (5 runs) |
| gfx1100 (7900 XTX)     | 9B Qwen 3.5 HFP4G32 | **+1.55%** | 1.007-1.019× (5 runs) |
| gfx1100 (7900 XTX)     | 0.8B HFP4G32        | **+0.6%**  | 1.001-1.009× (5 runs) |

Default policy in `qwen35.rs` forward():
- **gfx12 + gfx11**: default-ON
- **other archs (RDNA1/2, CDNA)**: default-OFF (opt-in via `HIPFIRE_GRAPH=1`)
- **MoE configs**: always direct unless `HIPFIRE_GRAPH_MOE=1` (known numerical drift)
- `HIPFIRE_GRAPH=0` always wins as kill switch

### Silent correctness fix: `invalidate_x_caches_for` on rotations

`ensure_fp16_x` / `ensure_fp8_x` cache the FP16/FP8 conversion result by source pointer (`fp16_x_source_ptr` / `fp8_x_source_ptr`). On the MFP4G32 path, `gemv_mfp4g32_with_rotate` writes rotated activations into the persistent `mq_x_rot` scratch buffer — **the pointer is stable across calls but the data turns over per rotation**. The cache silently returned stale FP16/FP8 from the previous gemv on every subsequent weight_gemv call within a forward pass.

Coherence detectors missed it because the output stayed vaguely on-topic, but the activations being passed to gemv kernels were silently wrong post-call-1 in every MFP4G32 forward. Fix adds `invalidate_x_caches_for(dst_ptr)` to all rotation entry points: `rotate_x_mq` / `_batched`, `fused_rmsnorm_rotate_mq` / `_batched`, `fused_silu_mul_rotate_mq` / `_batched`, and the new `mq_rotate_x_dual_fp8`. Memory checkpoint `ada4eddf` had flagged this caveat as "latent in HFQ4 test too" — now fixed for both fp8 and fp16 caches.

### OnceLock env-gate

`is_fp8_wmma_enabled()` (and now `is_dot2_gemv_enabled()`) cache the env var lookup in `std::sync::OnceLock`. Reading via `std::env::var` on every dispatch was a ~4 µs/call overhead in tight bench loops (syscall + String alloc), which was masking the synthetic FP8 win and distorting all our A/B perf measurements. Cleanup that always applies.

## What ships opt-in default-off (research scaffold)

Three independent hypotheses tested and falsified in production. Code is correct, well-instrumented, and saved for future researchers — the bench tools especially are reusable.

### Hypothesis: FP8 WMMA prefill on gfx12

`gemm_qkv_hfp4g32_wmma_fp8.gfx12.hip` — uses `wmma_f32_16x16x16_fp8_fp8_w32_gfx12`. Microbench measured **1.87× raw WMMA throughput** vs FP16. End-to-end QKV prefill measured **0.71-1.10×** (loss at small N, ~at-parity at N=2048).

Root cause: HFP4G32's per-32-block UE8M0 scale forces post-WMMA `acc[j] += partial[j] * row_scale[j] * UE8M0[j]` chains on F32 accumulator. The FP16 sibling sidesteps this by baking scale into the half operand pre-WMMA. The required cross-lane shuffle + FMA chain costs roughly what the WMMA-throughput advantage delivers.

Gated by `HIPFIRE_FP8_WMMA=1` + `FP8_WMMA_MIN_BATCH=1024` so it never triggers at decode (N=1 measures 0.71×).

### Hypothesis: FP8 dot4 decode GEMV on gfx12

`gemv_hfp4g32_fp8.gfx12.hip` — uses `dot4_f32_fp8_fp8`. Synthetic bench measured **1.04-1.42× wins** on FFN shapes (gate_up M=11008 hits 99% peak BW vs 78% fallback). Tried 4 variants:

| Variant | Production decode |
|---|---|
| v1 (ensure_fp8_x pre-pass) | -7% |
| v2 (inline F32→FP8 cvt) | -7% |
| v3 / A3 (fused mq_rotate_x_dual_fp8) | -6% |
| v3 + FP8_GEMV_MIN_M shape-gate | -6% |

rocprof attribution: per-call FP8 GEMV is ~3 µs slower than fallback in production despite winning by ~4 µs in synthetic isolation. The ~30 µs gap between synthetic and production per call is cross-kernel context cost (L2 thrashing from interleaved attention/conv1d/norm ops between gemvs, scheduler interference). FP8 path pays MORE of this than fallback.

### Hypothesis: gfx11 v_dot2_f32_f16 trickle-down

`gemv_hfp4g32_dot2.gfx11.hip` — uses `__builtin_amdgcn_fdot2` (one v_dot2_f32_f16 = 2 FP16 muls + 1 FP32 add). Synthetic bench measured **1.13-2.08× wins** on 7900 XTX (gate_up hit 84% peak BW vs 40% fallback). Tested two kernel variants (single-carry chain + 4 parallel partial accumulators).

Production decode on 9B HFP4G32: **0.90× consistently across both variants and 5 paired runs each**. Same trap as gfx12 FP8 — synthetic wins evaporate in production workload context.

NRMSE 2e-4 (FP16 multiply intermediate; much more precise than FP8's 2.7%). Correctness intact, perf empirically falsified.

## What this branch empirically established

Across three independent hypothesis lines on two arches:

| Lever class | Verdict on this codebase + ROCm 7.2 |
|---|---|
| **Launch-reduction (β / hipGraph capture)** | **alive, shipped default-on for gfx11+gfx12** |
| Per-kernel ALU optimization (FP8 dot4, FP8 WMMA, fdot2) | dead in production despite synthetic wins |
| Activation pre-pass kernels with src_ptr caching | trap unless paired with rotation invalidation (now fixed) |
| Speculative-decode cross-stream pipelining | already tested in PR5 D0-D3b: -1.7% / -7% |

**The single positive lever** that crosses zero in production on this codebase is launch-overhead amortization. Three independent kernel-level ALU experiments all lost in production despite synthetic wins — that's not bad luck, it's a structural property of this codebase + ROCm 7.2 + decode workload pattern (cross-kernel L2/scheduler context costs dominate per-call kernel time).

## Files touched

**Kernels added (opt-in research):**
- `kernels/src/gemm_qkv_hfp4g32_wmma_fp8.gfx12.hip` — prefill FP8 WMMA QKV
- `kernels/src/pack_f32_to_fp8.gfx12.hip` — F32→E4M3 pre-pass
- `kernels/src/gemv_hfp4g32_fp8.gfx12.hip` — decode FP8 dot4 GEMV
- `kernels/src/mq_rotate_x_dual.gfx12.hip` — fused FWHT + FP8 pack
- `kernels/src/gemv_hfp4g32_dot2.gfx11.hip` — gfx11 fdot2 GEMV

**Rust wiring:**
- `crates/rdna-compute/src/kernels.rs` — 5 new source-include consts
- `crates/rdna-compute/src/dispatch.rs` — ensure_fp8_x, rotate_x_mq_dual_fp8, FP8 + dot2 dispatch wrappers, FP8_WMMA_MIN_BATCH + FP8_GEMV_MIN_M empirical thresholds, OnceLock env-gates, invalidate_x_caches_for and its call sites in all rotation entry points
- `crates/hipfire-arch-qwen35/src/qwen35.rs` — HIPFIRE_GRAPH default-on policy for gfx11+gfx12

**Bench + test tooling:**
- `crates/rdna-compute/examples/test_gemm_hfp4g32_fp8.rs`
- `crates/rdna-compute/examples/test_gemv_hfp4g32_fp8.rs`
- `crates/rdna-compute/examples/test_gemv_hfp4g32_dot2.rs`
- `crates/rdna-compute/examples/bench_gemv_hfp4g32_bw.rs`
- `crates/rdna-compute/examples/bench_gemv_mfp4g32_rotate.rs`

## Memory references (where the detailed empirical data lives)

- `memory/project_hypothesis_beta_graph_wins_2026_05_11.md` — full β data tables, rocprof breakdown, FP8 + β conflict analysis
- `memory/project_gfx11_dot2_trickle_down_falsified_2026_05_11.md` — fdot2 synthetic vs production analysis
- `memory/project_fp8_wmma_hfp4g32_2026_05_10.md` — earlier FP8 WMMA prefill analysis

## Test plan

- [x] **Correctness**: all opt-in research kernels pass `cmp_tol` correctness tests against F32 reference (NRMSE 2.7% for FP8 paths, 2e-4 for dot2 path)
- [x] **β + cleanup fixes**: 9B Qwen 3.5 MFP4G32 on hiptrx gfx1201, 5-run paired (FP8=0/GRAPH=0 vs FP8=0/GRAPH=1) — all coherence detectors PASS, decode +2.4-2.7%
- [x] **β default-on gfx12**: 9B Qwen 3.5 MFP4G32, 5-run paired (HIPFIRE_GRAPH=0 vs no-env) — all PASS, decode +2.4-2.7%
- [x] **β default-on gfx11**: 9B + 0.8B HFP4G32 on local 7900 XTX, 5-run paired each — all PASS, decode +0.6-1.55%
- [x] **rocprof attribution**: gfx12 FP8 ON vs OFF, kernel-level stats CSV produced, gap localized to per-FP8-call context cost (+3 µs vs fallback in production)
- [x] **Opt-in FP8 + dot2 paths default to OFF**: verified via dispatch routing in `gemv_hfp4g32` / `gemm_qkv_hfp4g32` / `gemv_mfp4g32_with_rotate`
- [ ] **Multi-arch sweep**: gfx1010 / gfx1030 / gfx906 still default-off for β; need bench before flipping their defaults. **Not in this PR.**
- [ ] **MoE archs (Qwen 35B-A3B)**: β remains default-off for MoE; the known DeltaNet state-drift bug is unfixed. **Not in this PR.**

## Why this round didn't ship a bigger win

The branch's original v2 #2 scope (batched WMMA prefill) shipped at `ada4eddf` + `9145f06a`. This PR is the v2 #3-5 follow-on after extensive empirical exploration. The honest summary: synthetic bench wins are not predictive of production decode wins on this codebase + ROCm 7.2. The +2.4-2.7% gfx12 / +0.7-1.55% gfx11 β win is what the empirical method actually validated.

Three other levers (FP8 prefill, FP8 decode dot4, gfx11 fdot2 trickle-down) showed clean synthetic wins of 1.04-2.08× but lost in production. Those are kept as research scaffold so the next researcher has the bench + rocprof methodology in place; rebuilding it would take a full day.

### DIFF
--- a/crates/hipfire-arch-qwen35/src/qwen35.rs
+++ b/crates/hipfire-arch-qwen35/src/qwen35.rs
@@ -3808,6 +3808,16 @@ fn is_batchable_la(dt: DType, arch: &str) -> bool {
             | "gfx1200" | "gfx1201"
         );
 
+    // HFP4G32 / MFP4G32 (v2 #2 batched WMMA prefill): same arch gate as
+    // MQ3. The 4 fused kernels (gemm_qkv/qkvza/gate_up/residual_hfp4g32_wmma)
+    // ship in pairs for gfx11 + gfx12; identical eligibility to llama.rs
+    // (see hipfire_runtime::llama::is_batchable_la).
+    let fp4_with_wmma = matches!(dt, DType::HFP4G32 | DType::MFP4G32)
+        && matches!(arch,
+            "gfx1100" | "gfx1101" | "gfx1102" | "gfx1150" | "gfx1151"
+            | "gfx1200" | "gfx1201"
+        );
+
     // Lloyd-MQ3 (MQ3G256Lloyd) on gfx11: Phase 5 of issue #116 ships the
     // gemm_*_mq3g256_lloyd_wmma family alongside the existing HFQ3 WMMA
     // path; group stride differs (112 B Lloyd vs 104 B HFQ3) so dispatch
@@ -3833,7 +3843,7 @@ fn is_batchable_la(dt: DType, arch: &str) -> bool {
         && matches!(arch, "gfx1200" | "gfx1201")
         && std::env::var("HIPFIRE_LLOYD_GFX12").ok().as_deref() == Some("1");
 
-    mq3_uniform_with_wmma || lloyd_mq3_with_gfx11_wmma || lloyd_mq3_with_gfx12_wmma
+    mq3_uniform_with_wmma || lloyd_mq3_with_gfx11_wmma || lloyd_mq3_with_gfx12_wmma || fp4_with_wmma
 }
 
 /// Process one chunk of up to `pbs.max_batch` tokens through the batched
@@ -4222,10 +4232,11 @@ fn forward_prefill_chunk(
                 // together (the all-together corruption-prevention rule from
                 // docs/plans/mq-lloyd-batched-prefill-followup.md). MQ4-Lloyd
                 // is wired in a separate PR (issue #182).
-                let is_mq = matches!(layer.wqkv.gpu_dtype, DType::MQ4G256 | DType::MQ6G256 | DType::MQ3G256 | DType::MQ3G256Lloyd);
+                let is_mq = matches!(layer.wqkv.gpu_dtype, DType::MQ4G256 | DType::MQ6G256 | DType::MQ3G256 | DType::MQ3G256Lloyd | DType::MFP4G32);
                 let is_6bit = matches!(layer.wqkv.gpu_dtype, DType::MQ6G256 | DType::HFQ6G256);
                 let is_mq3 = matches!(layer.wqkv.gpu_dtype, DType::MQ3G256);
                 let is_mq3_lloyd = matches!(layer.wqkv.gpu_dtype, DType::MQ3G256Lloyd);
+                let is_fp4 = matches!(layer.wqkv.gpu_dtype, DType::HFP4G32 | DType::MFP4G32);
 
                 // Batched rmsnorm (+ FWHT for MQ) for the LA preamble.
                 // x_batch / x_rot_batch are [N × dim] contiguous. For HFQ
@@ -4264,6 +4275,18 @@ fn forward_prefill_chunk(
                     // 104 B/group HFQ3-stride; X is already FWHT-rotated by
                     // fused_rmsnorm_rotate_mq_batched above.
                     gpu.gemm_qkvza_hfq3g256_wmma(
+                        &layer.wqkv.buf, &layer.wz.buf, &layer.w_beta.buf, &layer.w_alpha.buf,
+                        &pbs.x_rot_batch,
+                        &pbs.dn_qkv_batch, &pbs.dn_z_batch, &pbs.dn_beta_batch, &pbs.dn_alpha_batch,
+                        layer.wqkv.m, layer.wz.m, layer.w_beta.m, layer.w_alpha.m,
+                        layer.wqkv.k, n,
+                    )?;
+                } else if is_fp4 {
+                    // HFP4G32: 17-B blocks (vs HFQ4's 136-B groups), per-row 16-B header.
+                    // MFP4G32: same storage as HFP4 + offline-FWHT weights; X is already
+                    // rotated above when is_mq, so this branch handles both unrotated
+                    // (HFP4) and post-rotation (MFP4) activations identically.
+                    gpu.gemm_qkvza_hfp4g32(
                         &layer.wqkv.buf, &layer.wz.buf, &layer.w_beta.buf, &layer.w_alpha.buf,
                         &pbs.x_rot_batch,
                         &pbs.dn_qkv_batch, &pbs.dn_z_batch, &pbs.dn_beta_batch, &pbs.dn_alpha_batch,
@@ -4404,10 +4427,11 @@ fn forward_prefill_chunk(
                 // at quant time; math requires dot(rot(W), rot(x)) = dot(W,x)).
                 // For HFQ weights no rotation is needed — the activation
                 // feeds gemm_hfq{4,6}g256_residual directly.
-                let wo_is_mq = matches!(layer.wo.gpu_dtype, DType::MQ4G256 | DType::MQ6G256 | DType::MQ3G256 | DType::MQ3G256Lloyd);
+                let wo_is_mq = matches!(layer.wo.gpu_dtype, DType::MQ4G256 | DType::MQ6G256 | DType::MQ3G256 | DType::MQ3G256Lloyd | DType::MFP4G32);
                 let wo_is_6bit = matches!(layer.wo.gpu_dtype, DType::MQ6G256 | DType::HFQ6G256);
                 let wo_is_mq3 = matches!(layer.wo.gpu_dtype, DType::MQ3G256);
                 let wo_is_mq3_lloyd = matches!(layer.wo.gpu_dtype, DType::MQ3G256Lloyd);
+                let wo_is_fp4 = matches!(layer.wo.gpu_dtype, DType::HFP4G32 | DType::MFP4G32);
                 let wo_input = if wo_is_mq {
                     gpu.rotate_x_mq_batched(
                         &pbs.dn_normed_batch, &pbs.dn_normed_rot_batch, layer.wo.k, n,
@@ -4431,6 +4455,11 @@ fn forward_prefill_chunk(
                         &layer.wo.buf, wo_input, &pbs.x_batch,
                         layer.wo.m, layer.wo.k, n,
                     )?;
+                } else if wo_is_fp4 {
+                    gpu.gemm_hfp4g32_residual(
+                        &layer.wo.buf, wo_input, &pbs.x_batch,
+                        layer.wo.m, layer.wo.k, n,
+                    )?;
                 } else {
                     gpu.gemm_hfq4g256_residual(
                         &layer.wo.buf, wo_input, &pbs.x_batch,
@@ -4439,10 +4468,11 @@ fn forward_prefill_chunk(
                 }
 
                 // FFN: rmsnorm (+ rotate for MQ).
-                let ffn_is_mq = matches!(layer.w_gate.gpu_dtype, DType::MQ4G256 | DType::MQ6G256 | DType::MQ3G256 | DType::MQ3G256Lloyd);
+                let ffn_is_mq = matches!(layer.w_gate.gpu_dtype, DType::MQ4G256 | DType::MQ6G256 | DType::MQ3G256 | DType::MQ3G256Lloyd | DType::MFP4G32);
                 let ffn_is_6bit = matches!(layer.w_gate.gpu_dtype, DType::MQ6G256 | DType::HFQ6G256);
                 let ffn_is_mq3 = matches!(layer.w_gate.gpu_dtype, DType::MQ3G256);
                 let ffn_is_mq3_lloyd = matches!(layer.w_gate.gpu_dtype, DType::MQ3G256Lloyd);
+                let ffn_is_fp4 = matches!(layer.w_gate.gpu_dtype, DType::HFP4G32 | DType::MFP4G32);
                 if ffn_is_mq {
                     gpu.fused_rmsnorm_rotate_mq_batched(
                         &pbs.x_batch, &layer.ffn_norm, &pbs.x_rot_batch, dim, config.norm_eps, n,
@@ -4479,6 +4509,14 @@ fn forward_prefill_chunk(
                         layer.w_gate.m, layer.w_up.m,
                         layer.w_gate.k, n,
                     )?;
+                } else if ffn_is_fp4 {
+                    gpu.gemm_gate_up_hfp4g32(
+                        &layer.w_gate.buf, &layer.w_up.buf,
+                        &pbs.x_rot_batch,
+                        &pbs.gate_ffn_batch, &pbs.up_batch,
+                        layer.w_gate.m, layer.w_up.m,
+                        layer.w_gate.k, n,
+                    )?;
                 } else {
                     gpu.gemm_gate_up_hfq4g256(
                         &layer.w_gate.buf, &layer.w_up.buf,
@@ -4495,10 +4533,11 @@ fn forward_prefill_chunk(
                 // is purely element-wise and uses numel() as its length,
                 // so a [N × hidden_dim] tensor processes all rows in one
                 // launch with no batch offset needed.
-                let w_down_is_mq = matches!(layer.w_down.gpu_dtype, DType::MQ4G256 | DType::MQ6G256 | DType::MQ3G256 | DType::MQ3G256Lloyd);
+                let w_down_is_mq = matches!(layer.w_down.gpu_dtype, DType::MQ4G256 | DType::MQ6G256 | DType::MQ3G256 | DType::MQ3G256Lloyd | DType::MFP4G32);
                 let w_down_is_6bit = matches!(layer.w_down.gpu_dtype, DType::MQ6G256 | DType::HFQ6G256);
                 let w_down_is_mq3 = matches!(layer.w_down.gpu_dtype, DType::MQ3G256);
                 let w_down_is_mq3_lloyd = matches!(layer.w_down.gpu_dtype, DType::MQ3G256Lloyd);
+                let w_down_is_fp4 = matches!(layer.w_down.gpu_dtype, DType::HFP4G32 | DType::MFP4G32);
                 if w_down_is_mq {
                     gpu.fused_silu_mul_rotate_mq_batched(
                         &pbs.gate_ffn_batch, &pbs.up_batch, &pbs.ffn_hidden_batch,
@@ -4526,6 +4565,11 @@ fn forward_prefill_chunk(
                         &layer.w_down.buf, &pbs.ffn_hidden_batch, &pbs.x_batch,
                         layer.w_down.m, layer.w_down.k, n,
                     )?;
+                } else if w_down_is_fp4 {
+                    gpu.gemm_hfp4g32_residual(
+                        &layer.w_down.buf, &pbs.ffn_hidden_batch, &pbs.x_batch,
+                        layer.w_down.m, layer.w_down.k, n,
+                    )?;
                 } else {
                     gpu.gemm_hfq4g256_residual(
                         &layer.w_down.buf, &pbs.ffn_hidden_batch, &pbs.x_batch,
@@ -4550,10 +4594,11 @@ fn forward_prefill_chunk(
                 // launch covers all N tokens at once.
                 let kv_dim = config.n_kv_heads * config.head_dim;
                 let q_dim = config.n_heads * config.head_dim;
-                let qkv_is_mq = matches!(layer.wq.gpu_dtype, DType::MQ4G256 | DType::MQ6G256 | DType::MQ3G256 | DType::MQ3G256Lloyd);
+                let qkv_is_mq = matches!(layer.wq.gpu_dtype, DType::MQ4G256 | DType::MQ6G256 | DType::MQ3G256 | DType::MQ3G256Lloyd | DType::MFP4G32);
                 let qkv_is_6bit = matches!(layer.wq.gpu_dtype, DType::MQ6G256 | DType::HFQ6G256);
                 let qkv_is_mq3 = matches!(layer.wq.gpu_dtype, DType::MQ3G256);
                 let qkv_is_mq3_lloyd = matches!(layer.wq.gpu_dtype, DType::MQ3G256Lloyd);
+                let qkv_is_fp4 = matches!(layer.wq.gpu_dtype, DType::HFP4G32 | DType::MFP4G32);
 
                 // 1. rmsnorm (+ rotate for MQ) for the attn preamble.
                 if qkv_is_mq {
@@ -4588,6 +4633,17 @@ fn forward_prefill_chunk(
                     // X is already FWHT-rotated by fused_rmsnorm_rotate_mq_batched
                     // above; call the bare HFQ3 WMMA (no second rotation).
                     gpu.gemm_qkv_hfq3g256_wmma(
+                        &layer.wq.buf, &layer.wk.buf, &layer.wv.buf,
+                        &pbs.x_rot_batch,
+                        &pbs.fa_q_full_batch, &pbs.fa_k_batch, &pbs.fa_v_batch,
+                        layer.wq.m, layer.wk.m, layer.wv.m,
+                        layer.wq.k, n,
+                    )?;
+                } else if qkv_is_fp4 {
+                    // HFP4G32 / MFP4G32 FP4 batched WMMA. X is already
+                    // rotated above for MFP4 (is_mq path) — same kernel
+                    // covers both unrotated HFP4 and rotated MFP4 inputs.
+                    gpu.gemm_qkv_hfp4g32(
                         &layer.wq.buf, &layer.wk.buf, &layer.wv.buf,
                         &pbs.x_rot_batch,
                         &pbs.fa_q_full_batch, &pbs.fa_k_batch, &pbs.fa_v_batch,
@@ -4824,10 +4880,11 @@ fn forward_prefill_chunk(
 
                 // 9. wo residual: x_batch += wo · (optional rotate)(fa_attn_out_batch).
                 // Same MQ rotation requirement as the LA wo path.
-                let fa_wo_is_mq = matches!(layer.wo.gpu_dtype, DType::MQ4G256 | DType::MQ6G256 | DType::MQ3G256 | DType::MQ3G256Lloyd);
+                let fa_wo_is_mq = matches!(layer.wo.gpu_dtype, DType::MQ4G256 | DType::MQ6G256 | DType::MQ3G256 | DType::MQ3G256Lloyd | DType::MFP4G32);
                 let fa_wo_is_6bit = matches!(layer.wo.gpu_dtype, DType::MQ6G256 | DType::HFQ6G256);
                 let fa_wo_is_mq3 = matches!(layer.wo.gpu_dtype, DType::MQ3G256);
                 let fa_wo_is_mq3_lloyd = matches!(layer.wo.gpu_dtype, DType::MQ3G256Lloyd);
+                let fa_wo_is_fp4 = matches!(layer.wo.gpu_dtype, DType::HFP4G32 | DType::MFP4G32);
                 let fa_wo_input = if fa_wo_is_mq {
                     gpu.rotate_x_mq_batched(
                         &pbs.fa_attn_out_batch, &pbs.fa_attn_out_rot_batch, layer.wo.k, n,
@@ -4851,6 +4908,11 @@ fn forward_prefill_chunk(
                         &layer.wo.buf, fa_wo_input, &pbs.x_batch,
                         layer.wo.m, layer.wo.k, n,
                     )?;
+                } else if fa_wo_is_fp4 {
+                    gpu.gemm_hfp4g32_residual(
+                        &layer.wo.buf, fa_wo_input, &pbs.x_batch,
+                        layer.wo.m, layer.wo.k, n,
+                    )?;
                 } else {
                     gpu.gemm_hfq4g256_residual(
                         &layer.wo.buf, fa_wo_input, &pbs.x_batch,
@@ -4860,10 +4922,11 @@ fn forward_prefill_chunk(
 
                 // 10. FFN: rmsnorm (+ rotate for MQ), gate+up, silu_mul
                 // (+ rotate for MQ), w_down residual.
-                let fa_ffn_is_mq = matches!(layer.w_gate.gpu_dtype, DType::MQ4G256 | DType::MQ6G256 | DType::MQ3G256 | DType::MQ3G256Lloyd);
+                let fa_ffn_is_mq = matches!(layer.w_gate.gpu_dtype, DType::MQ4G256 | DType::MQ6G256 | DType::MQ3G256 | DType::MQ3G256Lloyd | DType::MFP4G32);
                 let fa_ffn_is_6bit = matches!(layer.w_gate.gpu_dtype, DType::MQ6G256 | DType::HFQ6G256);
                 let fa_ffn_is_mq3 = matches!(layer.w_gate.gpu_dtype, DType::MQ3G256);
                 let fa_ffn_is_mq3_lloyd = matches!(layer.w_gate.gpu_dtype, DType::MQ3G256Lloyd);
+                let fa_ffn_is_fp4 = matches!(layer.w_gate.gpu_dtype, DType::HFP4G32 | DType::MFP4G32);
                 if fa_ffn_is_mq {
                     gpu.fused_rmsnorm_rotate_mq_batched(
                         &pbs.x_batch, &layer.ffn_norm, &pbs.x_rot_batch, dim, config.norm_eps, n,
@@ -4898,6 +4961,14 @@ fn forward_prefill_chunk(
                         layer.w_gate.m, layer.w_up.m,
                         layer.w_gate.k, n,
                     )?;
+                } else if fa_ffn_is_fp4 {
+                    gpu.gemm_gate_up_hfp4g32(
+                        &layer.w_gate.buf, &layer.w_up.buf,
+                        &pbs.x_rot_batch,
+                        &pbs.gate_ffn_batch, &pbs.up_batch,
+                        layer.w_gate.m, layer.w_up.m,
+                        layer.w_gate.k, n,
+                    )?;
                 } else {
                     gpu.gemm_gate_up_hfq4g256(
                         &layer.w_gate.buf, &layer.w_up.buf,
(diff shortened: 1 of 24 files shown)


Kaden-Schutt:
## Deep-dive 1/6 — β empirical data: full bench tables

### β decode lift, 5-run paired comparison, all measurements

**gfx1201 R9700, 9B Qwen 3.5 MFP4G32 (hiptrx, 2026-05-11):**
| iter | HIPFIRE_GRAPH=0 | HIPFIRE_GRAPH=1 | speedup |
|---|---|---|---|
| 1 | 85.5 tok/s | 87.1 | 1.019× |
| 2 | 85.5 | 87.5 | 1.023× |
| 3 | 85.3 | 87.6 | 1.027× |
| 4 | 85.4 | 87.5 | 1.025× |
| 5 | 85.2 | 87.5 | 1.027× |

Mean **+2.4%**, range 1.019-1.027×, all-runs-positive, no overlap.

Default-on verification (HIPFIRE_GRAPH=0 forced off vs no-env default ON, 5 paired):
| iter | =0 | default | speedup |
|---|---|---|---|
| 1 | 85.1 | 87.6 | 1.029× |
| 2 | 85.2 | 87.3 | 1.025× |
| 3 | 85.3 | 87.3 | 1.023× |
| 4 | 85.0 | 87.3 | 1.027× |
| 5 | 85.1 | 87.3 | 1.026× |

Mean **+2.6%**, identical pattern → policy works correctly.

**gfx1100 7900 XTX, 9B HFP4G32 (local box, 2026-05-11):**

Pre-default 5-run (HIPFIRE_GRAPH=0 vs HIPFIRE_GRAPH=1):
| iter | =0 | =1 | speedup |
|---|---|---|---|
| 1 | 99.3 | 100.0 | 1.007× |
| 2 | 99.1 | 99.9  | 1.008× |
| 3 | 98.6 | 99.5  | 1.009× |
| 4 | 99.0 | 100.0 | 1.010× |
| 5 | 99.1 | 99.4  | 1.003× |

Mean **+0.7%**, range 1.003-1.010×.

Post-default 5-run (HIPFIRE_GRAPH=0 vs no-env default ON; DPM state different across sessions):
| iter | =0 | default | speedup |
|---|---|---|---|
| 1 | 109.8 | 111.5 | 1.015× |
| 2 | 109.3 | 111.4 | 1.019× |
| 3 | 109.8 | 110.6 | 1.007× |
| 4 | 109.2 | 111.3 | 1.019× |
| 5 | 109.4 | 111.1 | 1.016× |

Mean **+1.55%**. Absolute baseline shifted (DPM/thermal), direction stable.

**gfx1100 7900 XTX, 0.8B HFP4G32 (local box, 2026-05-11):**
| iter | =0 | =1 | speedup |
|---|---|---|---|
| 1 | 276.0 | 278.3 | 1.008× |
| 2 | 277.9 | 278.2 | 1.001× |
| 3 | 276.9 | 278.0 | 1.004× |
| 4 | 275.1 | 276.9 | 1.007× |
| 5 | 275.9 | 278.4 | 1.009× |

Mean **+0.6%**.

### Why gfx12 wins bigger than gfx11

The gfx1201 R9700 measured ~25 µs per-launch overhead via rocprof; the gfx1100 7900 XTX has lower per-launch overhead (~10-15 µs based on the smaller absolute win). When less overhead is available to amortize, β saves less. Both wins are real and consistent.

### What β actually captures

In `qwen35.rs` `forward_scratch()`, the captured region is `forward_scratch_layers` — the entire decode forward pass including all gemvs, attention kernels, residuals, rmsnorm chain, and lm_head. The embedding lookup (1 launch/token, depends on host-side token id) and the sampler (1 launch + D2H sync, mandatory host blocking point) remain outside the graph.

### Coexistence with FP8

⚠ β and HIPFIRE_FP8_WMMA=1 **CONFLICT** (-3.2% combined). When launch overhead is amortized via graphs, FP8's per-kernel context-cost disadvantage becomes the dominant factor. Both env vars can be set together but the FP8 gate is documented as opt-in default-off precisely because of this interaction.

Kaden-Schutt:
## Deep-dive 2/6 — FP8 falsification: why each variant lost in production

### v1 — `ensure_fp8_x` pre-pass + decode FP8 dot4 GEMV

**Architecture**: separate `pack_f32_to_fp8.gfx12.hip` kernel converts F32 X → FP8 scratch once per gemv call (cached by src_ptr), then `gemv_hfp4g32_fp8_gfx12` reads pre-packed FP8 and uses `dot4_f32_fp8_fp8`.

**Synthetic bench (hot cache, isolated)**: 1.04-1.42× on FFN shapes (gate_up M=11008 hits 99% peak BW vs 78% fallback).

**Production decode on 9B MFP4G32**: 0.93× — the `mq_x_rot` scratch buffer has stable pointer but per-rotation data turnover, forcing cache invalidation on every weight_gemv call. The pack launch (~5-10 µs per call) ended up on the critical path of every gemv, exceeding the dot4 ALU savings (~3-5 µs per call).

### v2 — inline F32→FP8 cvt (no pre-pass)

**Architecture**: GEMV kernel reads F32 directly and converts to FP8 per K-block via `cvt_pk_fp8_f32`. Eliminates the pack launch.

**Synthetic bench**: 0.79-0.97× — actually worse than v1!

**Why**: each warp now reads K F32 values (4 bytes/elem) instead of K FP8 bytes (1 byte/elem). For M=11008 with K=2048, total activation reads per warp scaled from ~22 MB (v1, FP8 scratch) to ~88 MB (v2, F32 source). x fits in L2 but 88 MB of reads from 1024 warps × 8 KB exceeds L2 capacity (16 MB on R9700), forcing DRAM refills.

### v3 / A3 — fused FWHT + FP8 pack (`mq_rotate_x_dual_fp8.gfx12.hip`)

**Architecture**: extend the rotation kernel (`mq_rotate_x`) to write BOTH F32 (for legacy consumers) AND FP8 outputs in one launch. Eliminates the separate pack launch while keeping FP8 memory access pattern for the GEMV.

**Synthetic bench**: 1.03-1.27× on FFN shapes via the production rotate+gemv path.

**Production decode on 9B MFP4G32**: 0.94× — closer to break-even but still negative. Per-shape breakdown:

| Shape | env=0 µs | env=1 µs | delta |
|---|---|---|---|
| wo (qkv-q M=2048 K=2048) | 18.56 | 20.13 | **+1.57 µs (LOSS)** |
| gate (M=11008 K=2048) | 22.36 | 18.77 | -3.59 µs (win) |
| up (M=11008 K=2048) | 22.36 | 18.77 | -3.59 µs (win) |
| down (M=2048 K=11008) | 24.28 | 20.10 | -4.18 µs (win) |

Per-token expected: 32 layers × (-50 µs wo + 115 µs gate + 115 µs up + 134 µs down) = **+314 µs/token saved**. Measured: **-700 µs/token loss**. ~1 ms unexplained gap = L2 thrashing from interleaved attention/conv1d/norm ops between gemvs.

### v3 + shape-gate (FP8_GEMV_MIN_M=4096)

Skip FP8 dispatch when M < 4096 (where the per-shape bench showed losses). Still 0.94× — confirms the issue is workload-level, not shape-level.

### v3 + batch-gate (FP8_WMMA_MIN_BATCH=1024)

For the prefill FP8 WMMA path, only enable at large N. Prevents decode (N=1) regression. Still 0.94× on decode — the gate prevents WMMA path regression but doesn't help the dot4 GEMV path either way.

### rocprof attribution (gfx12 9B MFP4G32, FP8=0 vs FP8=1)

| Kernel | env=0 | env=1 |
|---|---|---|
| gemv_hfp4g32 (fallback, all shapes) | 4981 × 33 µs avg = 164.7 ms | 1280 × 3.8 µs avg = 4.9 ms (small-M only) |
| gemv_hfp4g32_fp8_gfx12 | — | 3701 × **46 µs avg** = 170.7 ms (large-M only) |
| mq_rotate_x_dual_fp8 | — | 1301 × 1.7 µs = 2.2 ms |
| pack_f32_to_fp8 | — | 1280 × 1.4 µs = 1.7 ms |

Apples-to-apples (3701 large-M calls): baseline 159.8 ms = **43 µs/call**, FP8 170.7 ms = **46 µs/call**. **FP8 is +3 µs SLOWER per call in production** despite winning by 4 µs in synthetic isolation.

The ~30 µs gap between synthetic (15.3 µs gate_up) and production (46 µs gate_up) per call is launch + L2 + scheduler context cost. FP8 path pays 7 µs more of this than fallback (~23 µs context cost for fallback). That's where the 4 µs synthetic win is consumed and reversed.

Kaden-Schutt:
## Deep-dive 2/6 — FP8 falsification: why each variant lost in production

### v1 — `ensure_fp8_x` pre-pass + decode FP8 dot4 GEMV

**Architecture**: separate `pack_f32_to_fp8.gfx12.hip` kernel converts F32 X → FP8 scratch once per gemv call (cached by src_ptr), then `gemv_hfp4g32_fp8_gfx12` reads pre-packed FP8 and uses `dot4_f32_fp8_fp8`.

**Synthetic bench (hot cache, isolated)**: 1.04-1.42× on FFN shapes (gate_up M=11008 hits 99% peak BW vs 78% fallback).

**Production decode on 9B MFP4G32**: 0.93× — the `mq_x_rot` scratch buffer has stable pointer but per-rotation data turnover, forcing cache invalidation on every weight_gemv call. The pack launch (~5-10 µs per call) ended up on the critical path of every gemv, exceeding the dot4 ALU savings (~3-5 µs per call).

### v2 — inline F32→FP8 cvt (no pre-pass)

**Architecture**: GEMV kernel reads F32 directly and converts to FP8 per K-block via `cvt_pk_fp8_f32`. Eliminates the pack launch.

**Synthetic bench**: 0.79-0.97× — actually worse than v1.

**Why**: each warp now reads K F32 values (4 bytes/elem) instead of K FP8 bytes (1 byte/elem). For M=11008 with K=2048, total activation reads per warp scaled from ~22 MB (v1, FP8 scratch) to ~88 MB (v2, F32 source). x fits in L2 but 88 MB of reads from 1024 warps × 8 KB exceeds L2 capacity (16 MB on R9700), forcing DRAM refills.

### v3 / A3 — fused FWHT + FP8 pack (`mq_rotate_x_dual_fp8.gfx12.hip`)

**Architecture**: extend the rotation kernel to write BOTH F32 (legacy consumers) AND FP8 outputs in one launch. Eliminates the separate pack launch while keeping FP8 memory access pattern.

**Synthetic bench**: 1.03-1.27× on FFN shapes via the production rotate+gemv path.

**Production decode on 9B MFP4G32**: 0.94× — closer to break-even but still negative. Per-shape:

| Shape | env=0 µs | env=1 µs | delta |
|---|---|---|---|
| wo (M=2048 K=2048) | 18.56 | 20.13 | **+1.57 µs (LOSS)** |
| gate (M=11008 K=2048) | 22.36 | 18.77 | -3.59 µs (win) |
| up (M=11008 K=2048) | 22.36 | 18.77 | -3.59 µs (win) |
| down (M=2048 K=11008) | 24.28 | 20.10 | -4.18 µs (win) |

Per-token expected: 32 layers × (-50 + 115 + 115 + 134 µs) = +314 µs/token saved. Measured: -700 µs/token loss. ~1 ms unexplained gap = L2 thrashing from interleaved attention/conv1d/norm ops between gemvs.

### rocprof attribution (gfx12 9B MFP4G32, FP8=0 vs FP8=1)

| Kernel | env=0 | env=1 |
|---|---|---|
| gemv_hfp4g32 (fallback) | 4981 × 33 µs avg = 164.7 ms | 1280 × 3.8 µs (small-M only) |
| gemv_hfp4g32_fp8_gfx12 | — | 3701 × **46 µs avg** = 170.7 ms |
| mq_rotate_x_dual_fp8 | — | 1301 × 1.7 µs |
| pack_f32_to_fp8 | — | 1280 × 1.4 µs |

Apples-to-apples (3701 large-M calls): baseline 43 µs/call, FP8 46 µs/call. **FP8 is +3 µs SLOWER per call in production** despite winning by 4 µs in synthetic isolation.

The ~30 µs gap between synthetic (15.3 µs gate_up) and production (46 µs gate_up) per call is launch + L2 + scheduler context cost. FP8 path pays 7 µs more of this than fallback.

Kaden-Schutt:
## Deep-dive 3/6 — gfx11 fdot2 trickle-down: synthetic 2.08× vs production 0.90×

### The hypothesis

After gfx12 FP8 was falsified, the natural next experiment was the kernel-structure improvements applied to gfx11 with its available primitives:
- gfx11 has `v_dot2_f32_f16` (fdot2 builtin: 2 FP16 muls + 1 FP32 add per VALU)
- Current `gemv_hfp4g32.hip` fallback measured 40-44% peak BW on FFN shapes on 7900 XTX (so plenty of ALU headroom)
- No FP8 conversion overhead (F32→FP16 cvt is essentially one VALU per element)
- Same per-block UE8M0 architecture but applied post-fdot2 to F32 acc

**Expected**: per-call kernel savings cleanly translate to production win, unlike FP8 which had memory-pressure issues.

### Two kernel variants tested

**v2 (single-carry chain)**: `partial = fdot2(..., partial)` ×4 per K-block.
- 272 total instructions (vs 662 in fallback)
- 26 VGPRs (vs 83 in fallback)
- Synthetic 1.07-1.95× on FFN shapes
- But: 4-deep dependency chain on `partial` register; ~4-cycle fdot2 latency × 4 = 20+ cycle stall

**v3 (4 independent partial accumulators)**: `p0,p1,p2,p3 = fdot2(...,0)` then balanced-tree reduce.
- Better ILP, dual-issue friendly
- Synthetic 2.08× on gate_up (78%→84% peak BW)
- Same NRMSE 2e-4 (much better than FP8's 2.7%)

### Synthetic bench (7900 XTX, K=2048)

| Shape | Fallback %BW | v3 dot2 %BW | Speedup |
|---|---|---|---|
| qkv-q M=2048 K=2048 | 21% | 21% | 0.99× |
| qkv-kv M=512 K=2048 | 7.8% | 7.5% | 0.96× |
| **gate_up M=11008 K=2048** | 40% | **84%** | **2.08×** |
| w_down M=2048 K=11008 | 46% | 61% | 1.31× |
| med M=4096 K=2048 | 43% | 49% | 1.13× |
| small M=1024 K=2048 | 21% | 21% | 1.03× |

### Production smoke (9B HFP4G32, 7900 XTX, 5-run paired)

| iter | fallback decode | dot2 v3 decode | ratio |
|---|---|---|---|
| 1 | 109.6 | 99.1 | 0.904× |
| 2 | 109.5 | 98.9 | 0.903× |
| 3 | 109.7 | 98.8 | 0.901× |
| 4 | 109.7 | 98.9 | 0.901× |
| 5 | 109.8 | 99.0 | 0.902× |

**Mean: 0.90× — 10% slower in production** despite 2.08× synthetic win on the same hardware and similar shapes. Tight variance, no overlap with baseline.

### Why the synthetic-to-production gap

Same architectural finding as gfx12 FP8:

1. Synthetic bench reuses same `x` tensor across 200 trials → caches stay hot, kernel runs in isolation
2. Production interleaves gemv with rmsnorm, residuals, attention, conv1d, sample — these evict cache and contend for memory pipeline
3. The new kernel's memory access pattern (FP16 conversion of F32 reads, scattered tile layout) interacts worse with the production op sequence than the fallback's all-F32 path
4. Per-call launch overhead (~25 µs on this codebase, established via rocprof) dominates per-call kernel savings (~2-5 µs) for small shapes

The fdot2 instruction itself emits correctly (verified via llvm-objdump: clean `v_dot2acc_f32_f16` instructions). The kernel is correct (NRMSE 2e-4). It just doesn't translate to a production win.

### Pattern across three independent attempts

| Variant | Synthetic | Production decode |
|---|---|---|
| gfx12 FP8 dot4 GEMV (v1/v2/v3) | 1.04-1.42× | 0.93-0.94× |
| **gfx11 fdot2 GEMV (v2/v3)** | **1.13-2.08×** | **0.90×** |
| HIPFIRE_GRAPH=1 (β) | n/a | **+0.6 to +2.6%** |

Three independent kernel-level ALU experiments all lose in production despite synthetic wins. That's not bad luck on FP8 specifically — it's a structural property of this codebase + ROCm 7.2 + decode workload pattern. Only launch-reduction (β) survives contact with production.

Kaden-Schutt:
## Deep-dive 4/6 — `invalidate_x_caches_for`: silent stale-data correctness bug

### The bug

`ensure_fp16_x` and `ensure_fp8_x` in `dispatch.rs` cache the conversion result by source pointer:

```rust
fn ensure_fp16_x(&mut self, x: &GpuTensor, n_elems: usize) -> HipResult<*mut c_void> {
    let src_ptr = x.buf.as_ptr();
    // ...
    let must_convert = self.capture_mode || self.fp16_x_source_ptr != src_ptr;
    if must_convert {
        // launch convert kernel, write to scratch
        self.fp16_x_source_ptr = src_ptr;
    }
    Ok(self.fp16_x_scratch.as_ref().unwrap().as_ptr())
}
```

This works correctly when the caller passes a fresh `x` per call. But on the MFP4G32 decode path, `gemv_mfp4g32_with_rotate` does:

```rust
self.rotate_x_mq(x, x_rot, k)?;        // writes new rotated data to mq_x_rot
self.gemv_hfp4g32(a_raw, x_rot, y, m, k)
```

Where `x_rot` is an alias to `gpu.mq_x_rot` (a persistent scratch buffer). The pointer is stable across every weight_gemv(MFP4G32) call. The data turns over per rotation.

`gemv_hfp4g32` (when the FP8 path is active) calls `ensure_fp8_x(x_rot, k)`. On call 2..N within the same forward, the cache sees `fp8_x_source_ptr == src_ptr` and **silently returns stale FP8 from the previous gemv's rotation**.

### How long this was silently broken

Memory checkpoint `ada4eddf` (v2 #2 batched WMMA prefill) had noted:
> ensure_fp16_x src_ptr cache caveat noted (latent in HFQ4 test too)

But it was never patched. The bug existed in `ensure_fp16_x` even without the FP8 work — any caller that wrote to a previously-cached source would silently use stale data. The FP8 work was the first to widely exercise this pattern, which exposed the issue.

### Why coherence detectors missed it

The probe used `lru_cache_pep8_strict.txt` + temperature 0.0. With stale activation data on weight_gemv calls 2..N, the model still produced **coherent output on-topic with the prompt** — no single-token attractors, no n-gram loops, no special-token leaks. The detectors are calibrated to catch gross failure modes (loops, attractors, halt-in-think) but not subtle precision degradation. So `verdict: OK` across every test even though the values were wrong.

### Fix

Add `invalidate_x_caches_for(dst_ptr)` as a `dispatch.rs` helper:

```rust
fn invalidate_x_caches_for(&mut self, dst_ptr: *mut c_void) {
    if self.fp16_x_source_ptr == dst_ptr {
        self.fp16_x_source_ptr = std::ptr::null_mut();
    }
    if self.fp8_x_source_ptr == dst_ptr {
        self.fp8_x_source_ptr = std::ptr::null_mut();
    }
}
```

Called at the end of every rotation kernel that overwrites `mq_x_rot`:
- `rotate_x_mq`
- `rotate_x_mq_batched`
- `fused_rmsnorm_rotate_mq`
- `fused_rmsnorm_rotate_mq_batched`
- `fused_silu_mul_rotate_mq`
- `fused_silu_mul_rotate_mq_batched`
- `mq_rotate_x_dual_fp8_gfx12` (new, from this PR)

### Why this is shipped default-on regardless of FP8 verdict

The bug existed before FP8 — any code path that ever calls `ensure_fp16_x` followed by `ensure_fp16_x` again on a buffer that was overwritten in between (e.g., a future kernel that rewrites mq_x_rot, or prefill batched paths) would silently corrupt activations. The fix is correctness, not perf.

Performance impact of the fix on the non-FP8 path: zero. The cache invalidation is conditional (only fires when src_ptr matches) and adds 2 pointer compares per rotation kernel — sub-nanosecond cost. The fp16/fp8 caches retain their hit-on-same-src-ptr-with-same-data behavior; they just stop incorrectly hitting after the data has been overwritten.

### Verification

5-run paired smoke on 9B Qwen 3.5 MFP4G32 (gfx1201): coherence detectors PASS in both pre-fix and post-fix states (the FP8 path was off in the pre-fix runs that established the baseline anyway). Post-fix with FP8 ON: still PASS, confirming the cache invalidation doesn't break the FP8 path correctness either.

Kaden-Schutt:
## Deep-dive 5/6 — rocprof methodology + the synthetic-vs-production gap (lessons for next reviewer)

### How the production gap was located

After v1/v2/v3 of FP8 dot4 GEMV all measured negative on production decode despite winning on synthetic bench, I ran `rocprofv3 --kernel-trace --stats` on a 20-token coherence_probe run with HIPFIRE_FP8_WMMA=0 vs HIPFIRE_FP8_WMMA=1 on 9B Qwen 3.5 MFP4G32 on hiptrx R9700:

```bash
HIPFIRE_FP8_WMMA={0,1} rocprofv3 --kernel-trace --stats \
  --output-format csv -d /tmp/rocprof-{baseline,fp8} -- \
  ./target/release/examples/coherence_probe \
    --model ~/.hipfire/models/qwen3.5-9b.mfp4 \
    --prompt-file benchmarks/prompts/lru_cache_pep8_strict.txt \
    --max-tokens 20 --temperature 0.0
```

The `*_kernel_stats.csv` output gave per-kernel Calls/TotalDurationNs/AverageNs/MinNs/MaxNs/StdDev. The smoking gun was apples-to-apples comparison of GEMV-class kernels:
- Fallback at large-M shapes: **43 µs avg per call** (3701 calls × 159.8 ms total)
- FP8 dot4 at same large-M shapes: **46 µs avg per call** (3701 calls × 170.7 ms total)
- Synthetic isolated bench on same shape: **15.3 µs for FP8, 19.5 µs for fallback** (FP8 wins 4 µs)

**The production-vs-synthetic gap is ~28 µs/call for fallback and ~31 µs/call for FP8.** That extra 3 µs is what FP8 loses by.

### Why synthetic bench misleads

`test_gemv_hfp4g32_fp8.rs` and similar harnesses call the kernel 200 times in a tight loop on the same `x`/`w` tensors:

```rust
for _ in 0..warmup { gpu.gemv_hfp4g32_fp8_gfx12(...) }
let t = Instant::now();
for _ in 0..trials { gpu.gemv_hfp4g32_fp8_gfx12(...) }
```

- `ensure_fp8_x` cache hits all 200 trials (same src_ptr, same data)
- L2 is hot for weights and activations
- No interleaved attention/conv1d/norm kernels evicting cache lines
- No host-side bookkeeping introducing variance
- All 200 calls back-to-back at peak DPM state

Production decode has **none** of these. Each gemv call sees a different x (or stale-then-invalidated), the L2 is shared with all the layer's other kernels, the scheduler is juggling many concurrent waves, and there's host-side bookkeeping between forward passes.

### Methodology for next kernel-class experiment

If anyone reaches for "I have a new kernel that's 1.5× faster on synthetic bench":

1. **Don't ship default-on based on synthetic data alone.** The gap on this codebase has been measured at 6-21% loss between synthetic predictions and production outcomes.

2. **A/B against production decode.** The `coherence_probe` 5-run paired methodology in this branch:
   - 9B model at max-tokens=120, temperature=0.0, fixed prompt
   - Pair (forced-off, default/forced-on) per iteration
   - Variance < 1.5% across iterations is sufficient
   - All 5 iterations should be same-sign (no overlap with zero)

3. **rocprof if production disagrees with synthetic.** The kernel_stats.csv gives apples-to-apples per-kernel comparison. Look for the per-call avg gap, not just the total.

4. **Bench both routes.** If you write a new dispatch fn, also add a `_fallback` direct-call entry like `gemv_hfp4g32_fallback` so you can A/B without an env gate (which has its own overhead — see deep-dive 6 for the OnceLock fix).

### Tools shipped in this PR for the next researcher

| Tool | Purpose |
|---|---|
| `test_gemv_hfp4g32_fp8.rs` | Correctness + synthetic perf vs fallback, 6 production shapes |
| `test_gemv_hfp4g32_dot2.rs` | Same harness for gfx11 dot2 variant |
| `test_gemm_hfp4g32_fp8.rs` | Prefill correctness + perf vs FP16 WMMA reference, N sweep 1-2048 |
| `bench_gemv_hfp4g32_bw.rs` | Effective BW probe per shape, peak-% indicator |
| `bench_gemv_mfp4g32_rotate.rs` | Production-realistic path (rotate + gemv) — closest synthetic to production we have |

The bench tooling stays useful regardless of which kernel variants survive — future work should bench against these harnesses AND against `coherence_probe` 5-run paired smoke before flipping any default.

Kaden-Schutt:
## Deep-dive 6/6 — review notes: rebase state, threshold tuning, follow-ups

### Branch state vs origin/master

This branch is 5 commits ahead and 50 commits behind origin/master at PR-open time. The 50 commits behind are mostly the quality-eval harness work merged via #233 — not expected to conflict with this branch's kernel work in any substantive way. Reviewer can rebase locally or use GitHub's "Rebase and merge" / "Squash and merge" at the appropriate time.

The one upstream commit worth knowing about is `5716dcfa fix(hfp4): tighten quantize + load to match kernel K%256 constraint` — touches `qwen35.rs:750/756`, `hfq.rs`, and `main.rs`. Lives in a different file region from this branch's edits; rebase should be conflict-free.

### Empirical thresholds in this PR (tunable knobs)

Two constants in `dispatch.rs` encode breakeven measurements:

```rust
/// Prefill FP8 WMMA gate. At N < 1024 the FP8 WMMA path measures
/// 0.71-0.94× vs FP16 WMMA on gfx1201. Never enabled at decode (N=1).
const FP8_WMMA_MIN_BATCH: usize = 1024;

/// Decode FP8 dot4 GEMV shape gate. At M < 4096 the fallback wins or
/// ties on gfx1201. Empirical embodiment of "Option α" shape-based
/// routing — still net-negative in production, but isolates the per-
/// shape behavior cleanly.
const FP8_GEMV_MIN_M: usize = 4096;
```

These should be reconsidered before any future attempt to flip the FP8 default-on. As of this PR, the FP8 paths are opt-in default-off, so the thresholds only affect researchers who set `HIPFIRE_FP8_WMMA=1`.

### Env vars introduced

| Var | Default | Purpose |
|---|---|---|
| `HIPFIRE_FP8_WMMA` | unset (off) | Opt-in to FP8 WMMA prefill + FP8 dot4 decode |
| `HIPFIRE_DOT2_GEMV` | unset (off) | Opt-in to gfx11 fdot2 decode GEMV |
| `HIPFIRE_GRAPH` | default-on for gfx11+gfx12 | β kill switch when set to "0" |
| `HIPFIRE_GRAPH_MOE` | unset (off) | Diagnostic bypass for MoE graph drift |

Both `HIPFIRE_FP8_WMMA` and `HIPFIRE_DOT2_GEMV` use OnceLock — the env var is read once at first dispatch, not per-call.

### Follow-up tasks (not in this PR)

1. **Multi-arch β coverage**: bench HIPFIRE_GRAPH=1 on gfx1010 (5700 XT), gfx1030 (6950 XT), gfx906 (Vega 20 / MI50) before flipping their defaults. Same 5-run paired methodology.

2. **MoE graph divergence root cause**: HIPFIRE_GRAPH_MOE=1 currently produces output drift after ~30-50 tokens on Qwen 35B-A3B (numerical reordering somewhere in DeltaNet state-update kernels). The repro is in `qwen35.rs:2942-2946`. Unlocking β for MoE would lift A3B inference.

3. **PR5 D0-D3b speculative prefetch with new baselines**: memory note says PR5 was net-negative (-1.7% / -7%) on hipx in 2026-05-08, but that was without β default-on and without these cleanup fixes. Worth a fresh re-bench if anyone tackles speculative decode optimization.

4. **Decode path coverage**: this PR's wins target the dense Qwen 3.5 decode forward (`forward_scratch`). The `forward_scratch_multi` (multi-GPU PP) path doesn't yet hit the β branch — would need parallel measurement and policy plumbing.

5. **HFP4G32 K%256 constraint upstream**: see `5716dcfa` on origin/master. This branch doesn't conflict but the constraint reminder applies to any new FP4 kernel.

### Why the FP8 + dot2 research is shipped vs reverted

The decision to keep the FP8 + dot2 kernels as opt-in scaffold rather than reverting them:

- **Reproducer cost**: rebuilding the kernel + dispatch + test harness from scratch is ~1 day of work each. Keeping them costs zero (default-off, no codepath unless gated).
- **Empirical record**: the production-vs-synthetic gap finding has implications for future kernel work on this codebase. Having the live kernels available means anyone can re-bench them under different conditions (different ROCm version, different model, different arch).
- **Bench tooling**: the test_gemm/test_gemv/bench_gemv harnesses are the cleanest decode-path bench infra in the repo. They stay usable for the next kernel-class experiment regardless of which kernels they target.
- **Negative-result documentation**: memory notes (`project_fp8_wmma_hfp4g32_2026_05_10.md`, `project_hypothesis_beta_graph_wins_2026_05_11.md`, `project_gfx11_dot2_trickle_down_falsified_2026_05_11.md`) capture the analysis. The shipped code is the receipts.

### Review checklist

For a quick review pass:

- [ ] `qwen35.rs:2962-2970` — β default-on policy for gfx11+gfx12 (one match arm, comment block above explains)
- [ ] `dispatch.rs:180-220` — OnceLock env-gate helpers (replace per-call `std::env::var`)
- [ ] `dispatch.rs` rotation kernels — verify each calls `self.invalidate_x_caches_for(xrp)` at the end
- [ ] `dispatch.rs` gemv_hfp4g32 + gemv_mfp4g32_with_rotate dispatch routing — verify FP8/dot2 paths are gated off by default
- [ ] `dispatch.rs:gemv_hfp4g32_fallback` — direct-call entry for A/B benching; preserves the legacy path

Everything else is opt-in scaffold + bench tools.

PR #245, opened 2026-05-13
feat: Jinja-everywhere phase 1-3a — structured tools/messages

Kaden-Schutt:
## Summary

Opens the daemon's chat-templating path to **structured tools and multi-turn
messages**, replacing the hand-rolled Hermes-block-in-system-prompt rendering
the CLI ships today. Lifts hipfire's tool-call support from "the one format
the system-prompt instructs" toward "whatever the upstream chat_template
emits" — XML for Qwen3.5/3.6, JSON for Hermes-trained models, etc.
Backward-compat invariant: gated on `HIPFIRE_JINJA_CHAT=1` (still off by
default); Jinja-off daemons consume the exact same legacy `prompt`+`system`
text as today, byte-identical.

Phase breakdown (one commit per layer, easy per-layer review):

- **Phase 1** (`6f94d8af`) — Daemon JSONL schema gains `tools` and
  `messages` fields. AR and DFlash paths both render through
  `JinjaChatFrame::render_messages` when present, synthesizing
  `[system?, user]` if no explicit history was passed.
- **Phase 1 smoke** (`eadf1007`) — `scripts/agentic-gate-jinja-tools.sh`
  drives an end-to-end smoke that actually exercises the DFlash branch
  (with the matching per-family DFlash drafter) and asserts the
  `done.dflash` signature so silent fall-through to AR is caught.
- **Phase 2** (`198cebcf`) — CLI `/v1/chat/completions` now passes
  `body.tools` structurally and maps `body.messages` to the daemon's
  typed `Message[]` shape. Legacy inline-ChatML + text-rendered
  tools-block are still sent so HIPFIRE_JINJA_CHAT=0 daemons see no
  behavior change. `parseOneToolCall` gains a probe for Qwen3.5/3.6's
  native `<function=NAME><parameter=K>V</parameter>...</function>` shape
  with typed value coercion (numbers / bools / null / nested JSON).
- **Phase 3a** (`abcfe7da`) — Multi-GPU PP path (`generate_multi`) gets
  the same Jinja branch as AR and DFlash. Compile-verified locally;
  runtime needs multi-GPU hardware.

VL path (`generate_vl`) and the Jinja-default-flip are explicitly punted to
follow-up PRs.

## Test plan

- [x] **Unit tests**: 161 Rust runtime lib tests pass (5 new in
      `prompt_frame`: minimal/assistant-tool-call/tool-response
      deserialize round-trip, `{% if tools %}` block firing,
      multi-turn render with assistant `tool_calls` + tool response
      carrying `tool_call_id`).
- [x] **Unit tests**: 18 TS parser tests pass (14 prior + 4 new XML cases:
      Qwen3.6 native shape with `<parameter=K>V</parameter>` siblings,
      typed-value coercion, zero-arg call, regression coverage for
      `<function=NAME>{JSON}` corruption shape).
- [x] **End-to-end daemon smoke** (AR + DFlash): `qwen3.6-27b.mq4` +
      `qwen36-27b-dflash-mq4.hf4` under `HIPFIRE_JINJA_CHAT=1` emits
      native XML `<function=get_weather><parameter=city>...` —
      `done.dflash=true`, τ=2.25, cycles=12.
- [x] **End-to-end CLI smoke** (full pipeline): `hipfire serve` + curl
      `/v1/chat/completions` with structured tools → daemon → Jinja →
      XML output → CLI parser → standard OpenAI `tool_calls` response
      with typed `arguments={"city":"San Francisco","unit":"f"}`,
      `finish_reason="tool_calls"`, `content=null`.
- [x] **Regression battery** (4 cells × 2 envs on `qwen3.5-35b-a3b.mq4`):
      `scripts/agentic-gate.sh` produces **identical per-cell outcomes**
      with and without `HIPFIRE_JINJA_CHAT=1` (2 PASS + 2 SOFT_WARN; the
      SOFT_WARN cells are pre-existing fragility from the gate's
      no-reset-between-cells design + daemon-honors-system-only-on-
      seq_pos==0 — reproduces on master).
- [ ] **Phase 3a runtime** (multi-GPU PP): needs hiptrx/hipx hardware
      to actually execute. Compile-verified only in this PR.
- [ ] **VL path**: deferred to follow-up (image-pad token positioning
      non-trivial).

### DIFF
--- a/cli/index.ts
+++ b/cli/index.ts
@@ -1479,6 +1479,67 @@ async function serve(port: number) {
           return "";
         };
 
+        // Strip <think>...</think> blocks from historical assistant text. Same
+        // rationale as the inline-ChatML builder below (line 1513): the Qwen3.5
+        // chat template doesn't carry thinking forward, and including it in
+        // history-shaped structured messages pollutes the KV cache.
+        const stripThinkingInline = (s: string): string =>
+          s.replace(/<think>[\s\S]*?<\/think>\s*/g, "")
+           .replace(/<think>[\s\S]*$/, "");
+
+        // Map OpenAI chat-completion messages to the daemon's structured
+        // `messages` JSONL field (Phase 2 of Jinja-everywhere). Roles:
+        //   developer → system (OpenAI o1/o3 alias — chat templates
+        //                       only know system/user/assistant/tool).
+        //   tool_calls.function.arguments is JSON-string in OpenAI;
+        //   the daemon expects a raw JSON value, so we parse here and
+        //   pass through any non-string `arguments` unchanged.
+        //
+        // The daemon arbitrates whether to consume `messages` or fall
+        // back to the legacy inline-ChatML `prompt` based on
+        // HIPFIRE_JINJA_CHAT — clients don't need to know which path
+        // fires. We always send both shapes so backward-compat with
+        // Jinja-off daemons holds.
+        const mapMessagesToStructured = (msgs: any[]): any[] => {
+          const out: any[] = [];
+          for (const m of msgs) {
+            if (!m || typeof m !== "object") continue;
+            let role: string = m.role;
+            if (role === "developer") role = "system";
+            if (role !== "system" && role !== "user" && role !== "assistant" && role !== "tool") {
+              continue;
+            }
+            const entry: any = { role, content: "" };
+            if (role === "assistant") {
+              entry.content = stripThinkingInline(extractText(m.content));
+              if (Array.isArray(m.tool_calls) && m.tool_calls.length > 0) {
+                const tcs: any[] = [];
+                for (const tc of m.tool_calls) {
+                  const fn = tc?.function ?? tc ?? {};
+                  let args: any = {};
+                  if (typeof fn.arguments === "string") {
+                    try { args = JSON.parse(fn.arguments); }
+                    catch { args = { _raw: fn.arguments }; }
+                  } else if (fn.arguments !== undefined) {
+                    args = fn.arguments;
+                  }
+                  tcs.push({ name: fn.name ?? "unknown", arguments: args });
+                }
+                if (tcs.length > 0) entry.tool_calls = tcs;
+              }
+            } else if (role === "tool") {
+              entry.content = extractText(m.content);
+              if (typeof m.tool_call_id === "string" && m.tool_call_id.length > 0) {
+                entry.tool_call_id = m.tool_call_id;
+              }
+            } else {
+              entry.content = extractText(m.content);
+            }
+            out.push(entry);
+          }
+          return out;
+        };
+
         // Extract system message. OpenAI's o1/o3-style reasoning surface
         // (and pi-coding-agent) sends `role:"developer"` instead of
         // `role:"system"` for the same purpose — strict instructions that
@@ -1692,6 +1753,24 @@ async function serve(port: number) {
         }
         if (systemPrompt) genParams.system = systemPrompt;
 
+        // Phase 2: structured tools + messages passed alongside the
+        // legacy prompt/system text. Under HIPFIRE_JINJA_CHAT=1 the
+        // daemon's Jinja path renders `messages` + `tools` through the
+        // model's upstream chat_template (XML tool-call format on
+        // Qwen3.5/3.6 etc.) and ignores `prompt`/`system`. Under
+        // HIPFIRE_JINJA_CHAT=0 (default) the daemon ignores the
+        // structured fields and falls back to the inline-ChatML prompt
+        // + text-rendered tools-block already built above. We send both
+        // shapes so the same OpenAI request works against either
+        // daemon mode without per-client awareness.
+        if (Array.isArray(body.tools) && body.tools.length > 0) {
+          genParams.tools = body.tools;
+        }
+        const structuredMessages = mapMessagesToStructured(messages);
+        if (structuredMessages.length > 0) {
+          genParams.messages = structuredMessages;
+        }
+
         // Parse tool calls from model output: <tool_call>{"name":..., "arguments":...}</tool_call>
         //
         // Defensive against MQ4 quantization drift on structured-token positions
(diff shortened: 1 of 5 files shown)


unverbraucht:
Thanks for the careful per-phase split — the one-commit-per-layer structure made this very pleasant to review. The serde-typed `Message` / `ToolCall` shape is the right call, the env-gate keeps the default safe, and the new TS parser cases on `parse_tool_calls.test.ts` are well-targeted. I ran the unit + parser tests locally and they all pass (161 Rust + 18 TS), and I dumped the embedded `chat_template` from `qwen3.6-27b.mq4` to probe the failure-mode branches that the happy-path tests don't reach.

A few things I'd love to see addressed before this lands, plus some smaller observations.

---

### High priority

**1. `arguments=null` on an assistant `tool_call` silently kills the Jinja render and falls back to Plain.**

`ToolCall.arguments: serde_json::Value` defaults to `Value::Null` under serde when the field is missing. The real qwen3.6 template walks arguments via `tool_call.arguments | items`, which raises on Null:

```
template render: invalid operation: cannot convert value into pairs (in chat:120)
```

I reproduced this against the real embedded template using `JinjaChatFrame::render_messages`. Trigger path through `cli/index.ts:1520-1525`: a client sending OpenAI-shaped `{"function":{"name":"ping","arguments":null}}` (legitimate for zero-arg calls) reaches the daemon as `arguments: Value::Null`.

When this fires, all three dispatch paths (`generate`, `generate_dflash`, `generate_multi`) fall back to `ChatFrame::Plain`, which has no notion of `messages_history` and re-frames only the current `prompt` as a single user turn. The whole tool-call history is silently dropped.

Smallest fix in `prompt_frame.rs`:

```rust
#[derive(Clone, Debug, serde::Serialize, serde::Deserialize)]
pub struct ToolCall {
    pub name: String,
    #[serde(default = "default_args")]
    pub arguments: serde_json::Value,
}

fn default_args() -> serde_json::Value {
    serde_json::Value::Object(Default::default())
}
```

Plus a belt-and-suspenders coercion in the CLI mapper (`cli/index.ts:1523`):

```ts
} else if (fn.arguments !== undefined && fn.arguments !== null) {
  args = fn.arguments;
}
```

Worth a unit test in `prompt_frame.rs` (`render_messages` on an assistant `ToolCall` with `arguments=Null`) — the current `render_messages_with_history_and_tools_includes_assistant_call` always supplies a populated object so this branch isn't covered.

**2. Jinja-failure fallback writes to stderr only — the client never sees what happened.**

All three Jinja-failure sites (`daemon.rs:3382`, `:2158`, `:2768`) log via `eprintln!` and then silently re-frame with `ChatFrame::Plain`. A pipe-fed client (the canonical contract) doesn't see stderr; the model just runs with degraded context.

I verified three additional template-side raises against the real qwen3.6 template that all trigger this silent-degradation path:

| Scenario | Template behavior | Daemon-side effect |
| --- | --- | --- |
| `arguments=null` on assistant tool_call | `tool_call.arguments \| items` raises | Plain fallback, history lost |
| System message in middle of history | `raise_exception('System message must be at the beginning.')` | Plain fallback, system silently dropped |
| Empty `messages` array | `raise_exception('No messages provided.')` | Plain fallback, empty user turn |

The upstream template uses `raise_exception(...)` deliberately to surface bad inputs; on our side those raises are being inverted into "looks fine, but the model gets garbage context."

Smallest fix is a JSONL event before the fallback so clients can see the degradation:

```rust
let _ = writeln!(
    stdout,
    r#"{{"type":"jinja_fallback","id":"{}","reason":"{}"}}"#,
    id, e.to_string().replace('"', "'"),
);
```

A stronger version would emit `"type":"error"` and skip the request when `messages_history.is_some()` — at that point the client has explicitly opted into the structured shape, so falling back to a guess is worse than failing loudly.

---

### Medium priority

**3. Multi-turn under `HIPFIRE_JINJA_CHAT=1` silently drops `messages_history` on `seq_pos > 0`.**

`generate()` line 3331 and `generate_multi()` line 2716 gate Jinja on `m.seq_pos == 0 && m.chat_template.is_some()`. The PR description acknowledges this as Phase-2 scope, but the daemon code itself doesn't surface the limitation: `messages_history` is parsed every call (so format errors are caught) and then silently ignored past turn 0.

This works today only because `cli/index.ts:1586` still emits the legacy `<tool_response>` text shape in `prompt` alongside the new structured `messages` array — Plain catches it. As soon as anyone "cleans up" the CLI to stop sending the legacy text under `HIPFIRE_JINJA_CHAT=1`, every multi-turn tool flow breaks silently.

Easiest mitigation for now: emit a one-shot `"type":"jinja_skipped","reason":"seq_pos>0_multiturn_pending"` event the first time a request supplies `messages_history` while seq_pos>0. That at least gives downstream operators a visible signal before flipping the default in a later PR.

**4. PFlash compression runs and emits events even when Jinja discards `q_tokens`.**

In `generate()` (lines 3248–3302), `q_tokens` is computed and PFlash compression fires *before* the Jinja branch. If Jinja succeeds (line 3380), it re-encodes from the rendered template string and `q_tokens` is thrown away entirely. The `pflash_compressed` event has already been emitted to the client, and `pflash_summary` rides through to the trailing `done.pflash` block — both describing a compression that has zero effect on what the model sees.

The in-source comment says "PFlash compression is bypassed under Jinja," but actually it runs, emits telemetry, then gets overwritten. Operators correlating compression ratios with output quality on the Jinja path will be chasing a phantom.

Either suppress the `pflash_compressed` event when `try_jinja` will succeed, or guard the PFlash branch on `!try_jinja` so the work doesn't happen at all.

---

### Lower priority

**5. `coerceParamValue` numeric coercion is lossy for leading-zero strings.**

`cli/index.ts:1942`:

```ts
if (/^-?\d+(?:\.\d+)?(?:[eE][+-]?\d+)?$/.test(s)) {
  const n = Number(s);
  if (Number.isFinite(n)) return n;
}
```

`"007"` → `7`, `"01234"` → `1234`, `"1e3"` → `1000`. The current tests cover `"42"` and `"3.14"` but no leading-zero or scientific-notation cases.

Minimal fix:

```ts
if (/^-?\d+(?:\.\d+)?(?:[eE][+-]?\d+)?$/.test(s)) {
  const looksLikeIntId = /^0\d/.test(s);  // 01234 — keep as string
  if (!looksLikeIntId) {
    const n = Number(s);
    if (Number.isFinite(n)) return n;
  }
}
```

Add a regression test:

```ts
test("Qwen3.6 XML keeps leading-zero strings as strings", () => {
  const raw = '<function=call><parameter=phone>0415123</parameter></function>';
  const r = parseOneToolCall(raw);
  expect(r!.arguments).toEqual({ phone: "0415123" });
});
```

**6. `ChatFrame::Plain` `panic!` on `Role::System | Role::Tool` is a developer assertion in a `pub` API.**

`prompt_frame.rs:176-179`. Currently only reached from internal tests, but the public `build_multi_turn` would crash the entire daemon process if a future caller wired `messages_history` into Plain without the role filter. Worth converting to `Err` and propagating, or `#[doc(hidden)]`-ing behind a private helper.

**7. `parseToolCalls` outer gate only fires on `<tool_call>` substring — verified safe today, but worth one regression test.**

I verified by dumping the qwen3.6 template that the assistant's tool-call output is always wrapped in `<tool_call>...</tool_call>` (the template's own example prompt shows the wrapper, so the model copies it). The new TS tests exercise `parseOneToolCall` directly on raw `<function=...>` strings — passing those tests doesn't actually validate the end-to-end CLI path because the outer `parseToolCalls` gate would never let those payloads through.

Suggestion: add one test at the `parseToolCalls` (outer) layer with the full `<tool_call>\n<function=foo>\n<parameter=...>\n</parameter>\n</function>\n</tool_call>` shape so a future model that drops the outer wrapper is caught early.

**8. Smoke gate `agentic-gate-jinja-tools.sh` uses fixed sleeps instead of polling for `"type":"loaded"`.**

90s after `load`, 45s after `generate`. Brittle on cold-pagecache or slower-storage hosts (a 22 GB a3b on a 24 GB card with mmap warming can easily exceed 90s). Other gates have the same shape so it's consistent — non-blocking, just flagging since this gate is the only CI signal that the Jinja+tools path actually runs.

Separately: the gate `exit 0` on missing-model silently passes CI on hosts without the fixture. `exit 77` (autotools skipped convention) would let CI distinguish skipped from pass.

---

### Architectural notes (not blockers)

- **No Plain-vs-Jinja byte-identity fixture for a real tokenizer.** The PR description's regression-battery claim is a *behavioral* equivalence (agentic-gate cells match with/without `HIPFIRE_JINJA_CHAT=1`), not a byte-identity claim. Worth a committed fixture (a single rendered prompt + its token sequence under both paths against `qwen3.6-27b.mq4`) before any future PR flips the default. The τ-sensitivity rule in CLAUDE.md is exactly the failure mode this guards against — one stray newline difference is a 17% τ swing.
- **Coherence-gate.sh wasn't run with `HIPFIRE_JINJA_CHAT=1`.** Worth a single `HIPFIRE_JINJA_CHAT=1 ./scripts/coherence-gate.sh` run on qwen3.6-27b before the default flip, capturing tok/s + τ deltas against the locked baselines.
- **`generate()` / `generate_dflash()` / `generate_multi()` share ~40 lines of synthesize-messages + render + fall-back logic.** A `build_prompt_tokens(...)` helper would deduplicate cleanly; not blocking for Phase 3a but worth folding into a Phase-2 refactor.

---

### What I verified locally

- `cargo test --release -p hipfire-runtime --lib` → 161 passed, 0 failed.
- `bun test cli/parse_tool_calls.test.ts` → 18 pass, 0 fail.
- Dumped the embedded `chat_template` from `qwen3.6-27b.mq4` and ran six edge-case scenarios through `JinjaChatFrame::render_messages` against the real template. Scenarios 1 (plain multi-turn), 4 (system+user), 5 (assistant + empty tool_calls), 6 (tool-response tail) render fine. Scenarios 2 (system-not-at-start) and 3 (empty messages) raise as the template intends. Scenario "assistant tool_call with arguments=null" raises — the basis of finding #1.
- Confirmed by template inspection that the assistant's tool-call emission is always `<tool_call>`-wrapped, so the CLI's outer gate behaves correctly today.

I didn't run the agentic / DFlash gates on a hot model — happy to do that on gfx1151 if useful, but the daemon-side findings above don't need GPU to reproduce.

---

Overall this is a really clean piece of work — the phasing is appropriately conservative, the `JinjaChatFrame::render_messages` shape will scale well as more arches come online, and `Strict` undefined is the right default *once* the fallback path surfaces failures to the client. Findings #1 and #2 are the main things I'd want addressed before this lands; the rest can ride a follow-up.

PR #269, opened 2026-05-16
fix(qwen35): guard hipGraph capture against TriAttention eviction stream conflict

aldrouil:
## Summary

Fix a hipGraph capture-stream conflict crash and a silent-garbage-output bug when TriAttention eviction (KV cache compaction) is active on gfx12 (RDNA4). The daemon panics with hipError 906 ("operation would make the legacy stream depend on a capturing blocking stream") during generation with qwen3.6-27b.mq4 on RX 9070 XT.

- [x] `crates/rdna-compute` — new `memcpy_htod_auto()` dispatch helper
- [x] `crates/hipfire-arch-qwen35` — graph capture guard + 11 FA-layer H2D copy replacements

## The Bug (two parts)

**Part 1 — Capture-time crash:** TriAttention eviction compacts the KV cache, setting `compact_offset > 0`. FA layer handlers then update RoPE position buffers via synchronous `gpu.hip.memcpy_htod()` on the legacy/null stream. During HIP graph capture (default-ON for gfx12), this conflicts with the capturing blocking stream → hipError 906.

**Part 2 — Silent garbage output:** Even if the crash were fixed, during *graph replay* the `stream_write_value32` path updates pos_buf before launch, but the captured H2D copy inside FA layers then overwrites it with the stale position from capture time. RoPE reads wrong positions → garbled output that passes no error check.

### 1. Capture-aware H2D memcpy (`crates/rdna-compute/src/dispatch.rs`)

Added `memcpy_htod_auto()` — routes to `memcpy_htod_async` on the active (capturing) stream when `capture_mode == true`, falling back to sync `memcpy_htod` otherwise. Prevents hipError 906 during graph capture. Replaced 11 occurrences of `gpu.hip.memcpy_htod(&s.pos_buf, ...)` across FA layer handlers in both MQ4 and MQ3 Lloyd branches.

### 2. Gate graph capture/replay on eviction status (`crates/hipfire-arch-qwen35/src/qwen35.rs`)

Added a `no_compact` guard: skip graph capture/replay when `kv_cache.compact_offset > 0`, falling back to direct execution where sync H2D copies work correctly. This is the real fix for Part 2 (stale positions during replay). Commit 1 is defensive for any future code paths doing H2D during capture.

**Total: ~40 lines changed across 2 files.**

## Test plan

- [x] Model loads and generates without panic on gfx1201 with TriAttention sidecar auto-attached
- [x] Graph replay path still activates when `compact_offset == 0` (no perf regression in normal operation)
- [ ] `./scripts/coherence-gate-dflash.sh` — not yet tested (no RDNA3 card available; tested on gfx1201 only)

## Open questions

1. **Perf impact of disabling graphs during eviction:** graph replay gives ~0.6–0.7% decode win on gfx11 and ~2.4–2.7% on gfx12. Disabling it when `compact_offset > 0` loses this during active eviction periods. Future work could compute absolute RoPE positions on-device, allowing graph capture to coexist with eviction.
2. **Other archs:** tested only on gfx1201. gfx11 behavior may differ — graph capture is also ON by default there but the mechanics vary slightly. Worth verifying on RDNA3.
3. **Prefill path:** `forward_prefill_batch` has similar FA-layer code paths with compact_offset H2D copies. Prefill doesn't currently use graph capture, but worth auditing if someone adds it later.

### DIFF
--- a/crates/hipfire-arch-qwen35/src/qwen35.rs
+++ b/crates/hipfire-arch-qwen35/src/qwen35.rs
@@ -3053,13 +3053,19 @@ pub fn forward_scratch(
         _ => panic!("unsupported embedding format"),
     }
 
-    if use_graph && gpu.graph_exec.is_some() {
+    // Compact offset (TriAttention eviction) breaks graph capture/replay:
+    // during replay, stream_write_value32 updates pos_buf but the captured
+    // H2D copy of (pos + compact_offset) inside FA layers overwrites it,
+    // causing RoPE to read stale positions. Fall back to direct execution.
+    let no_compact = kv_cache.compact_offset == 0;
+
+    if use_graph && gpu.graph_exec.is_some() && no_compact {
         // ── Graph replay path ──
         // Update pos_buf on the device via stream write (no host→device copy).
         let stream = gpu.active_stream.as_ref().unwrap();
         gpu.hip.stream_write_value32(stream, &scratch.pos_buf, pos as u32, 0)?;
         gpu.graph_launch()?;
-    } else if use_graph && gpu.graph_exec.is_none() {
+    } else if use_graph && gpu.graph_exec.is_none() && no_compact {
         let pos_i32 = pos as i32;
         if !gpu.ar_forward_warmed_up {
             // ── Warmup: run direct so kernel JIT and lazy scratch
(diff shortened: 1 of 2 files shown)


Kaden-Schutt (approved):
Approving for the hipGraph + TriAttention/CASK eviction stream-conflict fix.

Smoke coverage I ran on gfx1100:
- Base `origin/master` fails after forced eviction with hipError 906; this PR passes the same forced-eviction graph smoke.
- Verified the before/after path for `--kv-mode asym3`, `--kv-mode q8`, and `--kv-mode fwht4` with `HIPFIRE_GRAPH=1`, 9B target, CASK/TriAttention sidecar, forced `compact_offset > 0`.
- `cargo check` passed for `rdna-compute`, `hipfire-arch-qwen35`, and the `dflash_spec_demo` example surface.
- `./scripts/coherence-gate-dflash.sh --fast` passed on 27B DFlash prose/code.

Follow-up: DDTree + CASK/TriAttention eviction still has a separate hidden-state compaction issue (`target_hidden_host` is not compacted when `position` is reduced). I opened #272 to track that separately; it should not block merging this graph stream-conflict fix.

PR #283, opened 2026-05-17
[codex] fix-forward MQ4 cleanup from #277

### LINKED ISSUE #208: gfx906 MQ4/HFQ4 cleanup — 11 outstanding items (parallel of MQ6 fixes)
## Summary

Three-way review (claude + gemini) of the MQ6 audit branch's HFQ4/MQ4 sibling found that several MQ6 issues fixed in commit `5ea9050` also affect MQ4 — pre-existing in master. None are BLOCKERs (correctness is unaffected). All are mechanical cleanups parallel to fixes that just landed for MQ6.

Issue #207 covers MoE kernel optimization gaps (FFN body wave32-on-gfx906, dp4a port, MMQ port, etc.). This issue covers MQ4-specific dispatcher / build-infra cleanup that #207 doesn't touch.

Full reviews:
- [`gfx906_mq4_review_claude.md`](https://github.com/Kaden-Schutt/hipfire/blob/master/gfx906_mq4_review_claude.md)
- [`gfx906_mq4_review_gemini.md`](https://github.com/Kaden-Schutt/hipfire/blob/master/gfx906_mq4_review_gemini.md)

## TL;DR

**0 BLOCKERs. 5 SCOPE-level fixes + 6 NIT items.** ~70 LOC code + ~15 LOC docs in a single follow-up commit. Both reviewers (claude + gemini) reached identical classifications.

### 🟠 SCOPE-1 — `gemm_hfq4g256_mmq_set_gfx906` bytes calc missing X read

**File:** `crates/rdna-compute/src/dispatch.rs:7445-7447`

```rust
// Current:
let bytes = crate::profile::gemv_hfq4g256_bytes(m, k)
    + batch_size * m * 4;       // ← missing + batch_size * k for X read

// Should be:
let bytes = crate::profile::gemv_hfq4g256_bytes(m, k)
    + batch_size * k            // X read (Q8_1)
    + batch_size * m * 4;       // Y write (set semantics)
```

The residual sibling at line 7342-7344 correctly counts the X read; the set sibling does not. Identical asymmetry to the MQ6 B-SET-BYTES bug fixed in `5ea9050`. Profiling-only impact (~2% under-report on M=3584 K=4096 B=128); no functional bug.

### 🟠 SCOPE-2 — `should_use_mmq` env reads uncached

**File:** `crates/rdna-compute/src/dispatch.rs:332-364`

`should_use_mmq` reads `HIPFIRE_MMQ` (line 336) and `HIPFIRE_MMQ_MIN_BATCH` (line 357) on every invocation. Called from at least 9 dispatch sites (HFQ4 + HFQ6 paths). On 27B prefill, ~288+ syscalls per chunk.

**Fix:** wrap in `OnceLock` matching `gemv_dp4a_enabled`'s pattern. ~20 LOC.

### 🟠 SCOPE-3 — `_mmq_set_gfx906` no defensive gating

**File:** `crates/rdna-compute/src/dispatch.rs:7376+`

`gemm_hfq4g256_mmq_set_gfx906` accepts any batch size with the simple staircase, no `debug_assert!` guarding the caller-must-gate contract. The MQ6 sibling now has this assert (added in `5ea9050`). Apply the same to MQ4:

```rust
debug_assert!(
    should_use_mmq(&self.arch, batch_size) || self.capture_mode,
    "_mmq_set_gfx906 called at non-winning B={} — caller must gate first",
    batch_size,
);
```

### 🟠 SCOPE-4 — Non-residual `gemv_hfq4g256` no gfx906 wave64 path

**File:** `crates/rdna-compute/src/kernels.rs:457-481`

```rust
match arch {
    "gfx1030" | "gfx1031" => RDNA2-specific variants,
    "gfx1100" | "gfx1101" | "gfx1102" => RDNA3 variant,
    _ => (GEMV_HFQ4G256_SRC, "gemv_hfq4g256"),  // ← gfx906 lands here (wave32)
}
```

gfx906 falls to the gfx1010 baseline wave32 kernel for non-residual GEMV. **Rarely called** — most decode/prefill goes through residual or batched paths which have the proper gfx906 fast paths. Affects some lm_head and debug paths.

**Fix:** add a gfx906 arm selecting a wave64 source. Could share kernel code with `gemv_hfq4g256_residual_wave64.hip` minus the `+=` semantics. Lower priority because the path is rare.

### 🟠 SCOPE-5 — 5 MQ4-only diagnostic env vars uncached

**File:** `crates/rdna-compute/src/dispatch.rs:6778-6967` (multiple sites)

The MQ4 residual MMQ dispatcher carries an extensive bisect/dump infrastructure that the MQ6 dispatcher does not have. All read on every dispatch:

| Env var | Site | Purpose |
|---|---|---|
| `HIPFIRE_MMQ_DUMP` | `dispatch.rs:6791` | dumps inputs/outputs of Nth call to `/tmp/` |
| `HIPFIRE_MMQ_K_FILTER` | `dispatch.rs:6911` | restricts MMQ to specific K values |
| `HIPFIRE_MMQ_CALL_FILTER` | `dispatch.rs:6917` | call-index range gate |
| `HIPFIRE_MMQ_LAYER_FILTER` | `dispatch.rs:6921` | layer-index range gate |
| `HIPFIRE_MMQ_TRACE` | `dispatch.rs:6947` | per-call dispatch trace |
| `HIPFIRE_WO_MMQ` | `dispatch.rs:6966` | force MMQ for the wo path |
| `HIPFIRE_FP16` | `dispatch.rs:3647, 4079, 4468, 6773` | force FP16 fallback |

On 27B prefill at B=128: ~960 syscalls per chunk just on these.

**Fix:** OnceLock-cache each. ~30 LOC. The infrastructure is still useful for future regression bisects — keep the env var surface, just cache the reads.

### 🟡 NIT-1 — A-GUARD-RATIONALE comment

**File:** `crates/rdna-compute/src/dispatch.rs:7889` (approx)

The `&& !self.capture_mode` guard on `gemm_hfq4g256_dp4a` says: "Quantize launch must be reachable from the captured graph or pre-baked." This rationale is wrong — `ensure_q8_1_mmq_x` calls `launch_maybe_blob` which IS recorded into the captured graph (validated empirically by Phase B.2 fa8785b lifting the equivalent guard for MMQ paths and seeing bit-identical DFlash output).

The actual concern is that `ensure_q8_1_mmq_x` may fire `hipMalloc` or `ensure_kernel`/JIT on first use; both are first-use-only, so capture is safe after warmup. Same wrong rationale as the MQ6 PR2 case (now corrected on the audit branch).

**Fix:** rewrite the guard's inline comment to reflect the actual invariant. Pure docs.

### 🟡 NIT-2 — A-PREFETCH-VGPR (HFQ4 prefetch kernel VGPR not measured)

**File:** `kernels/src/gemv_hfq4g256_residual_wave64_prefetch.hip:21-23`

Comment claims "should land around 41-44 VGPR" — but `docs/perf-checkpoints/2026-05-05-gfx906-decode-investigation.md`'s VGPR table only measured the BASE wave64 kernel (29 VGPR), not the prefetched variant. Same forensic gap as the MQ6 A-PREFETCH-VGPR finding.

**Fix:** rocprof `--stats` pass on the prefetched variant; update the comment with the real measurement.

### 🟡 NIT-3 — Misleading `__shfl_down` comment in HFQ4 wave64 GEMV

**File:** `kernels/src/gemv_hfq4g256_residual_wave64.hip` (around the `for (int offset = 16; offset > 0; offset >>= 1)` block)

Comment claims "the `__shfl_down` chain stays within the 32-lane warp" — factually wrong. `__shfl_down` defaults to `width=warpSize=64` on wave64 archs; `offset=16` reads from `lane+16` which crosses the 32-lane boundary. The math is still correct because lanes 0 and 32 each write their own row's output and only need their own warp-half's reduction. But adding `width=32` to "match the comment" would BREAK the kernel.

**Fix:** rephrase same as MQ6 H6 fix: "lanes 16-31 read cross-warp but don't write output — do NOT add `width=32`."

### 🟡 NIT-4 — MMQ body OOB row-clamping invariant doc gap

**File:** `kernels/src/gemm_hfq4g256_residual_mmq_gfx906_body.cuh`

The HFQ4 MMQ body has the same `row = (row0 + i < M) ? (row0 + i) : (M - 1)` clamp pattern as HFQ6 (lines 83, 99). The same `_full_*` requires `M%128==0` invariant. **The HFQ4 body has no equivalent doc block** — only LDS-layout docs.

**Fix:** copy the doc block from HFQ6's body.cuh (added in `5ea9050`) into HFQ4's body.cuh, with `200` → `136` byte references and HFQ4-specific quirks noted (signed nibbles + zp_eff shift).

### 🟡 NIT-5 — `gp` pointer recomputed 4× per thread per window

**File:** `kernels/src/gemm_hfq4g256_residual_mmq_gfx906_body.cuh:99`

Same loop pattern as HFQ6: `gp = A + ((long long)row * groups_per_row + kg) * 136` recomputed inside the per-pair loop. Compiler may or may not hoist. Lower priority — multiply is cheap in flight.

**Fix:** explicit row-outer loop. ~5 LOC kernel restructure.

### 🟡 NIT-6 — `block_q8_1_mmq` struct + `QK8_1` macro duplicated

**Files:** ~8 kernel files duplicate the struct; ~7 files duplicate the `#define QK8_1 32` macro.

ODR violation risk if a definition drifts (e.g., someone changes block size in one file). Pre-existing pattern that started with HFQ4 (Phase B.1) and was copied into HFQ6 (Phase B.2).

**Fix:** factor into a shared header `kernels/src/q8_1_mmq.cuh` and `#include` from each kernel file. Affects both MQ4 and MQ6.

### 🟠 INFRA-1 — `kernels/compiled/gfx906/` is empty

```
$ ls kernels/compiled/gfx906/ | wc -l       # 0
$ ls kernels/compiled/gfx1010/ | wc -l      # 398
$ ls kernels/compiled/gfx1030/ | wc -l      # 399
$ ls kernels/compiled/gfx1100/ | wc -l      # 435
$ ls kernels/compiled/gfx1200/ | wc -l      # 421
$ ls kernels/compiled/gfx1201/ | wc -l      # 418
```

**Every** gfx906 kernel (MoE and non-MoE, including all the MQ4 paths above) JIT-compiles at runtime via `ensure_kernel()`'s hipcc invocation on first use. The result IS cached to `~/.hipfire_kernels/gfx906/<arch>/` after first use, so steady-state is fine — but cold-start adds ~1-3 seconds per unique kernel. On A3B with ~50 distinct kernel names that's ~1-2 min of first-load latency.

**Fix:** populate the in-tree `kernels/compiled/gfx906/` during release builds. The build infrastructure to compile per-arch kernels exists for the other 5 archs; gfx906 just isn't wired in. Likely a CI/release-script gap, not a code change. **Highest user-visible impact for the smallest effort.**

This was identified in the MoE audit (originally surfaced by glm5) but applies to ALL gfx906 kernels, not just MoE. Cross-references issue #207 but is broader in scope.

### 🟡 FORENSIC-1 — MQ4 B=8 / B=24 perf unmeasured

**Context:** the MQ6 work measured B=8 at 0.86× regression on HFQ6 weights and chose to gate it off via `hfq6_mmq_winning_size` (only B=16 and B≥32 enter MMQ). MQ4 has no equivalent gate — `should_use_mmq` simply requires `B≥8`, so all B≥8 enter MMQ on gfx906.

**Risk:** if MQ4 B=8 also regresses ~15% (analogous to MQ6), the current "MMQ-on at all B≥8" routing leaves perf on the table. If it doesn't regress, MQ4's routing is correct and MQ6's gate is over-conservative. The justification at `dispatch.rs:343-352` cites `B ∈ [12, 14]` measurements; B=8 specifically was not measured.

**Action:** re-run the per-batch sweep for MQ4 at B ∈ {8, 12, 14, 16, 24, 32, 40, 48, 56, 64} on M=3584 K=4096. If any size regresses, add an `hfq4_mmq_winning_size` helper matching the HFQ6 pattern. ~15 min bench work; can be its own follow-up.

## Out of scope for this issue

- **MoE kernel optimization** — covered by #207 (7 gaps identified).
- **`test_hfq4_mmq.rs` correctness test** — exists for HFQ6 (14 tests), missing for HFQ4. Worth doing eventually (~200 LOC mechanical port from `test_hfq6_mmq.rs`) but defer until kernel changes warrant the safety net.
- **MoE matchers missing `HFQ4G256`** (`moe_ffn_all_mq4` only checks MQ4G256) — defer until an HFQ4 MoE quant ships.
- **Mixed-dtype paths** — dead code for current models.

## Suggested commit shape

Single follow-up commit on top of `5ea9050`:

1. SCOPE-1 (set-bytes, 3 LOC)
2. SCOPE-2 (`should_use_mmq` cache, ~20 LOC)
3. SCOPE-3 (defensive assert, 3 LOC)
4. SCOPE-5 (5 diag env vars cached, ~30 LOC)
5. NIT-1, NIT-3 (docs comments, ~10 LOC)
6. NIT-4 (body.cuh doc block, ~30 LOC of comments)

Total: ~70 LOC code + ~40 LOC docs. Defer SCOPE-4 (rare path), NIT-2 (forensic), NIT-5 (perf-marginal), NIT-6 (cross-cutting refactor), INFRA-1 (build-infra), FORENSIC-1 (bench).

INFRA-1 should be its own infrastructure issue/PR since it affects all gfx906 kernels not just MQ4.

Kaden-Schutt:
## Summary

Fix-forward of @unverbraucht's #277 onto current `master` after #277 became dirty against the newer gfx906/HFQ4 parity work from #281.

This keeps the cleanup intent from #277 while preserving the current `master` dispatch behavior:

- cache hot-path env reads for `HIPFIRE_MMQ`, `HIPFIRE_MMQ_MIN_BATCH`, `HIPFIRE_FP16`, `HIPFIRE_WO_MMQ`, and `HIPFIRE_LM_HEAD_WMMA`
- keep current gfx906 `B >= 8` MMQ routing and #281's `mmq_screen_rejected`/dp4a fallback behavior
- add the gfx906 set-mode MMQ defensive `debug_assert!`
- fix the gfx906 set-mode MMQ profile byte accounting to include the Q8_1 X read
- carry over the HFQ4 gfx906 kernel invariants docs and wave64 shuffle comment clarification

Supersedes #277 after conflict resolution. Closes #208's cleanup-track subset represented by #277 if merged.

## Validation

- `git diff --check`
- `cargo check -p rdna-compute`
- `cargo check --tests -p rdna-compute`
- `MAX_TOKENS=2 HIPFIRE_VERIFY_GRAPH=0 cargo run --release --features deltanet --example greedy_dump -p hipfire-runtime -- ~/.hipfire/models/qwen3.5-0.8b.mq4 /tmp/pr277-fixforward-greedy-default.txt "Say OK."`
- `MAX_TOKENS=2 HIPFIRE_VERIFY_GRAPH=0 HIPFIRE_FP16=0 HIPFIRE_LM_HEAD_WMMA=0 cargo run --release --features deltanet --example greedy_dump -p hipfire-runtime -- ~/.hipfire/models/qwen3.5-0.8b.mq4 /tmp/pr277-fixforward-greedy-fp16off.txt "Say OK."`

Both runtime smokes completed on local `gfx1100`, loaded `qwen3.5-0.8b.mq4`, captured `338` graph blobs, and ended with `done`.

### DIFF
--- a/crates/rdna-compute/src/dispatch.rs
+++ b/crates/rdna-compute/src/dispatch.rs
@@ -288,14 +288,18 @@ fn has_mmq_dp4a_or_wmma(arch: &str) -> bool {
 ///   `0` / `off`            — force MMQ off (debug / regression bisect)
 ///   `1` / `on`             — force MMQ on at every batch (legacy behavior)
 ///   `auto` / unset / other — auto-route by batch_size threshold (default)
+///
+/// Both env reads (`HIPFIRE_MMQ`, `HIPFIRE_MMQ_MIN_BATCH`) cache via
+/// `OnceLock` — no per-dispatch syscalls on the hot path (called from
+/// 9+ dispatch sites, ~288+ syscalls/chunk on 27B prefill otherwise).
 fn should_use_mmq(arch: &str, batch_size: usize) -> bool {
     if !has_mmq_dp4a_or_wmma(arch) {
         return false;
     }
-    match std::env::var("HIPFIRE_MMQ").ok().as_deref() {
-        Some("0") | Some("off") => false,
-        Some("1") | Some("on") => true,
-        _ => {
+    match mmq_env_override() {
+        Some(false) => false,
+        Some(true) => true,
+        None => {
             // Per-arch default min_batch:
             //   gfx906: 8 — empirically validated for both prefill (pp512
             //     within noise of min_batch=16) and DFlash 27B verify
@@ -313,13 +317,61 @@ fn should_use_mmq(arch: &str, batch_size: usize) -> bool {
             //     than MMQ at small batches; flip only when MMQ amortization
             //     dominates.
             let arch_min_batch: usize = if arch == "gfx906" { 8 } else { 256 };
-            let min_batch = std::env::var("HIPFIRE_MMQ_MIN_BATCH")
-                .ok()
-                .and_then(|s| s.parse::<usize>().ok())
-                .unwrap_or(arch_min_batch);
+            let min_batch = mmq_min_batch_override().unwrap_or(arch_min_batch);
             batch_size >= min_batch
         }
     }
+}
+
+/// Cached env-var read for `HIPFIRE_MMQ`. `Some(true)` = force on,
+/// `Some(false)` = force off, `None` = auto.
+fn mmq_env_override() -> Option<bool> {
+    static CACHE: OnceLock<Option<bool>> = OnceLock::new();
+    *CACHE.get_or_init(|| {
+        match std::env::var("HIPFIRE_MMQ").ok().as_deref() {
+            Some("0") | Some("off") => Some(false),
+            Some("1") | Some("on") => Some(true),
+            _ => None,
+        }
+    })
+}
+
+/// Cached env-var read for `HIPFIRE_MMQ_MIN_BATCH` (per-arch min-batch override).
+fn mmq_min_batch_override() -> Option<usize> {
+    static CACHE: OnceLock<Option<usize>> = OnceLock::new();
+    *CACHE.get_or_init(|| {
+        std::env::var("HIPFIRE_MMQ_MIN_BATCH")
+            .ok()
+            .and_then(|s| s.parse::<usize>().ok())
+    })
+}
+
+/// Cached env-var read for `HIPFIRE_FP16=0` (disable FP16 fast paths).
+/// Called from 12+ dispatch sites on the prefill hot path; uncached read
+/// is one syscall + String alloc per dispatch.
+fn fp16_disabled() -> bool {
+    static CACHE: OnceLock<bool> = OnceLock::new();
+    *CACHE.get_or_init(|| {
+        std::env::var("HIPFIRE_FP16").map_or(false, |v| v == "0")
+    })
+}
+
+/// Cached env-var read for `HIPFIRE_WO_MMQ=1` (opt-in MMQ for the wo path
+/// on RDNA3+/RDNA3.5 while the tiled path is validated).
+fn wo_mmq_enabled() -> bool {
+    static CACHE: OnceLock<bool> = OnceLock::new();
+    *CACHE.get_or_init(|| {
+        std::env::var("HIPFIRE_WO_MMQ").ok().as_deref() == Some("1")
+    })
+}
+
+/// Cached env-var read for `HIPFIRE_LM_HEAD_WMMA=0` (disable the gfx11
+/// LM-head WMMA fast path).
+fn lm_head_wmma_disabled() -> bool {
+    static CACHE: OnceLock<bool> = OnceLock::new();
+    *CACHE.get_or_init(|| {
+        std::env::var("HIPFIRE_LM_HEAD_WMMA").map_or(false, |v| v == "0")
+    })
 }
 
 /// Tensor stored on the GPU. Tracks shape and element type.
@@ -4901,7 +4953,7 @@ impl Gpu {
             }
         }
         // Fast paths for prefill (batch_size > 1). Disable with HIPFIRE_FP16=0.
-        if batch_size > 1 && !std::env::var("HIPFIRE_FP16").map_or(false, |v| v == "0") {
+        if batch_size > 1 && !fp16_disabled() {
             // Wave64 FP16 hybrid — best of both worlds for gfx906 (MI50).
             if is_gcn5_wave64(&self.arch) {
                 // gfx906 dp4a MMQ split: qkv + z route through the new MMQ
@@ -5364,7 +5416,7 @@ impl Gpu {
             }
         }
         // Fast paths for prefill (batch_size > 1). Disable with HIPFIRE_FP16=0.
-        if batch_size > 1 && !std::env::var("HIPFIRE_FP16").map_or(false, |v| v == "0") {
+        if batch_size > 1 && !fp16_disabled() {
             // Wave64 FP16 hybrid — best of both worlds for gfx906 (MI50).
             if is_gcn5_wave64(&self.arch) {
                 // gfx906 dp4a MMQ: route q+k+v through the new MMQ kernel.
@@ -5766,7 +5818,7 @@ impl Gpu {
             }
         }
         // Fast paths for prefill (batch_size > 1). Disable with HIPFIRE_FP16=0.
-        if batch_size > 1 && !std::env::var("HIPFIRE_FP16").map_or(false, |v| v == "0") {
+        if batch_size > 1 && !fp16_disabled() {
             // gfx906 dp4a MMQ — default-on at batch_size ≥ 8 (per
             // should_use_mmq's gfx906 default). Quantize X once, screen
             // both weights, dispatch MMQ for each in set mode (add=0).
@@ -9071,7 +9123,7 @@ impl Gpu {
         }
 
         // Fast paths for prefill (batch_size > 1). Disable with HIPFIRE_FP16=0.
-        if batch_size > 1 && !std::env::var("HIPFIRE_FP16").map_or(false, |v| v == "0") {
+        if batch_size > 1 && !fp16_disabled() {
             // gfx906 dp4a MMQ residual path — default-on at batch ≥ 8 per
             // should_use_mmq's gfx906 default. Distinguishes two reasons
             // MMQ might NOT fire:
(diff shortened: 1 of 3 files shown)


Kaden-Schutt:
Additional prompt smoke on local `gfx1100` with `qwen3.5-0.8b.mq4`:

Command shape: `MAX_TOKENS=16 HIPFIRE_VERIFY_GRAPH=0 ./target/release/examples/greedy_dump ...`, run once with default env and once with `HIPFIRE_FP16=0 HIPFIRE_LM_HEAD_WMMA=0` for each prompt.

Prompts:

- `Say OK, then name one color.`
- `Write a Python function add(a, b).`
- `In one sentence, explain why cache invalidation is tricky.`

Result:

- all 6 runs completed with `[hipGraph] captured 338 blobs, instantiated` and `done`
- default/env-disabled token streams were identical on the Python prompt; the other two diverged slightly, as expected when the FP16/LM-head fast paths are intentionally disabled
- decoded previews were coherent, e.g. `Okay, I'm ready to help...`, `Okay, I'll say "OK" and then name one color.`, and `Here's a thinking process that leads to the suggested Python function...`

No crash, empty output, raw corruption, or obvious token-attractor behavior observed in this prompt smoke.

PR #322, opened 2026-05-23
feat(arch-deepseek4): add DeepSeek V4 Flash support (arch_id=9)

nwoolmer:
Bring DeepSeek V4 Flash onto the canonical upstream serving path. Targets the production deepseek-v4-flash.mq2lloyd build: MQ2-Lloyd routed experts (96% of bytes), Q8_0 attention + router + shared-expert weights, F16 compressor / indexer / Hyper-Connections / norms. KV cache at runtime is F32 (raw SWA window + compressed-KV indexer). Single binary, no Python in the hot path.

What's added:
- hipfire-arch-deepseek4 crate (~10,700 lines): Architecture trait impl, forward pass (Hyper-Connections, compressed-KV indexer, tail-only RoPE, Q/O-LoRA, raw SWA), MTP speculative-decode head, deepseek4_chat example.
- 72 new HIP kernels (DeepSeek V4 attention SWA + indexer + Hyper-Connections + compressor + MQ2-Lloyd MoE GEMVs + tail-only YaRN RoPE + glue).
- 73 new Gpu dispatch methods + arch-aware gemm_q8_0_wmma (gfx12 / RDNA3+).
- Daemon arm (arch_id=9, generate_deepseek4) with batched prefill + MTP spec-decode parity with deepseek4_chat.
- Quantizer: --format deepseek4-q8-mtp and variants, --include-prefix mtp., deepseek_v4 model_type to arch_id 9.
- hfq_split tool (partition .hfq by tensor prefix, for MTP sidecar).
- Gates: DeepSeek V4 arms in coherence-gate.sh + new coherence-gate-deepseek4-mtp.sh sibling; pre-commit HOTSPOT regex covers DeepSeek V4 paths.

HIPFIRE_DEEPSEEK4_* env vars follow the project's arch-prefix convention; routed MoE, expert upload, and deterministic MoE-down are default-on.

Perf (gfx1151 / Strix Halo / Radeon 8060S, deepseek-v4-flash.mq2lloyd):
- Plain decode (461-tok prompt, B=64): 40.42 PP / 13.61 TG tok/s
- Path B grouped MoE, B=512: 51.79 PP tok/s (+28% over scalar K4)
- MTP K=2 code prompt: 17.51 TG tok/s, 72.6% accept (+29% over plain)

Gates pass: cargo check + clippy clean for deepseek4 crate; coherence-gate.sh DeepSeek V4 cases OK; coherence-gate-deepseek4-mtp.sh --fast OK.

See docs/deepseek4-pr-body.md for the full PR description and follow-up items (gfx12 WMMA variants, RDNA1/2 fallback, VL, multi-GPU pipeline).

## Post-PR follow-ups (31 commits since base scaffold)

The base V4F scaffold (`0e6066b`) lands a forward pass and `deepseek4_chat`. Driving real pi-coding-agent sessions through the daemon arm surfaced an end-to-end layer of work — chat template, tool-calling, prompt cache, and multi-turn correctness. All the follow-up commits target what the daemon needs to be a usable OpenAI-compat target for a coding agent.

**Chat template + sampler parity** (`ee8603a`, `9f6c57e`)
- Port HF chat template + temp/top-p sampler into the daemon arm. Tokenizer now treats every `added_tokens` entry as atomic, not just specials, so DSML markers round-trip cleanly.

**DSML tool-calling** (`351ac54`, `5bf6c35`, `3408f87`, `efef672`, plus the `1fa6058`/`09922af` exploration that was reverted via `e13549e`/`0d62ad5`)
- Streaming DSML parser: `<think>…</think>` → reasoning events, `<｜DSML｜tool_calls>…</｜DSML｜tool_calls>` → structured ToolCalls. Multi-turn history render. Recovery for V4F MQ2-Lloyd's emitted `<｜DSML｜tool name="X">` variant (deterministic divergence from HF's `invoke` form). DSML system-prompt block byte-aligned with HF's reference encoder + load-bearing `\n\n` prefix on tools-only requests.
- Parser hardened later: recover from V4F's bare `</｜DSML｜>` truncated close; continue past malformed parameters with missing `string="..."` attr; tag-bounded attr search so a broken param can't grab the next param's value. New unit tests cover both recovery paths and the think-mode parser priming.

**Grammar-guided decoder** (`8bf98b0`, `15c68b0`, `6f1eb55`)
- Sample-time logit mask over a DSML state machine forces V4F into valid tool-call structure. Schema `required` blocks invoke-close until all required params are emitted. The `Out` state constrains once the atomic `｜DSML｜` token lands, so the model can't drift into `<｜DSML｜tool_cbl>` / `<｜DSML｜calling>` near-misses. `Free` mode in `Out` / `InParamBody` short-circuits the mask scan.
- StreamParser primed to `InThink` for High/Max think modes — the prompt ends with `<think>`, so the model's first emit is reasoning body, not the `<think>` tag itself.

**Compressor correctness** (`5d5df20`, `2233542`)
- APE (Absolute Position Encoding) wired into the compressor in BOTH sequential and batched paths — missing APE was the root cause of long-context recall corruption (`mariozechner` → `marioze`, `v20.19.6` → `v20.19.20`).
- Compressor RoPE position default switched to `"start"` to match the antirez/ds4 reference. Capture-path and graph-replay paths now share a single `fill_pos_array_host` helper — they previously carried independently-edited copies of the loop with different defaults, so the captured graph wrote one `main_rope_pos` at capture time and a different one at replay time.

**OpenAI-compat layer** (`2cab2ce`, `c17d6c2`, `9112ec0`, `fe1c753`)
- Streaming SSE: `finish_reason: "tool_calls"` on structured tool-call emission. Daemon now emits an authoritative `finish_reason ∈ {stop, length, tool_calls}` in the `done` envelope; CLI streaming + non-streaming both honor it (clamped to OpenAI-valid set with legacy "stop" fallback).
- Non-streaming path now accumulates `reasoning` events into `message.reasoning_content` — previously dropped silently on every think-mode V4F turn.
- All daemon JSONL emits (`emit_error_with_id`, `emit_committed_event`, V4F spec-decode + Qwen2 token emit) go through `serde_json::json!` so user-supplied `id` and BPE-decoded `text` can't corrupt the stream with embedded `"` / `\` / control chars. Removed the now-unused `json_escape` helper.
- Diagnostic env vars `HIPFIRE_DUMP_REQUEST=1` (serve) and `HIPFIRE_DEEPSEEK4_DUMP_PROMPT=1` (daemon) write each request body / each decoded prompt to `/tmp/` for offline debugging.

**Multi-turn history rendering** (`19534a3`, `a689fc1`, `95a4093`, `55353c9`, `1c2917c`, `bcb11be`)
- Serve's V4F-conditional path stops double-rendering history into the prompt: live user text only, no Hermes `<tools>` block, no `extractContent(null)` → `"null"` literal. The arch gate now applies AFTER the reload-and-update of `currentArch` so a model swap takes effect before the gate fires.
- Daemon emits `</think>` after `<｜Assistant｜>` on replayed turns. Tool-result format matches the reference imatrix renderer: XML-escape `&<>` in the body so embedded markup can't terminate `</tool_result>` early, and stack consecutive `Role::Tool` messages inside one `<｜User｜>` turn (matches OpenAI's parallel-tool-call flow).
- `DeepseekV4State::reset()` resets `n_tokens` on the daemon's reset handler. Without this, `forward_prefill_batch_chunked` ran with `start_pos = sum_of_prior_prefill_lengths` and wrote the new conversation AFTER the prior one's KV slots — observed as path-recall corruption (`CLion` for `CLionProjects`, `/home/n/` for `/home/nick/`).

**Prefix cache** (`66151ac`, `2f67d1b`, `1a40170`, `2556d5a`)
- LCP-based prefix cache: serve compares the freshly-tokenized prompt against `m.conversation_tokens` (resident KV identity), skips prefill for the matching prefix, returns `usage.prompt_tokens_details.cached_tokens` per the OpenAI cached-tokens convention. SWA-safety analysis in the daemon comment — partial LCP only kicks in when `prompt_ids.len() >= prior.len()`.
- Per-turn token cache eliminates BPE re-encode drift: assistant turns the daemon emitted get cached under a stable fingerprint of `(content_trimmed, tool_calls_canonical)`; history render replays the verbatim emitted tokens instead of re-encoding through Jinja → tokenizer.encode. Fingerprint canonicalizes tool-call args via recursive key-sorted JSON so insertion-order variance doesn't drop cache hit rate. Fingerprint ignores `content` when `tool_calls` is non-empty (OpenAI clients send `content: null` on tool-call turns even when the model emitted prose). Empty-payload emissions don't get cached.
- Decoded vocab cache on `LoadedModel` as `Arc<Vec<String>>` — lazy-built once per model on first tool-using request. Previously rebuilt all ~129k `tokenizer.decode(&[id])` calls per request, adding tens of ms of overhead to every tool-using turn.

**HIP graph staleness across resets** (`0e12701`)
- `DeepseekV4State::reset()` also clears `ar_forward_warmed_up` so each new session retraces warmup → capture → replay against fresh state shape.
- `gpu.invalidate_graph_state()` is called from all three reset paths: daemon's `"reset"` handler, `unload_model` (V4F arm now actually frees state + weights + PBS), and the in-request `lcp == 0` soft reset. Without invalidation, the second decode of a fresh-context turn replayed a stale captured graph and crashed with `"download logits (graph path): HipError code 700: illegal memory access"`.
- New `free_gpu` impls on `DeepseekV4State`, `DeepseekV4LayerWeights`, `DeepseekV4Weights`, `PrefillBatchScratch` — V4F was missing from `unload_model` (only Qwen35 / LLaMA / Qwen2 had been wired), leaking tens of GB of routed-expert blobs until daemon exit.

## Summary

Adds DeepSeek V4 Flash as `arch_id=9` — new arch crate `hipfire-arch-deepseek4` (~10.7k lines) + 72 HIP kernels + dispatch wiring + daemon arm + quantizer + MTP sidecar tooling. Validated end-to-end on gfx1151 (Strix Halo / Radeon 8060S) at 40.4 PP / 13.6 TG plain decode and 17.5 TG MTP K=2. 31 follow-up commits drive the chat template, DSML tool-calling, grammar-guided decoder, multi-turn history, prefix cache, and HIP-graph correctness for real pi-coding-agent sessions.

- [x] `kernels/` (HIP source) — 72 new V4F-specific kernels + small touch-ups on a handful of existing ones (compressor / gemm WMMA variants / fused rmsnorm)
- [x] `crates/rdna-compute` (kernel dispatch / RDNA arch routing) — +73 `Gpu` dispatch methods, +72 kernel-source const declarations, arch-aware `gemm_q8_0_wmma` (gfx12 / RDNA3+ split)
- [x] `crates/hipfire-runtime` (LM runtime) — daemon V4F arm (arch_id=9 → `generate_deepseek4`), DSML stream parser + grammar matcher integration, prefix-cache (LCP + per-turn token cache + decoded-vocab cache), V4F state/weights/PBS free path in `unload_model`, `hfq_split` example tool, `shrink_pread_buf` for UMA OOM hygiene
- [x] `crates/hipfire-quantize` — `deepseek_v4` model_type → arch_id 9; new `--format deepseek4-q8-mtp` family; `--include-prefix mtp.` for sidecar quants; new `deepseek4_dequant_check` + `imatrix_inspect` example tools
- [x] examples / daemon — `examples/deepseek4_chat`, daemon's `generate_deepseek4` arm
- [x] docs / CI / scripts — `docs/MODELS.md` (catalog entry), `docs/architecture-ids.md` (arch_id=9), `docs/deepseek4-pr-body.md` (long-form PR write-up), `CREDITS.md` (antirez/ds4 reference), `scripts/coherence-gate-deepseek4-mtp.sh` (new sibling gate), `cli/registry.json` (model tag), `cli/index.ts` (MTP sidecar auto-pull, V4F arch-conditional prompt shaping, OpenAI `cached_tokens` + `finish_reason` + `reasoning_content` plumbing), `.githooks/pre-commit` (HOTSPOT regex covers DeepSeek V4)
- [x] **NEW crate** `crates/hipfire-arch-deepseek4/` — Architecture trait impl, forward pass, MTP head, DSML parser + grammar matcher, `deepseek4_chat` example

## Test plan

- [x] `cargo build --release --workspace --features deltanet` clean (warnings only in pre-existing `rdna-compute` / `hipfire-runtime` libs)
- [x] `cargo test -p hipfire-arch-deepseek4 --lib` — 46 tests pass (DSML parser round-trip, bare-close recovery, missing-string-attr recovery, think-mode parser priming, grammar matcher state transitions, schema-required enforcement)
- [x] `./scripts/coherence-gate.sh` clean — V4F arm: capital-of-France, sheep reasoning, LRU-cache code review all OK
- [x] `./scripts/coherence-gate-deepseek4-mtp.sh --fast` clean — V4F MTP K=2 spec-decode coherent (80.5% accept, unique_ratio=0.55)
- [x] Pre-commit hook runs both gates on every V4F-touching commit
- [ ] `./scripts/speed-gate.sh` — not applicable; current `bench_qwen35_mq4` is qwen35-only. V4F numbers documented in `tests/speed-baselines/gfx1151.txt` but not enforced. V4F-aware speed-gate is a documented follow-up.

## Architecture-trait change?

**No** — this PR does not modify the `Architecture` trait surface in `crates/hipfire-runtime/src/arch.rs`. It adds a new `impl Architecture for DeepseekV4` in the new `hipfire-arch-deepseek4` crate. The four behavior-override hooks (`LoopGuardOverrides`, `SamplerOverrides`, `PromptFrameOverrides`, `EosFilterOverrides`) are left at their defaults — V4F's daemon arm (`generate_deepseek4`) handles its own chat-template scaffolding and `<｜end▁of▁sentence｜>` EOS detection rather than routing through the standard runtime machinery. Adding the overrides for convention symmetry is a documented nice-to-have follow-up, not a functional gap.


nwoolmer (commented):
Self-reviewing, lots of cleanup to be done

Kaden-Schutt:
Pushed a fix-forward commit for the MTP/tool-call path:

`26e8744b fix(deepseek4): constrain mtp tool-call formatting`

What changed:
- Threads the existing DSML grammar matcher into DeepSeek4 MTP speculative decode when tools are present.
- Applies the same structural mask to both MTP draft logits and verifier logits before accepting/emitting tokens.
- Runs the spec-mode stream through the DSML parser so complete tool-call blocks emit structured `tool_calls` and set `finish_reason: "tool_calls"`.
- Leaves the normal non-tool MTP path dormant with respect to grammar; final-binary fast code K2 matched prior perf/acceptance.

Why:
- Plain structured-tool decode parsed correctly.
- MTP previously emitted a near-valid DSML call but malformed closing tags (`</｜DSML｜inv>`, `</｜DSML｜tool_call>`), so no structured `tool_calls` event was produced and the turn ended as `finish_reason: "stop"`.

Validation on hipx / gfx1151, final daemon md5 `d453ba0476f18ab8e1ae536616b4188b`:

| Probe | Result |
|---|---:|
| Plain structured tool decode | `13.17 tok/s`, parsed `read(path=/tmp/fibonacci.c)`, `finish_reason: "tool_calls"` |
| Forced-format MTP K2 | `13.85 tok/s`, parsed same tool call, `87.5%` accept, `finish_reason: "tool_calls"` |
| Forced-format MTP K3 | `15.77 tok/s`, parsed same tool call, `78.43%` accept, `finish_reason: "tool_calls"` |
| Old malformed MTP K2 baseline | `17.31 tok/s`, but no parsed tool call and `finish_reason: "stop"` |

So the valid tool-call path is not a regression versus valid AR: K2 is about +5%, K3 about +20%. The only slower comparison is against the old malformed raw MTP stream, which was not a valid tool-call result.

Non-tool MTP regression check:
- Full matrix after the fix: no hard errors, report `/tmp/coherence-deepseek4-mtp-20260527-052522.md`.
- Final-binary fast code K2: `19.17 tok/s` vs prior `19.18 tok/s`, same `81.7%` accept, report `/tmp/coherence-deepseek4-mtp-20260527-054502.md`.

Commands run:
- `cargo check -p hipfire-arch-deepseek4 -p hipfire-runtime --example daemon`
- `cargo test -p hipfire-arch-deepseek4 grammar::tests --lib`
- `cargo build --release --example daemon -p hipfire-runtime`
- `./scripts/coherence-gate-deepseek4-mtp.sh --full`
- `./scripts/coherence-gate-deepseek4-mtp.sh --fast`
- Structured tool daemon probes for AR, MTP K2, and MTP K3.

PR #347, opened 2026-05-27
fix(moe): default gfx1151 i8 MMQ MoE OFF (35B-A3B CJK regression)

nwoolmer:
## Summary

The i8 grouped-MMQ MoE path enabled by default on **gfx1151** (Strix Halo) in 6cf455dc produces incoherent output on `qwen3.6:35b-a3b`: the model emits CJK / structural repetition on simple English coding prompts. The standalone NRMSE channel test passes at ~0.4% vs the FP16 reference, so the regression slipped through tile-level validation but breaks the production MoE chain.

This PR flips the default to OFF on gfx1151 only. Power users can still opt in with `HIPFIRE_MOE_GROUPED_I8=1`. The underlying kernel correctness fix is left as separate work — diagnostic infrastructure for it is already in progress on master.

- **One-line change**: `unwrap_or(true)` → `unwrap_or(false)` in `gemm_hfq4g256_moe_grouped_wmma_k2`, gfx1151 branch only.
- **Untouched**: gfx11 dGPU (gfx1100/1101/1102/1103) — different kernel binary, no garbage reports against those archs. gfx12 — already opt-in.

## Reproducer

`scripts/repro-qwen3.6-a3b-garbage.sh` (already in tree). Confirmed on Strix Halo (gfx1151, AMD Ryzen AI MAX+ 395 / Radeon 8060S), `qwen3.6-35b-a3b.mq4`, temp=0.0, max_tokens=96, prompt = `benchmarks/prompts/dirty/dirty_fib.txt`:

**Default (pre-fix, i8 ON):**
```
秵,  2,  2, 2, 2, 2, 2, 2, 2

1. **分析**：
   - **观察**：
     - **观察**：
       - **观察**：
[repro] has_python_signal=0  has_cjk=1  ascii_alpha_ratio=0.000
[repro] VERDICT: garbage (regression present)
```

**With `HIPFIRE_MOE_GROUPED_I8=0` (equivalent to this fix):**
```
<think>
The user wants a Python function to compute the n-th Fibonacci number.
The requirements are:
1. Use memoization to keep recursion fast.
2. Handle base cases n == 0 and n == 1.
...
[repro] has_python_signal=1  has_cjk=0  ascii_alpha_ratio=0.924
[repro] VERDICT: coherent (model works)
```

## Test plan

- [x] Reproduce CJK garbage on master with default env
- [x] Confirm coherent output with `HIPFIRE_MOE_GROUPED_I8=0 HIPFIRE_MOE_GROUPED_I8_K4=0`
- [x] One-line dispatch flip is byte-equivalent to the env=0 case verified above
- [ ] CI green
- [ ] Coherence gate green on at least one gfx1151 model after rebuild

## Why not also disable on gfx11 dGPU?

gfx11 dGPU uses a different kernel binary (`gemm_hfq4g256_moe_grouped_mmq.gfx11_dgpu.hip`) and there are no reports of the same regression against it. The gfx1151 binary is its own per-arch file. Scope is intentionally narrow to the arch with the reported regression.

## Follow-up (not in this PR)

Root-cause the i8 path so the channel test catches the production failure mode — the divergence appears to live in the X-input chain (FP16 cache invalidation around `gated_norm_f32_batched`, per in-progress diagnostic work on master) rather than the WMMA tile itself.

### DIFF
--- a/crates/rdna-compute/src/dispatch.rs
+++ b/crates/rdna-compute/src/dispatch.rs
@@ -15177,10 +15177,14 @@ self.flags.rocblas_min_batch.unwrap_or(4)
     ) -> HipResult<()> {
         self.bind_thread()?;
         // gfx1151 (Strix Halo iGPU) i8 MMQ port: lift the compute ceiling
-        // from ~71 (FP16 WMMA) to ~140 TFLOPS (i8 WMMA). Opt-out via
-        // HIPFIRE_MOE_GROUPED_I8=0; default ON for gfx1151 only.
+        // from ~71 (FP16 WMMA) to ~140 TFLOPS (i8 WMMA). Opt-in via
+        // HIPFIRE_MOE_GROUPED_I8=1; default OFF for gfx1151 pending fix
+        // for the qwen3.6:35b-a3b CJK-output regression seen against
+        // commit 6cf455dc (kernel produces NRMSE-clean tile output in the
+        // channel test but corrupts the production MoE chain on real
+        // 35B-A3B prompts — model emits Chinese on English code prompts).
         let use_i8_gfx1151 = self.arch_caps.is_gfx1151()
-            && self.flags.moe_grouped_i8.unwrap_or(true);
+            && self.flags.moe_grouped_i8.unwrap_or(false);
         if use_i8_gfx1151 {
             // Optional deeper-pipeline variants (opt-IN, default OFF).
             // Same kernarg layout + scatter contract as the k2 default.


Kaden-Schutt:
Thanks @nwoolmer for the detailed repro and the X-input root-cause note. Before merging I tried to reproduce the CJK regression on **current master (`09e94b25`)** on a real Strix Halo **gfx1151** (Radeon 8060S, Ryzen AI MAX+ 395), and I **cannot get it to reproduce under any config I've tried**. Sharing the matrix and asking for a couple details so we don't default-off a working, ~6× faster path unnecessarily.

**Setup:** fresh `origin/master` (`09e94b2`) worktree, freshly built `daemon`, gfx1151 selected via `HIP_VISIBLE_DEVICES` (the 8060S iGPU; dev 0 here is a 5700 XT), model `qwen3.6-35b-a3b.mq4`, prompt `benchmarks/prompts/dirty/dirty_fib.txt`, temp 0.0. The i8 `.hsaco` was JIT-compiled fresh (verified mtimes).

**All coherent, 0 CJK:**

| config | result |
|---|---|
| i8 ON (k2 default) | coherent — prefill **749 tok/s** |
| i8 ON + k4 (`HIPFIRE_MOE_GROUPED_I8_K4=1`) | coherent |
| i8 ON + k8 (`HIPFIRE_MOE_GROUPED_I8_K8=1`) | coherent |
| i8 OFF (`HIPFIRE_MOE_GROUPED_I8=0`) | coherent — prefill 128 tok/s |
| asym3 KV / q8 KV (`HIPFIRE_KV_MODE=q8`) | coherent both |
| 5× sequential gens, same daemon session | all coherent (no cross-call X-cache degradation) |
| 2nd model `qwen3.5-35b-a3b.master.mq4` | coherent |
| forced cold recompile vs cached `.hsaco` | **byte-identical** (`md5 579e03db`), coherent — no stale-cache / hash-gate divergence |
| `coherence-gate.sh --full` (incl. a 583-token A3B MoE gen) | all OK, 0 CJK across the whole report |

So on this box the i8 grouped path is byte-deterministic and coherent, and disabling it costs ~5.9× prefill (749→128 tok/s).

Since your channel test passes but you saw real garbage, the discrepancy is most likely one of:
1. **master moved past the fix** — your note references in-progress X-input / `gated_norm_f32_batched` cache work "on master"; if that landed before `09e94b2`, the production failure may already be resolved.
2. **stale kernel cache on your box** — a `.hsaco` whose `.hash` didn't invalidate after a kernel-source change (ruled out here via forced recompile, but worth a check on your side).
3. **a config I didn't hit.**

Could you share:
- the exact commit you reproduced on (`git rev-parse HEAD`),
- whether clearing `.hipfire_kernels/gfx1151/` and recompiling changes the output, and
- the `scripts/repro-qwen3.6-a3b-garbage.sh` script — it's referenced here but isn't in `master`, `46e93d15`, or this PR branch as far as I can find?

If it reproduces on a specific commit I'm happy to bisect from there. Mainly want to avoid landing a perf regression for a bug that may already be fixed. 🙏

nwoolmer:
Mystery solved, Kaden's model diverged from HF state. New model is up, works fine. Closing as unnecessary.

PR #356, opened 2026-05-29
perf(deepseek4): MoE grouped WMMA optimizations — MMQLOAD default, prefetch, 8w variant

fivetide:
# MoE Grouped WMMA Optimizations for DeepSeek V4 Flash (MQ2 Lloyd)

## Commits on This Branch

| # | Commit | Area | Impact |
|---|--------|------|--------|
| 1 | `2c8bb56b` feat: wire mtp_mode/mtp_k config | MTP spec-decode | Config UX |
| 2 | `1f8619e1` multi-row wo_a GEMV + default-4w MoE for gfx11 | wo_a + MoE GEMM | -17% wo kernel; 4w default for gfx11 |
| 3 | `56bf0912` barrier-free nosync MoE grouped WMMA kernel | MoE GEMM | +75% MoE kernel BW, +12.9% wall |

Commits 1-2 by prior author, commit 3 adds the barrier-free (nosync) kernel as opt-in
behind `HIPFIRE_DEEPSEEK4_MOE_NOSYNC=1`.

## What Changed

**New kernel:** `gemm_mq2g256_lloyd_moe_grouped_wmma_4w_k2_nosync` — barrier-free
variant of the 4-warp grouped MoE MQ2-Lloyd WMMA GEMM. Eliminates `__syncthreads()`
and LDS X staging — each wave loads X directly from global memory, removing wave-level
serialization at the cost of 4× redundant global reads (free at ~4.8% BW utilization).

**Bugfix:** The MMQ-style index-pack preload (`_mmqload`) variant was dropped after
[nwoolmer's review](https://github.com/Kaden-Schutt/hipfire/pull/358#issuecomment-4580633064)
identified a bug (loading 8/16 packs, half-work). With the fix the kernel was perf-neutral,
so it was removed rather than carried as dead code.

## Benchmark: DeepSeek V4 Prefill BW (gfx1151, 40 CUs, ROCm 7.2.3)

Baseline: 4w default, nosync opt-out. Nosync: `HIPFIRE_DEEPSEEK4_MOE_NOSYNC=1`.

```
ctx=256
  MoE kernel:   971.5 ms → 553.5 ms  (+75%)
  Wall time:   3092.5 ms → 2692.4 ms (-12.9%)
  MoE BW:        10.8 GiB/s → 18.9 GiB/s
```

Measured with `profile_prefill_deepseek4 --prefill 256 --warmup 1 --pp-batch 256`.
[...]

PR #358, opened 2026-05-29
Reconcile #349 + #355 + #356 — DeepSeek4 MoE/prefill perfmaxx (needs gfx1151 validation)

Kaden-Schutt:
## What this is

A **reconciliation** of three open DeepSeek4 / MoE-prefill PRs into one branch on top of current `master`:

| PR | Author | Title |
|----|--------|-------|
| #349 | @fivetide | wire `mtp_mode`/`mtp_k` into global + per-model config |
| #356 | @fivetide | MoE grouped-WMMA opts (MMQLOAD default, prefetch, multi-row wo_a, barrier-free) |
| #355 | @Kaden-Schutt | batch verify lm-head (+2.07%) + gfx1151 prefill MoE investigation |

**Why reconcile:** #355 and #356 both grew from the same gfx1151 DeepSeek4 MoE/prefill perfmaxx line and **diverged** — both edit `dispatch.rs` / `kernels.rs` / `forward.rs` and both add overlapping kernel-variant families. Merging them independently would conflict and leave a confusing pile of half-wired variants. #349's config commit (`2c8bb56b`) is also embedded inside #356. This branch lands all three coherently so reviewers compare against one tree, not three moving targets.

> ⚠️ **Not hardware-validated.** Kaden's gfx1151 box (`hipx`) is offline and this machine is gfx1010. The reconciliation is **merge-correct and compiles clean** (`cargo check -p rdna-compute -p hipfire-arch-deepseek4 --features deltanet` → 0 errors), but the perf/coherence claims below were made by the original PR authors and have **not** been re-run post-merge. This PR needs a Strix Halo (gfx1151) contributor to validate before it lands on `master`.

## How conflicts were resolved

Both PRs are **purely additive** — every new dispatch fn and kernel const has a distinct name. Git only conflicted because it aligned the identical function-body templates. Resolution = **union of both sides**, verified: 0 conflict markers, each new function present exactly once, `cargo check` clean.

**`crates/rdna-compute/src/dispatch.rs`** — keep all new MoE entry points:
- from #356: `..._4w_k2_mmqload`, `..._4w_k2_mmqload_nosync`, `wo_per_group_batched_q8_0_multirow`
- from #355: `..._4w_k2_n32`, `..._4w_k2_cnd`, `..._8w_k2`, `wo_per_group_batched_q8_0_wmma_4w`

**`crates/rdna-compute/src/kernels.rs`** — union of all 4 new `*_SRC` kernel-source consts.

**`crates/hipfire-arch-deepseek4/src/forward.rs`** — the `ffn_batched` MoE dispatch (both gate_up and down) becomes a single `if/else if` chain:

```
n32 → cnd → 8w        (#355 research levers, all opt-in, default OFF)
  → nosync            (#356, opt-in)
  → mmqload           (#356, DEFAULT ON for 4w)
  → 4w_k2             (plain 4-warp)
  → base k2           (fallback)
```

#355's lm-head batch (+2.07%, default ON, token-identical by construction) and top-k-attn-direct changes merged **cleanly** (no conflict).

### 🔴 Reviewer decision flagged inline

The 4w-default **arch gate** differs between the two PRs:
- #356: `gpu.arch.starts_with("gfx11") || gpu.arch.starts_with("gfx12")` (broad)
- #355: `gpu.arch == "gfx1151"` (Strix-Halo only)

**This branch keeps #356's broader `gfx11 || gfx12` gate** (commented `[REVIEWER NOTE]` at the gate-up dispatch site in `forward.rs`). If the 4w/mmqload path hasn't actually been validated on a gfx11 **dGPU** (7900 XTX et al.), this should be narrowed back to `gfx1151` before merge. **@fivetide @nwoolmer — please confirm.**

## Validation needed on gfx1151 (Strix Halo)

Per `CLAUDE.md`: byte-identical prompts, record md5, fresh process per measure, median of 3–5, one `--max 16` warmup per cell.

**1. Coherence (mandatory, hard gate):**
```
./scripts/coherence-gate.sh
./scripts/coherence-gate-dflash.sh        # spec-decode paths touched
./scripts/coherence-gate-deepseek4-mtp.sh # if a DS4 -mtp model is present (covers #349)
```

**2. lm-head batch (#355) — confirm +2.07% AND token-identical:**
```
HIPFIRE_DEEPSEEK4_BATCH_HEAD=1  vs  =0
# K=2 MTP code prompt; expect ~19.5 vs ~19.1 tok/s, committed token ids identical
```

**3. MoE MMQLOAD default (#356) — confirm +20–35% prefill BW:**
```
HIPFIRE_DEEPSEEK4_MOE_MMQLOAD=0  vs  default (on)
# profile_prefill_deepseek4, ctx ∈ {128,256,512,1024}
```

**4. Spot-check the #355 opt-in levers still dispatch (all default OFF):**
```
HIPFIRE_DEEPSEEK4_MOE_N32=1 / _CND=1 / _8W=1     # should select n32/cnd/8w kernels, coherent
HIPFIRE_DEEPSEEK4_MOE_NOSYNC=1                    # barrier-free path
```

**5. Confirm the broader 4w arch gate on a gfx11 dGPU** (the reviewer-decision item above).

## Disposition

Do **not** merge until a contributor validates on real gfx1151 hardware. Once green, this single branch can replace #349 + #355 + #356.


nwoolmer:
## gfx1151 hardware validation — PR #358 (reconcile #349 + #355 + #356)

**Verdict: 🔴 DO NOT MERGE as-is.** The default DeepSeek4 MoE path ships a
**long-context coherence regression** on gfx1151. Build is clean and the perf
levers work, but the `MMQLOAD` index-pack preload (#356, default-on for gfx11+)
produces a hard token attractor on long generations. Short generations are
clean, which is why it slipped through. **Root cause (confirmed):** the mmqload
kernel preloads only **8 of the 16** weight packs per K-group and reuses each twice,
so it decodes **half the weights** (details + 2-line fix below). **Stopgap:**
default `HIPFIRE_DEEPSEEK4_MOE_MMQLOAD=0` (keeps the clean, fast 4w path).

### Environment
- **GPU:** gfx1151 (Strix Halo / Radeon 8060S APU), 125 GB UMA
- **Branch:** `reconcile/ds4-moe-perfmaxx-349-355-356` @ `ede90201`
- **Build:** `cargo build --release --features deltanet` → **0 errors** (PR only ran `cargo check`)
- **Model:** `deepseek-v4-flash.mq2lloyd` (81 GB) + `-mtp` addon
- **Method:** byte-identical prompts (md5 recorded), temp 0, fresh process per measure, warm DPM/page-cache.

### Step 1 — Coherence (mandatory) — ⚠️ BLOCKING FAIL on long context
**Broad gate** (`coherence-gate.sh --full`): **PASS**, no hard errors — all present
qwen models + `qwen3.6-35b-a3b` MoE + DS4 `cap`/`reason` clean and correct.

**DS4-MTP gate** (`--full`, K=2 and K=3): **PASS** — `code/prose/cap/math` at K=2
and `code-k3`/`prose-k3` all clean (uniq ≥ 0.54, accept 56–88%). *(One run showed
a `code-k3` HARD-FAIL with `no_committed_ids`; that was a daemon I accidentally
SIGTERM'd mid-case, not a model failure — clean on re-run.)*

**Long-context regression (the finding).** Prompt
`benchmarks/prompts/coherence_lloyd_long.txt` (md5 `f20bbc4f...`, a normal
180-token LRU-cache question), 220 generated tokens, spec K=3. Isolation matrix
(byte-identical prompt/seed, only env changed):

| cfg | config | MoE path | result |
|-----|--------|----------|--------|
| A | default | 4w + **mmqload** | **ATTRACTOR** — `"…also also also also…"`, uniq 0.117 |
| B | spec-decode OFF (plain greedy) | 4w + **mmqload** | **ATTRACTOR** (byte-identical to A) |
| C | `MMQLOAD=0 LLOYD_4W=0` (= master MoE) | base k2 | **CLEAN** — correct LRU answer |
| D | `BATCH_HEAD=0` | 4w + **mmqload** | **ATTRACTOR** |
| G | `LLOYD_4W=1 MMQLOAD=0` | 4w, no preload | **CLEAN** |
| H | `MOE_N32=1` | 4w-family, no preload | **CLEAN** |

**`MMQLOAD` (index-pack preload) is the sole discriminator** between attractor and
clean output. The regressing kernel is
`gemm_mq2g256_lloyd_moe_grouped_wmma_4w_k2_mmqload` (made default by commit
`4e9787e9`). **Exonerated:** spec-decode/MTP (B attracts identically with spec
off), the new SWA top-k-direct attention kernel (on in both A and C), the lm-head
batch (D), and the 4w tiling itself (G/H clean).

**Regression vs master:** `LLOYD_4W`/`MMQLOAD` default-on come from commits unique
to this branch (`1847caed`, `4e9787e9` in `master..HEAD`); on master the DS4 MoE
path *is* config C, which is clean. So this is a genuine PR-introduced regression,
not a pre-existing structural attractor.

### Step 2 — lm-head batch (#355): ✅ token-identical (perf claim not reproduced)
`BATCH_HEAD=1` vs `=0`, K=2 MTP, code prompt (md5 `9604bed9...`). Committed
token-ids are **identical** — and not just for the first generate: across 5
back-to-back generates (accumulating context, so 5 distinct context lengths) the
two sides matched **pairwise at every step** (`bh1_tN == bh0_tN` for N=1..5).
That also proves the path is **deterministic** (so the long-context attractor is a
deterministic divergence, not RNG). #355's "token-identical by construction" holds.

Perf: the **+2.07% did not reproduce — it's a small *regression* on this APU.** A
clean interleaved fresh-process A/B (3 trials, single generate, `MMQLOAD=0` to
isolate #355) gives median decode tok/s `BATCH_HEAD=1` **9.42** vs `=0` **10.06**
= **−6.35%**, consistent across all 3 trials (every run token-identical, hash
`d477a4d2`). So the lm-head batch is *correct* but a ~6% decode loss here at K=2,
not a win. (Absolute ~9–10 tok/s is ~half the PR's cited ~19.5 — different hardware;
only the relative delta compares. The batched head likely helps verify batches
larger than K+1=3.) Not blocking — it's token-identical — but worth revisiting
whether it should be default-on for gfx1151.

### Step 3 — MoE MMQLOAD prefill BW (#356): perf real, but the kernel is BROKEN
`profile_prefill_deepseek4 --no-profile`, default(on) vs `MMQLOAD=0`, median of 2,
warmup 3, synthetic tokens.

| ctx | on tok/s | off tok/s | Δ% |
|----:|---------:|----------:|---:|
| 128 | 80.4 | 71.4 | +12.6 |
| 256 | 97.3 | 89.4 | +8.8 |
| 512 | 119.0 | 93.1 | **+27.8** |
| 1024 | 100.7 | 97.0 | +3.8 |

The prefill speedup is **real** (+3.8…+27.8%, in the PR's +20–35% ballpark at 512).
**But it ships wrong math.** The example's `PREFILL_CHECK` (final-position argmax +
logit_sum over the full vocab, same synthetic input) shows the mmqload path
**diverges from the `off` reference at every ctx** and emits a **NaN**:

| ctx | on argmax (2 trials) | off argmax | note |
|----:|----------------------|-----------:|------|
| 128 | 162, 162 | 194 | wrong |
| 256 | **0 (logit_sum=NaN, logit_max=-inf)**, 11 | 11 | NaN, nondeterministic |
| 512 | 81, 81 | 201 | wrong |
| 1024 | 75, 75 | 19123 | wrong |

`off` is stable/self-consistent across trials; `mmqload` is wrong-and-sometimes-NaN.
**Root cause confirmed below** (the kernel decodes only half the weight packs). Note
part of the "+20–35%" is just doing **half the A-matrix work** — a *correct* mmqload
will be slower than these numbers.

### Step 4 — opt-in levers: all dispatch; one variant inherits the bug
Dispatch (profiled kernel table, ctx 256, 0 crashes) — every lever selects its
distinct kernel, and all are default-OFF (verified in `forward.rs`):

| lever | kernel dispatched | long-prompt coherence |
|-------|-------------------|------------------------|
| `default` | `..._4w_k2_mmqload` | ATTRACTOR (step 1) |
| `N32` | `..._4w_k2_n32` | CLEAN (step 1d) |
| `CND` | `..._4w_k2_cnd` | CLEAN |
| `8W` | `..._8w_k2` | CLEAN |
| `NOSYNC` | `..._4w_k2_mmqload_nosync` | **ATTRACTOR** |

`nosync` (a mmqload-derived, barrier-free variant) attracts exactly like the
default — confirming the fault is in the shared **mmqload index-pack preload**. The
non-preload variants (`n32`/`cnd`/`8w`) are all coherence-clean.

### Step 5 — broader gfx11||gfx12 gate: N/A
This machine is a gfx1151 **APU**, not a gfx11 dGPU. But the reviewer-flagged gate
question is moot: the default path fails coherence on gfx1151 itself, so the gate
should not ship broadened (or narrowed) until the `mmqload` kernel is fixed.

### What's good (mergeable) vs what blocks
**Good — keep:** #349 mtp config; #355 lm-head batch (token-identical, step 2); the
4w grouped tiling (`LLOYD_4W`) default — coherence-clean (G/H/cnd/8w) and a real
prefill win; the `n32`/`cnd`/`8w` opt-in levers (clean); the SWA top-k-direct
kernel (exonerated). Build is clean.

**Blocks — `gemm_mq2g256_lloyd_moe_grouped_wmma_4w_k2_mmqload.hip` (and its
`..._mmqload_nosync` sibling).**

### Root cause (exact) — the preload reads only half the weight packs
Each K-group is 72 B: 8 B codebook + **64 B = 16 uint packs** (256 MQ2 weights ×
2 bits), one pack per 16-wide k-tile. The clean baseline `_4w_k2.hip` loads a
distinct pack per k-tile (all 16):
```c
// _4w_k2.hip:151  (k_off_a=kt*16, k_off_b=(kt+1)*16)
const unsigned int pa = *(const unsigned int*)(gp + 8 + k_off_a / 4);  // pack kt
const unsigned int pb = *(const unsigned int*)(gp + 8 + k_off_b / 4);  // pack kt+1
```
The `_mmqload.hip` "preload" hoist gets the count and indexing wrong:
```c
// _4w_k2_mmqload.hip:108   ← preloads only 8 of the 16 packs (gp+8..gp+39)
unsigned int packs[8];
for (int i = 0; i < 8; i++) packs[i] = *(const unsigned int*)(gp + 8 + i * 4);
// _4w_k2_mmqload.hip:130   ← packs[kt/2]==packs[(kt+1)/2] for even kt → each
const unsigned int pa = packs[kt / 2];        //   pack reused for TWO k-tiles,
const unsigned int pb = packs[(kt + 1) / 2];  //   packs 8..15 never read
```
So it decodes weights for k-tiles `{0,0,1,1,2,2,…,7,7}` instead of `{0,1,…,15}` —
the upper 32 B of every group is ignored and the lower 8 packs are each duplicated.
Half-wrong weights ⇒ approximately-right short-term, drift → attractor long-term;
the NaN is the downstream blow-up. (The kernel's own comment — *"The 8 packs occupy
gp+8..gp+71 (64 B)"* — is the tell: 8 packs are 32 B; the region is 16 packs.)

### Fix (2 lines) + recommendation
```c
unsigned int packs[16];
for (int i = 0; i < 16; i++) packs[i] = *(const unsigned int*)(gp + 8 + i * 4);
...
const unsigned int pa = packs[kt];       // was packs[kt/2]
const unsigned int pb = packs[kt + 1];   // was packs[(kt+1)/2]
```
Apply the same to `_4w_k2_mmqload_nosync.hip`. Then:
1. Re-run the long-context coherence repro (below) — must be CLEAN.
2. Re-measure prefill — the real speedup will be **smaller** than +20–35% (it was
   partly from skipping half the weight loads); compare honestly vs the `off`/4w
   path to decide if the preload is still worth keeping default-on.
3. Add a **long-context** case to the DS4 coherence gate (the current short
   cap/reason cases pass even with this bug — that's the gap that let it through).
4. **Stopgap if not fixing now:** default `HIPFIRE_DEEPSEEK4_MOE_MMQLOAD=0` — keeps
   the clean `LLOYD_4W` 4w path (step 3 `off` = 71–97 tok/s).
5. The `gfx11||gfx12` vs `gfx1151` gate question is moot until the kernel is correct.

### Repro
All on `deepseek-v4-flash.mq2lloyd`, prompt `coherence_lloyd_long.txt`
(md5 `f20bbc4f...`), temp 0, K=3 MTP, max 220:
- attractor: default (or `HIPFIRE_DEEPSEEK4_MOE_NOSYNC=1`)
- clean: `HIPFIRE_DEEPSEEK4_MOE_MMQLOAD=0` (or `_N32=1`/`_CND=1`/`_8W=1`)
- synthetic NaN: `profile_prefill_deepseek4 <model> --prefill 256 --warmup 3 --no-profile`
  with vs without `HIPFIRE_DEEPSEEK4_MOE_MMQLOAD=0` (compare the `PREFILL_CHECK` line)

nwoolmer:
## ✅ Root-caused + fixed + verified on gfx1151

Follow-up to the validation above. The long-context attractor is a **2-line-per-file
bug in the mmqload index-pack preload**, now fixed and verified.

### The bug
Each K-group packs 256 MQ2 weights × 2 bits = **64 B = 16 uint packs** (`gp+8..gp+71`),
one pack per 16-wide K-tile. The `_mmqload` "preload" hoist loaded only **8** packs and
indexed `packs[kt/2]`/`packs[(kt+1)/2]` (equal for even `kt`) — so it decoded only the
first 32 B of weights, applied each pack to **two** K-tiles, and never read `gp+40..71`.
Half the weights were wrong. (The baseline `_4w_k2.hip` loads a distinct pack per K-tile,
all 16 — that's why it and `n32`/`cnd`/`8w` stayed clean.)

### The fix
```diff
--- a/kernels/src/gemm_mq2g256_lloyd_moe_grouped_wmma_4w_k2_mmqload.hip
+++ b/kernels/src/gemm_mq2g256_lloyd_moe_grouped_wmma_4w_k2_mmqload.hip
-        unsigned int packs[8];
+        unsigned int packs[16];
         #pragma unroll
-        for (int i = 0; i < 8; i++) {
+        for (int i = 0; i < 16; i++) {
             packs[i] = *(const unsigned int*)(gp + 8 + i * 4);
         }
@@ inner K-tile loop @@
-            const unsigned int pa = packs[kt / 2];
-            const unsigned int pb = packs[(kt + 1) / 2];
+            const unsigned int pa = packs[kt];
+            const unsigned int pb = packs[kt + 1];
```
Same change in `..._4w_k2_mmqload_nosync.hip` (it's a mmqload-derived variant and
inherited the bug — it attracted identically in testing).

### Verification (gfx1151, `deepseek-v4-flash.mq2lloyd`, rebuilt + JIT-recompiled)
| axis | before | after |
|------|--------|-------|
| synthetic argmax (mmqload vs `MMQLOAD=0`), ctx 128/256/512/1024 | **differs** (162/NaN/81/75 vs 194/11/201/19123) | **MATCH** (194/11/201/19123), **0 NaN** |
| long-context coherence (`coherence_lloyd_long`, K=3, default ON) | ATTRACTOR (uniq 0.117, `"…also also…"`) | **CLEAN** (uniq 0.617, correct LRU answer) |
| prefill tok/s, mmqload ON vs OFF | "+3.8…+27.8%" | **+0.0 / +0.6 / −0.2 / +1.1%** (flat) |

### Implication for the perf claim
The corrected kernel is **perf-neutral** vs the plain 4w path — the original "+20–35%"
was almost entirely the half-work bug (loading 8 of 16 packs). So with the fix, the
`mmqload` preload buys nothing here; reasonable options:
- land the correctness fix and **default `HIPFIRE_DEEPSEEK4_MOE_MMQLOAD` off** (the
  clean 4w path is equivalent + simpler), or
- keep it default-on (now correct, harmless) but don't count it as a speedup.

Either way the **correctness fix should land** so the kernel isn't a latent landmine,
and the DS4 coherence gate should grow a **long-context** case (the short cap/reason
cases passed even with this bug).

PR #367, opened 2026-05-31
Qwen3.5/3.6: grammar-guided tool calling + reliable prompt caching for agentic workloads

nwoolmer:
## Summary

Makes Qwen3.5/3.6 (dense + MoE) dependable as the backend for agentic coding clients (e.g. Pi and other OpenAI-compatible tool-calling agents). Two classes of failure are addressed:

1. **Malformed / dropped tool calls** — the model would emit slightly-off `<tool_call>` JSON, trip the repetition guard on legitimate code, or emit a tool call truncated by the response budget that the client couldn't parse.
2. **Expensive or terminated prefills** — multi-turn and client-edited histories either missed the prompt cache entirely (full cold re-prefill every turn) or ran a long silent prefill/`<think>` that blew the client's socket/wall deadline and terminated the stream.

Everything that changes output is opt-out-able via an env knob and defaults to behaviour-preserving.

### Grammar-guided tool calling
- Sample-time logit mask over a tool-call state machine (`grammar.rs`) forces valid `<tool_call>{...}</tool_call>` structure whenever a request carries `tools`. `is_free()` short-circuits the mask in free-text regions, so prose/answers stay unconstrained.
- The n-gram repetition guard no longer trips on legitimate code emission (3-byte × 6-repeat + uniform-byte filter — struct fields / double-newline blocks were false-positiving).
- Truncated tool calls are never forwarded: an unclosed `<tool_call>` at `finish_reason=stop` is dropped rather than emitted as broken JSON.
- **Chunked-write nudge** (`HIPFIRE_CHUNK_WRITE_NUDGE`, default on): when a file-writing tool is exposed, a short system directive steers the model to build large files incrementally across bounded tool calls. A single ~30K-token `write` can't finish before the client deadline at long-context decode rates → it truncates mid-output → the whole turn is lost.

### Reliable prompt caching
- **Per-turn reuse for agentic/DFlash chat**: reuse the cached prefix on pure forward-extension (`lcp == prior_len`). The recurrent DeltaNet state lives at the *end* of the prior conversation, so only an exact forward extension is safe to reuse — which is the common agentic shape (append a user/tool turn).
- **Checkpoint-resume on divergent renders** (`lcp < prior_len` — client dropped or edited earlier history): restore a `DeltaNetSnapshot` from the latest checkpoint ≤ `lcp`, rewind `seq_pos`, and re-prefill only the tail. FullAttention KV is positional, so `[0..pos]` stays resident and needs no snapshot. Implemented for **both** the AR and DFlash spec-decode paths; default on, gated to single-GPU + no-eviction; **byte-identical to a cold prefill** (verified).
- **Reasoning budget + force-answer watchdog**: a bounded `<think>` span (`max_think_tokens`, default 2048) plus a watchdog that force-closes thinking, so a long reasoning turn can't silently run past the client deadline and terminate the stream.

### Correctness / accounting fixes
- DFlash decode-abort previously left `dn_state=None`, panicking the *next* request at the cold-reset unwrap; the model slot is now restored before returning.
- Cached-turn `prompt_tokens` no longer double-counts (reports new-token prefill count, not the full rendered length).
- A streaming whitespace heartbeat keeps Bun's 255 s socket idle-timeout from killing long non-stream prefills; a stdin abort protocol stops prefill on client-cancel.

### Tooling
- Prompt-cache + DFlash bench matrix; checkpoint-resume byte-ident + perf A/B harnesses; qwen3.6 AR/DFlash prefill harness; grammar soak + cache-alignment tests; DFlash coherence gate fixed for qwen3.6 model naming.

## Validation

| Check | Scope | Result |
|---|---|---|
| Coherence gate (short) | qwen3.5 0.8b/cap · 4b/code · 9b/reason · 9b/tool-call | ✅ 4/4 OK, no hard errors |
| DFlash coherence gate | qwen3.6-27b dflash prose + code | ✅ no hard errors |
| DeepSeek4 MTP coherence | deepseek4 mtp code k2 | ✅ no hard errors |
| Grammar unit tests | `hipfire-arch-qwen35` | ✅ 73/73 |
| Tool-call parser tests | `cli/parse_tool_calls.test.ts` | ✅ 12/12 |
| Detector tests | `hipfire-detect` | ✅ 54 pass |
| Byte-ident resume == cold | DFlash (think-off) + AR (think-on) | ✅ identical signatures |
| Multi-turn prefix reuse | 15.7K-tok ctx, all 4 cells | ✅ 0 → 50 → 67 → 75 → 80% |
| Divergent checkpoint-resume | DFlash | ✅ 98% reuse, 6.0 s vs 105.8 s cold (**17.5×**) |
| Pi cache replay (4-turn) | reuse + tool calls | ✅ ~100%/turn, tool calls correct, 0 grammar violations |
| Prefill perf A/B (resume on/off) | ~8K ctx, qwen3.6-27b | ✅ 154.5 vs 154.9 tok/s (noise, <0.3%) |
| **Production (Pi session)** | 27K-ctx divergent retry | ✅ resume replayed **2004 tokens vs 27,159 cold** |
| Chunked-write nudge | live, qwen3.6-27b | ✅ tools path intact; model recalls + acts on nudge |

Notes:
- The **production** row is the strongest signal: on a real Pi session that terminated mid-write at 27K ctx, the retry's divergent render resumed from a checkpoint (replaying 2004 tokens instead of cold-prefilling 27,159) — exactly the failure loop this work targets.
- Coherence gates hard-fail only on panics / zero-tokens / timeouts; all soft outputs were reviewed and are fluent, on-topic, and free of attractors.

## KV-cache mode coverage

The caching fixes are **`kv_mode`-independent by construction** and were verified byte-identical across the asym + fwht KV families (not just q8) on qwen3.6-27b:

| KV mode | K layout | DFlash | AR |
|---|---|---|---|
| **fwht2** (prod default, *most aggressive*) | 2-bit FWHT-rotated K + Q8 V | ✅ byte-ident | ✅ byte-ident |
| **asym3** (documented default family) | 3-bit RotorQuant-rotated K + Q8 V | ✅ byte-ident | — |
| **asym2** (most lossy) | 2-bit RotorQuant-rotated K + Q8 V | ✅ byte-ident | — |
| q8 | uncompressed 8-bit | covered¹ | — |

¹ q8 keeps KV uncompressed (no rotation / no asym quant) — the *simplest* layout and a strict subset of the machinery the passing modes exercise, so it's covered a fortiori. `asym4`/`fwht3`/`fwht4` are bracketed too (3-/4-bit, strictly less aggressive than the 2-bit fwht2/asym2 that pass, identical mechanism).

Why it's mode-independent:
- Checkpoint-resume snapshots the **DeltaNet recurrent state** — governed by `state_quant` (a *separate* axis), never the FullAttention KV.
- FullAttention KV is **positional**: resume rewinds the position cursor and re-prefills the tail; `KV[0..ckpt]` stays resident (byte-identical to cold) and is **never re-quantized** — the tail re-writes through the same per-mode quantizer.
- Checkpoints land on **2048-multiples** → block-aligned, so no partial-block hazard for the asym/fwht block quantizers.

`ckpt_resume_enabled()` has no `kv_mode` gate (only the env knob + the single-GPU / no-eviction guard) — correct, given the byte-ident results above.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -224,11 +224,25 @@ back to AR silently.
 ```
 qwen35-9b-dflash-mq4.hfq    590f35403cd7f1d634945233234a12b7  557 MB
 qwen35-27b-dflash-mq4.hfq   7b6df2a4ee1c8d933f0a52e187d1860b  919 MB
-qwen36-27b-dflash-mq4.hfq   ecc64877dfe0a1312b6f4066c3920128  919 MB
+qwen36-27b-dflash-mq4.hfq   204c4c4ceab30cb9ebc118fa9d59a446  919 MB
 qwen3.6-27b.mq4             9a6acdc49bcaa6a7b52ac161444cb769   15 GB
 ```
 
-Any mismatch = re-pull or report.
+Any mismatch = re-pull or report. (The `qwen36-27b-dflash-mq4.hfq`
+checksum was refreshed 2026-05-30 from the stale `ecc64877…` — the HF
+file was re-uploaded since the original manifest; verify against the
+current `204c4c4c…`.)
+
+> **Sizes here are decimal (MB = 10⁶ bytes, GB = 10⁹ bytes), matching
+> Hugging Face's reported sizes and the `hipfire pull` progress bar.**
+> `ls -lh` / `du -h` report **binary** units (MiB = 2²⁰, GiB = 2³⁰) but
+> *label them* `M`/`G`, so a 919 MB file shows as `877M` in `ls`
+> (919 × 10⁶ ÷ 2²⁰ ≈ 877 MiB) and a 15 GB file shows as `14G`. This is
+> not a size mismatch or a truncated download — it's the same byte
+> count in two unit systems. When a download looks "smaller than the
+> manifest," divide by 1.048576 (MB→MiB) or 1.073742 (GB→GiB) before
+> assuming corruption; confirm with the md5, not the human-readable
+> size.
 
 ### Build from source (if you're on a dev branch)
 
--- a/cli/index.ts
+++ b/cli/index.ts
@@ -173,10 +173,22 @@ const CONFIG_DEFAULTS: HipfireConfig = {
   // causes MQ4/MQ6 models to emit gibberish at temp=0 because the penalty
   // applies uniformly even in greedy mode. 1.05 is user-validated.
   repeat_penalty: 1.05,
-  max_tokens: 512,
+  // 4096 is large enough for code-emit tool calls (Pi's `write`/`edit`
+  // tools pass entire file bodies as a string argument inside a single
+  // `<tool_call>` block) without being so large that a runaway thinking
+  // loop burns minutes of decode. Bumped from 512 in 2026-05-28 after
+  // a Pi session truncated a Zig source file mid-string at 512 tokens
+  // and silently dropped the tool call (parseToolCalls returned null
+  // on the unclosed `<tool_call>` block → finish_reason="stop").
+  max_tokens: 4096,
   max_seq: 32768,
   thinking: "on",
-  max_think_tokens: 0,
+  // Default reasoning budget (was 0 = unlimited). A non-zero cap bounds the
+  // <think> span so a long-reasoning turn force-closes and commits to its
+  // answer (daemon splices the continuation) instead of running until the
+  // client times out and terminates the stream mid-think. Override per-model
+  // or set 0 for unlimited (e.g. reasoning.effort=xhigh maps to 0).
+  max_think_tokens: 2048,
   host: DEFAULT_HOST,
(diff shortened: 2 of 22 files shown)


Kaden-Schutt (approved):
Approved after maintainer confirmation that the default changes are intentional. I pushed a narrow cleanup commit (`7be7e47d`) removing the duplicated `currentCacheCapable` assignment.

Validation I ran on the updated head:
- `git diff --check`
- `/home/kaden/.bun/bin/bun test parse_tool_calls.test.ts config_meta.test.ts` — 22 pass / 0 fail

Earlier hiptrx validation on this PR also passed grammar, prompt-frame, release daemon build, and live qwen3.6-27b DFlash cache/tool-call smoke tests.

PR #372, opened 2026-05-31
fix(qwen35): truncation-safe DeltaNet resume + bounded thinking for agentic serve

nwoolmer:
## Problem

Agentic clients driving **qwen3.6-35b-a3b** through `hipfire serve` hit two failure modes that produced garbage / hangs:

1. **Truncation poisons the next turn → garbage.** When a client cancels mid-decode, the AR `generate()` decode-abort path returned *without* resetting state (unlike the DFlash + AR prefill-abort paths). The non-reversible DeltaNet recurrent state + `conversation_tokens` + checkpoints were left dirty with **uncommitted** tokens; the next turn then resumed the prompt cache off that poisoned prior, drifting the recurrent state off-distribution → whitespace garbage / premature EOS, worse on each retry (`prior_tail` decayed real-code → `"   \tt   "` → EOS across three retries).
2. **Unbounded thinking → client timeout.** `check_force_answer` is one-shot, so after the first forced `</think>` close the model re-opens `<think>` and thinks until the client times out; the per-block `max_think_tokens` resets on each re-open so it can't bound the total.

This mirrors **llama.cpp's invariant** for non-reversible recurrent state: never carry uncommitted tokens past a truncation, and resume only from a checkpoint whose prefix is byte-identical (else recompute); never splice un-sampled tokens into the live state.

## Changes (single-GPU AR `generate` path)

**`reset DeltaNet state on AR decode-abort`** — full reset (`seq_pos`, `conversation_tokens`, `prefill_checkpoints`, S/scale/conv buffers, `kv`+`llama_kv` `compact_offset`), mirroring the DFlash + prefill-abort paths. After a cancel the prior is empty → the next turn cold-prefills clean. Post-fix every retained checkpoint sits on a committed prefix (`rpos ≤ lcp` guarantees token identity), so the resume optimization stays correct — documented as an in-code invariant. Also aligns prefill-abort to reset `llama_kv`.

**`keep the grammar matcher in sync through force-answer`** — the force-answer `</think>` injection forwarded tokens but never advanced `grammar_matcher` (the normal sample path does), leaving it stale → malformed tool calls after a forced close. Now advances the matcher over the forced tokens (state-equivalent to llama.cpp's reasoning-budget logit-mask, lower risk than a sampler rewrite).

**`bound unbounded thinking`** — (a) **latch** force-answer for the turn so *every* re-opened `<think>` is re-closed (not just the first); (b) **block** the `<think>` token after latch; (c) a re-arm-proof **total-think cap** `HIPFIRE_MAX_TOTAL_THINK_TOKENS` (default `0`=off): latch at the cap, hard-EOS a margin past it. The latch also makes the 180 s time-based force-answer actually *stick*. Note: the single-token block alone is insufficient (the model re-opens via multi-token *spelling*) — the latch re-close is the real catch.

## Verification (qwen3.6-35b-a3b, gfx1151)

- **`scripts/test-qwen35-abort-resume.sh`** — cancels a turn mid-decode, asserts the next turn enters with `conv_tok=0` (state reset) + no poisoned `[qwen-cache resume]` + coherent output. **PASS** (deterministic).
- **`scripts/test-qwen35-think-cap.sh`** — a hard reasoning prompt that previously ran to the client timeout now terminates server-side with a clean `finish_reason`. **PASS** (49 s vs minutes-if-unbounded; a separate run committed to a real answer in 31 s).
- Normal generation is unaffected — these are decode-loop / abort control-flow changes, not forward-pass/kernel changes (the pre-commit coherence-gate `HOTSPOT` correctly excludes `daemon.rs`); tool calls remain well-formed and the cap is default-off.

## Deferred follow-up

The **multi-GPU** path (`forward_*_multi`, ~`daemon.rs:5031`) has structurally identical force-answer / abort blocks that need the same latch / block / cap / reset. Left as a follow-up — untested here (no multi-GPU hardware).

## Config

The total-think cap ships **opt-in** (default off), settable via the env `HIPFIRE_MAX_TOTAL_THINK_TOKENS` **or** the new `max_total_think_tokens` field in `config.json` (added here — type + `validateConfigValue` + per-model override + config-TUI meta + `applyConfigEnv`; `config_meta.test.ts` passes). Suggested ~**2500–3000** to keep turns under typical ~255 s client timeouts while allowing real reasoning. The always-on fixes (abort-reset, matcher-sync, force-answer latch) need no config.

### DIFF
--- a/cli/index.ts
+++ b/cli/index.ts
@@ -33,6 +33,7 @@ export interface HipfireConfig {
   max_seq: number;        // KV cache capacity allocated at model load (shared across turns)
   thinking: string;       // "on" (model reasons in <think>, stripped from display) | "off" (suppress thinking)
   max_think_tokens: number; // per-turn budget for <think>...</think> reasoning (0 = unlimited)
+  max_total_think_tokens: number; // re-arm-proof TOTAL <think> budget across the turn (0 = off). Force-closes + blocks <think> re-open at the cap; hard-EOS past it. Bounds models that re-open <think> and out-think client timeouts.
   host: string;           // default serve bind address
   port: number;           // default serve port
   idle_timeout: number;   // serve: seconds of inactivity before unloading the model (0 = never)
@@ -189,6 +190,7 @@ const CONFIG_DEFAULTS: HipfireConfig = {
   // client times out and terminates the stream mid-think. Override per-model
   // or set 0 for unlimited (e.g. reasoning.effort=xhigh maps to 0).
   max_think_tokens: 2048,
+  max_total_think_tokens: 0,
   host: DEFAULT_HOST,
   port: DEFAULT_PORT,
   idle_timeout: 300,
@@ -245,6 +247,7 @@ function validateConfigValue(key: string, value: any): boolean {
     case "max_seq": return typeof value === "number" && Number.isInteger(value) && value >= 512 && value <= 524288;
     case "thinking": return ["on", "off"].includes(value);
     case "max_think_tokens": return typeof value === "number" && Number.isInteger(value) && value >= 0 && value <= 32768;
+    case "max_total_think_tokens": return typeof value === "number" && Number.isInteger(value) && value >= 0 && value <= 1000000;
     case "host": return typeof value === "string" && value.trim() === value && value.length > 0 && value.length <= 255 && !/\s/.test(value);
     case "port": return typeof value === "number" && Number.isInteger(value) && value >= 1 && value <= 65535;
     case "idle_timeout": return typeof value === "number" && Number.isInteger(value) && value >= 0 && value <= 86400;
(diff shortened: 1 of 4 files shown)


Kaden-Schutt (approved):
Validated and amended the multi-GPU runtime path on hiptrx. Added commit 223783cf to honor pp>1 aborts, reset uncommitted PP recurrent/KV state, and port force-answer/total-think latch behavior into generate_multi. Evidence: release daemon build passed; qwen3.5-0.8b pp=2 abort repro stops at token 3 with 0 tokens after abort and follow-up turn succeeds; qwen3.6-35b-a3b pp=2 original abort repro stops at token 3 with 0 tokens after abort; full pp-gate passed on devices 0,1. GitHub build and unit tests are green; rustfmt/clippy remain advisory branch-wide failures.

PR #388, opened 2026-06-02
fix(deltanet): requant GDN Q8/Q4 state per-token in batched prefill

fivetide:
## Problem

PARO Qwen3.6-35B-A3B echoes the prompt verbatim (then `<|im_end|>`) on the daemon's **batched-prefill** path. Per-token prefill/decode is correct.

Root cause: the batched GDN kernels (`gated_delta_net_q8` / `gated_delta_net_q4`, launched with `n_tokens = N`) dequantize the DeltaNet state once, run **all N token updates in FP32**, and requantize **once** at the end — collapsing N rounding steps into one. The per-token decode path (`n_tokens = 1`) requantizes after **every** token.

At the last prefill position this cadence mismatch drifts the state (post-attn norm reldiff 0.41%, individual `x_norm` elements up to 483%), which amplifies through rmsnorm → router GEMM → top-K into **zero-overlap expert selection** → echo. (MambaQuant-style amplification through the DeltaNet recurrence + MoE routing.)

The function's doc comment claimed it did "N offset-pointer launches, bit-exact with N × decode" — code that was never written. The actual launch was single-shot with `n_tokens = N`.

Not a regression from any in-flight refactor — reproduced and fixed against `master`.

## Fix

Move the requant **into** the kernel token loop: quantize → dequantize the state after every token, with the global store gated to the last token only.

- **Single launch preserved** — no per-token launch overhead, so the batched-prefill throughput win is retained (rejected the host-loop alternative because it is O(N) launches per LA layer).
- **Reconciled with the per-launch `frame` dither** (added on master in the a3b-hunt2 fix to de-bias decode stochastic rounding): the Q8 requant seeds on `frame + t` (host frame counter + in-launch token index), so every per-token requant in the batched launch gets fresh, data-independent entropy. Decode (`n_tokens=1`, `t=0`) reduces to plain `frame`. Result is **distributionally equivalent** to decode (the requant *cadence* is what fixes the echo), not byte-identical.
- Applied to both `gated_delta_net_q8.hip` and `gated_delta_net_q4.hip` (same latent bug; Q4 reachable via the `StateQuant::Q4` batched arm). Q4 keeps its `amax`-derived seed since upstream did not add a frame to the q4 kernel — its batched path is bit-exact with q4 decode. `gated_delta_net_f32` left unchanged (F32 state is exact).
- The fix lives entirely in the kernels; the host-side `frame` plumbing in `norm.rs` is upstream's and unchanged.

Rebased onto current `master` (resolves the overlap with the `frame` dither change).

## Verification

- Build: `cargo build --release --features deltanet --example a3b_smoke_forward` — clean.
- `a3b_smoke_forward` on `shisa-Qwen3.6-35B-A3B-PARO-packed`, batched (`HIPFIRE_SMOKE_PREFILL_BATCH=1`): coherent output matching the per-token baseline (`\nHere's a thinking process:\n\n1. **...`). Echo gone. Re-verified after the rebase/`frame` reconciliation.
- Batched-prefill perf preserved (~0.9 s vs 10.8 s per-token argmax).
- `scripts/coherence-gate.sh --full` (pre-rebase): all Qwen models OK, including the DeltaNet `qwen3.6-27b`.

## Note for reviewers (gfx1151 box)

The pre-commit spec gate also runs the DeepSeek-V4 MTP battery, which hard-fails on this machine due to a **pre-existing, unrelated** gfx12 WMMA builtin (`gemm_q8_0_wmma_gfx12` needs `gfx12-insts`/`wavefrontsize32`) that cannot compile on gfx1151. This PR touches no deepseek files and deepseek4 does not use `gated_delta_net`, so that gate failure is independent of this change.


fivetide:
## Additional finding: gemm_f32_batched layout bug

During review I found a **separate compounding bug** in the batched prefill path: `gemm_f32_batched` writes its output in head-first `[M×N]` layout (`Y[(size_t)m * N + n]` at `kernels/src/gemm_f32.hip:31`) but all downstream GDN consumers read token-first `[N×M]`.

On ParoQ4G128 models, `w_alpha`/`w_beta` are F32 (unquantized) and route exclusively through this kernel. The transposed output corrupts the alpha/beta gate values per-token (cos ~0.49 vs per-token baseline), which feeds wrong scaling into the GDN recurrence — even with the per-token requant fix in this PR.

**Fix applied** in commit `a1881fcd` on this branch (also pushed to `fivetide:fix/paro-batched-prefill-gdn-requant`):
- `kernels/src/gemm_f32.hip:31`: `Y[(size_t)m * N + n]` → `Y[(size_t)n * M + m]`
- `crates/hipfire-runtime/src/dflash.rs:637`: swapped A/B arguments (the old call happened to produce correct output only under the buggy formula; the sibling `F16`/`HFQ4G256` calls already use the `(&w.buf, x, y, w.m, w.k, batch)` convention)

**Verification** (from `feature/dispatch-unification` dump probes):
- alpha/beta cos improved from ~0.49 to 1.0
- GDN output cos improved from ~0.989 to ~0.9999
- final hidden state cos improved from ~0.928 to ~0.9998

**Doc:** full investigation is in `.codeinsight+research/HANDOVER-gemm_f32_layout_bug--paro-batched-prefill-gdn-requant.md` on the branch.

fivetide:
Closing as superseded by master. Both fixes landed via different paths:

- **Per-token GDN Q8 requant** (the core fix): master now has a more complete implementation with a `requant_per_token` flag (per-token mode for PARO, fast single-requant mode for MQ4) plus error-feedback (EF/sigma-delta) support. The behavior this PR targeted is covered.
- **F32 batched GEMM output layout fix** (`fix(gemm): correct F32 batched GEMM output layout`): landed in master as commit `25720cff`.

The branch also had a very old base (pre `a7a8d89b`) that made rebasing onto current master impractical.

Issue #392, opened 2026-06-03
[mq4 quant] Productionize AWQ+GPTQ (proven −33% KLD on 9B) + mixed-precision bit-allocation


Kaden-Schutt:
# [mq4 quant] Productionize AWQ+GPTQ (proven −33% KLD on 9B) + mixed-precision bit-allocation

**Scope:** improve the *existing* G256 mq4 quant within its current structure. The
separate new-quant-container ("hfqv2") work is owned by the maintainer and is **not**
part of this issue. This issue hands off the implementation of two *validated* quality
levers and one *refuted* dead-end (so nobody re-walks it).

---

## Background — what the investigation established (2026-06-02)

The current G256-affine mq4 quant (256-wide block, one affine scale) asymptotes around
**~0.10 KLD** on dense 4-bit. AWQ and GPTQ are *within-grid* optimizers. We characterized
exactly how far they go and what the next lever is.

### Finding 1 — AWQ+GPTQ is a real, reproduced −33% KLD win (at zero extra bytes)
Reproduced exactly on Qwen3.5-9B (q8/FWHT KV, prefill, 512 chunks, top-K=256 bf16 oracle):

| variant | KLD | PPL |
|---|---|---|
| flat MQ4 (G256 RTN) | 0.3215 | 8.72 |
| AWQ-alone | 0.1867 | 9.53 |
| **AWQ + GPTQ** | **0.1257** | 9.31 |
| Q8 reference floor | 0.0186 | — |

GPTQ-on-AWQ = **−32.7% KLD, lower PPL too** (clean win on both axes). **But this win
currently exists ONLY in the Python pipeline** (`scripts/mq4_masked_calib.py`). hipfire's
Rust GPTQ engine `crates/hipfire-quantize/src/gptq.rs` (1542 LOC, FP64, Frantar-OBS) is
**orphaned**: no `mod gptq;` declaration, `faer` is not a workspace dependency → it has
**never been compiled** and its ~24 unit tests have **never run**. The `*-gptq` recipes
wired in `main.rs` are a *simplified* sequential Lloyd-error-feedback (default damping 0.0
≡ plain Lloyd), **not** real OBS GPTQ. So the win is unshipped purely for plumbing reasons.

### Finding 2 — finer block grouping is NOT the lever; bit-depth is (REFUTED — do not chase)
A clean within-llama.cpp ladder (wikitext-2, 131K tokens, bf16 base):

| quant | KLD | × its Q8 floor |
[...]

PR #395, opened 2026-06-04
fix(engine): hunt-3 batch — 16 bugs (pp / sampling+MoE / daemon / bun-CLI) [DRAFT: pending A3B validation]

Kaden-Schutt:
## ⚠️ DRAFT — NOT YET GPU-VALIDATED

Builds green (cargo `--example daemon` + `bun build cli`), but the mandatory
coherence-gate + A3B behavioral validation has **not** run yet. This PR is
intentionally a draft pending a **hiptrx/gfx1201 + Qwen3.6-35B-A3B** run
(this box can't load A3B). Do not merge until the validation matrix below is green.

## What this is

Hunt-3: an adversarial multi-agent bug-hunt over the four engine surfaces the two
prior A3B swarms deliberately left untouched — **multi-GPU/PP**, **sampling + MoE
dispatch**, **daemon concurrency/streaming**, and the **Bun-CLI → backend entry**
(single-GPU A3B decode and the quant track #392 were excluded as saturated /
separate). 109-agent sweep → 3-skeptic perspective-diverse verify → 20 confirmed
findings. Then a **file-disjoint fix team → adversarial review → final fix**
pipeline (the review caught a real bug *in* a fix — see M-B). Catalog:
`docs/investigations/2026-06-03-hunt3-engine-bugs.md`.

## Fixes (16 distinct)

**Sibling-path divergence** — `generate_multi`/`generate_vl` were missing fixes `generate()` already had:
| id | bug | fix |
|----|-----|-----|
| H-A | `generate_multi` (pp>1) Jinja dropped the system prompt + skipped per-request cold-reset on turn 2+ (**the deferred #389 pp twin**) | ungate `try_jinja` + `jinja && seq_pos>0` cold-reset |
| H-F | `generate_vl` force-close emitted tokens with no committed-event → token-id stream desync | `emit_committed_event` in the force-close loop |
| M-A | `generate_multi` forward-error returns left DeltaNet dirty | call `reset_pp_uncommitted_state!` on both error returns |
| M-C | `generate_multi` had no grammar tool-call masking | lift `generate()`'s Matcher wrapper into its sample sites |
| M-D | `generate_vl` repeat-penalty scoped over prompt+gen | re-scope to generated-only |
| M-E | grammar/VL CPU sampler used a process-global wall-clock RNG (cross-request leak) | `llama::reset_cpu_sampler_rng(seed)` per request |
| M-F | OpenAI `stop` silently ignored | parse (cli→daemon) + match in `generate()`, `generate_dflash`, `generate_multi` |

**Memory-safety / leaks** (no `Drop` on `GpuTensor`):
| id | bug | fix |
|----|-----|-----|
| H-E | `forward_prefill_batch_multi` leaked per-band scratch on a later-band OOM (+ the residual intra-`new` partial-literal leak) | free already-allocated tensors on mid-loop alloc failure |
| M-B | NaN logit **panicked** the CPU greedy/argmax/top-k paths | NaN-safe argmax mirroring the GPU kernel (NaN never selected) — deepseek4 `sampling.rs`, `llama.rs`, `mtp_spec.rs`. *(The review caught that the first attempt, `unwrap_or(Less)`, still kept a trailing NaN — re-fixed with a `>`-fold.)* |

**Tokenizer:** H-C — `decode_hex_escapes` `byte as char` re-UTF8'd SentencePiece byte-fallback tokens into mojibake (Llama/SP-family only; Qwen unaffected) → byte-correct `decode_bytes` path + regression test.

**Daemon hardening:** H-D — clamp request-driven `max_seq` to `MAX_REQUESTED_SEQ` (512K) so a JSON-string `max_tokens` can't string-concat into a ~10M `max_seq` OOM. #5-sliver — dflash seed/prefill-abort now zeros DeltaNet (decode-abort already did, via the shipped H1 fix).

**Bun serve** (one stdin/stdout pipe, long-lived multi-client): H-B `recv()` throws `DaemonClosedError` instead of `process.exit` (one daemon crash no longer kills the whole serve; recovery in the generate catch sites **and** `drain()`); B-1 eviction takes the request lock; B-2 `stop()` awaits `proc.exited` before respawn; B-3 foreground-serve pidfile singleton-guard + read-back-compare cleanup; B-4 non-stream errors emit through the controller (no 255s hang); B-5 eviction resets capability state on non-EOF failure; B-6 `run()` HTTP-fallback only local-spawns when serve is gone; C-1 `body.stream === true`.

**Known-minor (deferred):** stop-string not stripped from emitted output + O(n²) re-decode per stop-using step — kept uniform with the shipped AR path, gated behind `!stop.is_empty()`.

## Validation matrix (run on hiptrx/gfx1201 + A3B before un-drafting)

- [ ] `./scripts/coherence-gate.sh` — fluent, on-topic, no attractor (touches forward/sampling/tokenizer)
- [ ] **H-A**: pp=2 + `HIPFIRE_JINJA_CHAT=1`, multi-turn with a system prompt — turn-2+ keeps the system prompt + answers cleanly (vs master rambling)
- [ ] **M-B**: NaN-logit / aggressive-quant greedy decode does not panic the daemon
- [ ] **H-C**: round-trip a CJK/emoji prompt on a SentencePiece/Llama model — bytes correct (no mojibake)
- [ ] **H-E / leaks**: VRAM flat over ~30 gens (pp>1 path) + baseline-return across load/unload cycles
- [ ] **H-B / bun**: kill the daemon mid-request under serve — other clients survive + serve self-restarts
- [ ] **perf**: A/B decode tok/s neutral vs master (all fixes are cold-path or gated)

### DIFF
--- a/cli/index.ts
+++ b/cli/index.ts
@@ -1093,8 +1093,25 @@ async function runViaHttp(
 
 // ─── Daemon IPC ─────────────────────────────────────────
 
+// hunt3 H-B: typed error thrown by recv() on daemon EOF in long-lived
+// (serve) mode. One-shot callers (run, bench, etc.) set `oneShot = true`
+// and recv() process.exit()s as before; serve leaves it false and catches
+// this to recover (500 + restart) instead of killing the whole serve for
+// all clients on a single daemon crash.
+class DaemonClosedError extends Error {
+  readonly code: number;
+  constructor(code: number) {
+    super("daemon closed");
+    this.name = "DaemonClosedError";
+    this.code = code;
+  }
+}
+
 class Engine {
   private proc: ReturnType<typeof spawn> | null = null;
+  // hunt3 H-B: one-shot mode (run/bench) → recv() exits on daemon EOF;
+  // long-lived serve leaves this false so recv() throws DaemonClosedError.
+  oneShot = false;
   private reader: {
     read(): Promise<{ done: boolean; value?: Uint8Array }>;
     releaseLock(): void;
@@ -1141,9 +1158,19 @@ class Engine {
         // unsupported environments, see #112) or via a real crash. In either
         // case, the daemon's stderr (which we inherit) already explained
         // what happened, so adding a Bun-rendered stack trace from here on
-        // top is pure noise. Exit cleanly with the daemon's own code (or 1
-        // if it hasn't exited yet) and let stderr stand on its own.
+        // top is pure noise.
         const code = (await this.proc?.exited) ?? 1;
+        // hunt3 H-B: in long-lived serve, a single daemon crash must NOT
+        // process.exit() the whole serve (kills every other client). Throw a
+        // typed error the serve handler catches → 500 + daemon restart. Clear
+        // buffered state and release the reader so a fresh start() is clean.
+        this.lines = [];
+        this.buffer = "";
+        try { this.reader?.releaseLock(); } catch {}
+        this.reader = null;
+        if (!this.oneShot) throw new DaemonClosedError(code);
+        // One-shot mode (run/bench): exit cleanly with the daemon's own code
+        // (or 1 if it hasn't exited yet) and let stderr stand on its own.
         process.exit(code === 0 ? 1 : code);
       }
       this.buffer += new TextDecoder().decode(value);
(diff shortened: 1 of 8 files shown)


Kaden-Schutt:
## ✅ GPU validation PASS — hiptrx / gfx1201 (4× R9700), fix daemon `fb36f539`

All seven matrix items green on real hardware (Qwen3.6-35B-A3B + dense models):

| check | result | evidence |
|-------|--------|----------|
| **coherence-gate** (`--full`) | ✅ PASS | "no hard errors"; A3B (3.5 + 3.6 mq4) answer the sheep riddle = **9** with fluent reasoning, no attractor/loop; 0.8b/4b/9b/27b dense + tool-call all coherent; clean `read` tool_calls |
| **H-A** pp=2 + Jinja multiturn | ✅ PASS | secret `ZEBRA-9931` planted **only** in the system message → recalled correctly on **turn 2 and turn 3**; `generate_multi` keeps system + clean multi-turn state |
| **H-C** tokenizer mojibake | ✅ PASS | `sp_decode_byte_fallback_cjk` + hex-escape tests pass (56/56 tokenizer tests); CJK byte-fallback roundtrips correct |
| **H-E / leaks** | ✅ PASS | summed VRAM **dead-flat at 28466 MB across 30 pp=2 gens** (0 MB drift) |
| **H-B** serve-crash survival | ✅ PASS | killed the daemon child mid-stream → **serve process survived** (no `process.exit`) and the next request returned a valid completion (self-healed, B-2) |
| **M-B** NaN / greedy no-panic | ✅ PASS (covered) | greedy temp=0 decode across the gate (incl. A3B 400-tok chain) — zero panics; argmax re-fix verified `>`-fold mirrors `argmax.hip` |
| **perf** A/B (9B greedy, q8 KV) | ✅ NEUTRAL | `probe_commits.sh 02634f4c fb36f539`: **100.8 → 100.5 tok/s (−0.3%)**, inside the ±1–3% session noise band |

deepseek4 entries skipped (model not present); pflash perf sub-gate intentionally skipped (it benches against *gfx1100* baselines — wrong arch, not touched by this PR).

Validated on an isolated worktree (`~/hipfire-hunt3-validate`). Ready to un-draft pending maintainer review.

PR #417, opened 2026-06-05
feat(detect): add long-state coherence detector

xynexus:
## Summary
- add `LongStateCollapse` to detect the first low-diversity 128-token window and distinguish early hard collapse from later soft warnings
- wire the new detector into `coherence_probe`, add repeat-penalty/window and state-quant request flags, and emit generated text/token IDs in JSON reports
- make speculative GDN tape replay honor the selected DeltaNet state quant mode instead of always using the Q8 replay path

## Validation
- `git diff --check origin/master..HEAD`
- `cargo test -p hipfire-detect attractor`
- `cargo check -p hipfire-runtime --example coherence_probe`
- `cargo check -p hipfire-arch-qwen35 --lib`


Kaden-Schutt:
Folded the FP32/Q4 DeltaNet **state** work onto this PR (it grew out of validating the replay-arm dispatch here). 3 commits: f32 batched+tree GDN kernels + dispatch-by-`dn_state.quant`, and the key perf fix — dropping the `prefill_batch_pbs_eligible` Q8-only gate that forced FP32/Q4 onto the per-token fallback. Result on gfx1100 27B-3.5/3.6: FP32 DFlash 33→147 tok/s (Q8 parity −3%), TTFT 5.6s→0.55s; Q8 regression-clean; 1.2e-7 spine-parity test. Dual-landing onto staging + master next.

PR #439, opened 2026-06-11
gemma4: rebase onto beta — 12B dense loads and generates; batched MQ6 GEMM

Kaden-Schutt:
## What this is now

This PR was a 156-commit draft against `master` from June. `beta` has since moved **1,125 commits** ahead of `master`, so replaying that history was not viable — it conflicted in exactly the files beta rewrote most (rdna-compute attention/dispatch/kernels, `llama.rs`, `qwen35.rs`, `speculative.rs`, `daemon.rs`, `hfq.rs`).

It has been **rebased onto `beta` and retargeted**: 7 commits, +21,713/−44. The original history is preserved on `feat/gemma4-unit-awq` (which carries all 97 of Kevin Read's commits) and on `feat/gemma4-kld`, `fix/gemma4-union-fixups`, `feat/gemma4-integrate`.

**Gemma 4 12B now loads, generates coherently, and is quantizable on hipfire.**

```
prompt: "Write a Python function that reverses a string, then explain it in one sentence."

```python
def reverse_string(text):
    return text[::-1]
```
This function uses Python's slicing syntax to create a new string by
stepping through the original string backwards from end to beginning.
```

`google/gemma-4-12B-it` → MQ4 (8.9 GB), gfx1201 / RX 9070 XT, 49.5 tok/s, clean stop on `<turn|>`.

## Approach

The diff split the work for us: **23,214 lines in 70 conflict-free new files** (the `hipfire-arch-gemma4` crate, 17 kernels, docs) versus **5,839/−615 across 48 modified files**. The new files were ported verbatim; the integration surface was re-implemented against beta's current shape rather than replayed.

Beta had also moved loading to a carrier registry, so the original `LoadedModel.gemma4_*` Option-field pattern is obsolete — state now lives in `ModelState::Gemma4(Gemma4Bundle)` alongside every other architecture.

## What works

- Load, arch detection (13 = `gemma4_text`, 22 = `gemma4_unified_assistant`), carrier + template
- Eager dense generate, chat template, stop tokens
- Quantize end to end: `Architecture: gemma4 (id=13)`, tied `lm_head` protected at Q8, QK-norm and all norms at F16
- Batched MQ6G256 GEMM (new — MQ6 was GEMV-only engine-wide)

## What is gated off, and why

Both are gated because they are **wrong**, not because they are unfinished. Measurements are in the commit messages.

**hipGraph capture** (`HIPFIRE_GEMMA4_GRAPH=1` to re-enable). The captured decode path collapses into a token attractor after the first token:

```
graph ON : 49.75 tok/s   "Hello!s율 bawass율율 bawaky interracial율jal…"
graph OFF: 49.72 tok/s   "Hello! How can I help you today?"
```

That second string is byte-identical to the Phase-2 Gate 2 expected output recorded in `findings/`. The graph is worth 0.06% — inside noise — so correctness costs nothing. This is the failure signature `AGENTS.md` documents: a captured graph replaying dangling stack-pointer kernargs. All six Gemma4 dispatch helpers added here go through `launch_maybe_blob` and are capture-safe, so the offending raw launch is elsewhere in the decode body.

**EAGLE spec-decode** (`HIPFIRE_GEMMA4_EAGLE=1`). Wired end to end, then disabled: the accept rule is greedy-argmax, so at temp 0 the committed stream must be byte-identical to AR. It is not, on two different quantizations:

| target | AR | EAGLE | τ | ratio |
|---|---:|---:|---:|---:|
| uniform MQ4 | 52.81 tok/s | 44.28 | 1.531 | 0.84× |
| Q8 | 38.25 tok/s | 27.90 | 1.912 | 0.73× |

Both incorrect **and** slower. Divergence starts at the first committed token, which points at the seed hidden / round-1 verify rather than acceptance rate.

## Batched MQ6G256 GEMM

Engine-wide there was exactly one MQ6 kernel — `gemv_mq6g256.hip` — with no `gemm_mq6*` anywhere. MQ6 is the `Promote6` target used by Qwen 3.5/3.6 and Gemma4 for precision-critical tensors (`down_proj`, `v_proj`, edge FFN), so it was a decode-only format by accident of tooling, and every batched consumer hard-errored:

```
gemma4 batch v_proj: dtype MQ6G256 has no batched proj kernel
```

Added `kernels/src/gemm_mq6g256.hip` + dispatch. Verified bit-exact against the `gemv_mq6g256_prerotated` oracle at batch ∈ {1,2,3,4,5,7,8,9,12,15,16,17} across three shapes including a group boundary (`K%256==0, K%1024≠0`). Bit-exactness is expected rather than surprising: both kernels map each lane to the same eight weights, walk groups in ascending order, and reduce with the identical 16/8/4/2/1 shuffle tree.

Reviewed adversarially; review caught that the verifier printed errors but never asserted, so a wrong kernel would have exited successfully and NaN would have slipped through entirely. Now asserts finite + bit equality per element, and is proven to fail — injecting +0.5 into one element panics with the index and both bit patterns.

## Shared-code discipline

Gemma must not change behaviour for an already-implemented architecture. Five violations were introduced during the port and removed:

1. **`llama.rs`** — five `head_dim == 256` guards widened to `256 || 512` for *every* arch. Reverted; `llama.rs` now has **zero deletions** vs beta. Gemma uses `new_gpu_fwht3_capped_filtered_gemma4` / `new_gpu_asym3_*_gemma4` doors delegating to a shared private inner.
2. **`kmap_mode == 2`** — typed mode mutated to stop promoting attn q/k/o on edge layers, which changes output **bytes** for any model already quantized `--kmap-mode typed`. Mode 2 restored byte-identical; Gemma's variant is a new **mode 3** (`--kmap-mode typed-gemma4`).
3. **`carriers.rs`** — two `maybe_screen_mmq` calls deleted from the **Qwen35 and LLaMA** load paths, commented "Gemma4Weights does not yet impl MmqScreenable". Restored.
4. **`sp_dummy_prefix`** — SentencePiece dummy-prefix went unconditional → config-derived defaulting **false**, silently re-tokenizing existing SP models. Silence now returns true (historic LLaMA convention); Gemma declares a normalizer so it resolves false on the derived path.
5. **`is_gpt2_bpe`** — made conjunctive, so it can only flip a verdict where `▁` strictly outnumbers `Ġ` (a SentencePiece vocab already being misclassified). Gemma has 1 `Ġ` against ~137k `▁`.

Every remaining deletion in shared code was inspected: condition extensions for `router.proj.weight` (a Gemma-only tensor name), the new parse arm, an import, and doc comments.

## Verification

- `cargo check --workspace --all-targets`: 0 errors
- `cargo test --lib --workspace --locked`: **1807 passed, 0 failed** (beta baseline 1800)
- GPU: gfx1201 only — see roadmap comment for the cross-arch gap

## Not done

- **No registry entry.** There is no published hipfire-quantized Gemma4 artifact on HF, so a row would point at a 404. What needs publishing is recorded rather than faked.
- Lowered/MoE decode (26B-A4B) is loaded but not wired — fails loudly.
- 31B quantizes (23.3 GB) but exceeds a 16 GB card; unverified.
- E2B/E4B need PLE (`hidden_size_per_layer_input`, `num_kv_shared_layers`) — genuinely separate work.

Co-authored-by: Kevin Read <me@kevin-read.com>


HUSRCF:
# Suggested PR Comment: Gemma4 E-Series Compatibility Guard

Thanks for putting the Gemma 4 support together. I independently ran a **functional sanity check of the standard 12B dense path**.

I noticed one possible compatibility issue with the public Gemma 4 E2B and E4B variants. I may have missed an existing check elsewhere, so please correct me if this is already handled.

The quantizer currently maps both `gemma4` and `gemma4_text` to arch 13. However, the current arch-13 configuration does not appear to represent the additional E-series semantics used by the public E2B/E4B configs:

- E2B: `hidden_size_per_layer_input=256`, `num_kv_shared_layers=20`, `use_double_wide_mlp=true`
- E4B: `hidden_size_per_layer_input=256`, `num_kv_shared_layers=18`, `use_double_wide_mlp=false`

At commit `86ab8d94a8ba79f18f85fc9c267e82ec57ee200b`, an E2B/E4B checkpoint is currently accepted as arch 13 and reaches the dense execution path even though these fields are not represented by `Gemma4Config`. I reproduced the following next-token top-1 token IDs using the same input IDs `[2, 9259]`; the E-series-aware column loads the same PR-produced Q8 HFQ artifact used by the current PR runtime.

| Case | Current PR runtime, Q8 HFQ | Official BF16 reference | E-series-aware runtime, same PR-produced HFQ |
| --- | ---: | ---: | ---: |
| Gemma 4 E2B | 226640 | 236888 | 236888 |
| Gemma 4 E4B | 242604 | 236888 | 236888 |
| Gemma 4 12B | 575 | 575 | n/a |

A Q8/BF16 top-1 difference alone would not establish an architectural issue. However, the same PR-produced E-series HFQ artifacts produce the BF16-matching token in an E-series-aware runtime, while **the 12B Q8 control matches BF16 in the current PR path**. The official BF16 loads also reported no missing, unexpected, or mismatched keys; the current PR runtime's E2B/E4B text generations were incoherent, while the 12B path produced a coherent answer.

As an additional dense-path sanity check, I ran a 12-case Gemma 4 12B Q8 matrix on an RDNA3 W7900 across Chinese and English, thinking enabled and disabled, and three task types. **All 12 cases terminated normally and produced task-appropriate outputs.** This is functional side verification, not an end-to-end performance claim.

Would it make sense to add a small feature-aware guard before quantization/loading, for example `hidden_size_per_layer_input > 0 || num_kv_shared_layers > 0`, and fail early with a targeted unsupported-variant message until an E-series-specific path is available? `use_double_wide_mlp` alone would not identify E4B, so the first two fields appear to be the safer routing guard; the double-wide flag remains an additional E2B execution semantic that a future implementation would need to preserve.

A small regression fixture for both public E2B and E4B configs could keep the routing boundary explicit. Longer term, these fields could select a separate E-series variant implementing PLE, attention-aware KV source mapping, and the E2B double-wide MLP behavior.

I would be happy to prepare the minimal fail-fast guard and config tests as a separate small PR if that would be useful. Separately, I have both E2B and E4B running in one of my forks and would be more than glad to share the implementation if it would help.

PR #446, opened 2026-06-11
feat(cohere2moe): Cohere2-MoE / North-Mini-Code-1.0 support (arch_id 12)

nwoolmer:
## Summary

Adds the **Cohere2-MoE** architecture (`arch_id 12`) to hipfire, serving **CohereLabs/North-Mini-Code-1.0** (30.5B total / 3B active, 128 experts top-8). New crate `hipfire-arch-cohere2moe`, end-to-end: quantize → forward → KLD/PPL sweep → daemon serving + agentic tool-calling. Validated on **gfx1151 (Strix Halo, RDNA3.5)**.

## Architecture

Three Cohere2 novelties handled (verified against the HF `modular_cohere2_moe` reference):
- **Parallel block** — a single RMSNorm feeds **both** attention and the MoE/dense branch; both add into the residual (`h = h + attn(LN(h)) + ffn(LN(h))`).
- **Interleaved attention** — 13 `full_attention` layers are **NoPE** (global, no positional embedding); 36 `sliding_attention` layers use **interleaved (GPT-J) RoPE** with a 4096 window. Dense prefix layer 0 gets `force_rope`.
- **Sigmoid top-8 router** (`norm_topk_prob=false`), dense prefix layer (`first_k_dense_replace=1`).
- **RMSNorm** (not base Cohere2's mean-centered LayerNorm), tied embeddings, native **bf16** weight dtype.

## Performance

Measured on **gfx1151** (AMD Ryzen AI MAX+ 395 "Strix Halo", RDNA3.5), greedy decode, via the `infer` example — the same `forward_batch` (batched prefill) and `decode_step` paths the daemon serves. Each cell is DPM-clock-warmed (`HIPFIRE_DPM_WARMUP_SECS=10`) over pre-compiled kernels; prefill prompts are deterministic exact-length token sequences (throughput is content-independent). Reproduce with **`scripts/bench-cohere2moe.sh`**.

The prefill path: **WMMA Q8 attention projections** + **batched `forward_batch`** (each weight read once per ≤64-token chunk) + **O(1)-LDS windowed flash Q8 attention** (removes the old ~16K LDS ceiling; sliding-window mask so context beyond the 4096 window attends correctly). Dev-measured deltas that built it (single-shell A/B at the cited commits — *not* `probe_commits.sh` fresh-process-attested, unlike the measured tables below): `forward_batch` ~2× (`031820fc`), WMMA Q8 projections ~9× (`93e6ce68`), batched daemon prefill ~15× (`6fa063ff`).

**Prefill throughput (tok/s)** — batched for the MQ4/MQ6 ship tiers, per-token for the Q8 reference†:

| Tier (size · bpw) | 512 | 2 048 | 8 192 | 16 384 | 32 768 |
|---|--:|--:|--:|--:|--:|
| **MQ4** · 16 GB · ~4.25 bpw · batched | **673.7** | 339.1 | 119.9 | 71.6 | 40.9 |
| **MQ6** · 23 GB · ~6.5 bpw · batched | 219.7 | 169.5 | 87.7 | 59.8 | 36.2 |
| Q8 · 31 GB · 8.5 bpw · per-token | 34.5 | 33.1 | 31.2 | 28.8 | 24.2 |

**Decode throughput (tok/s)** — single-token `decode_step` at the given context depth:

| Tier (size) | 512 | 2 048 | 8 192 | 16 384 | 32 768 |
|---|--:|--:|--:|--:|--:|
| **MQ4** · 16 GB | **50.4** | 47.3 | 39.8 | 31.2 | 23.3 |
| **MQ6** · 23 GB | 46.4 | 44.9 | 37.5 | 30.1 | 21.8 |
| Q8 · 31 GB | 33.4 | 32.3 | 28.1 | 24.6 | 18.6 |

† **Why Q8 runs per-token.** The batched prefill's grouped-MoE expert GEMM exists only as the FWHT-pre-rotated *indexed* kernels for the MQ4/MQ6 quant family (shared with the qwen35/lfm2/minimax path). Q8 (and the bf16 oracle) experts have no indexed/grouped kernel, so `forward_batch_supported` returns false and the whole model falls back to a per-expert `weight_gemv` loop per token. Q8 is the near-lossless *reference* tier, not a ship target, so the batched expert path was never built for it — the attention Q8 projections still use WMMA-Q8 even there.

**Takeaways:**
- Batched `forward_batch` lets the ship tiers prefill at **220–674 tok/s short-context** — roughly an order of magnitude over the per-token Q8 reference (~34 tok/s), reflecting both the batched grouped-MoE path *and* MQ4/MQ6's lighter expert GEMMs. The gap closes at 32K, where the 13 NoPE full-attention layers dominate and all tiers land at 24–41 tok/s.
- Lighter quants also **decode faster** (less weight bandwidth + the indexed MoE kernel): MQ4 leads at every depth; Q8's near-lossless quality costs ~33% decode throughput vs MQ4 at short context.
- **Ship reco (unchanged):** **MQ6** for code-gen — best quality (+1.0% PPL, see Quality) at 88–220 tok/s prefill / 37–46 tok/s decode short-context; **MQ4** when size/throughput matter most.

## Serving robustness + long-context prefill (follow-up)

A serving pass against an external agentic harness (which surfaced stalls, ~180 s response timeouts, and reasoning-only turns) added three fixes:

- **Windowed prefill → O(window).** The sliding-window flash kernel skipped the out-of-window dot products but still launched the full-sequence tile grid + reduce and ran Phase D (which loads V even for masked positions). An out-of-window-tile early-return makes sliding-layer attention **flat ~12 ms @ 32K vs 68 ms (5.5×)**; full-causal (every other arch) byte-identical. This is the dominant fix for the long-context prefill falloff / timeouts. *The Performance tables above predate it — the 16K/32K prefill + decode columns are now faster.*
- **Empty-turn guard.** North occasionally emits `<|END_THINKING|>` → `<|END_OF_TURN_TOKEN|>` with no response (a reasoning-only turn); the daemon masks that EOS and re-samples to force `<|START_TEXT|>`/`<|START_ACTION|>`.
- **Think-budget force-close.** A heavy reasoner that out-thinks `max_tokens` mid-`<think>` returned reasoning only; the daemon reserves answer budget (honoring `max_think_tokens`) and force-closes thinking so it always returns a result.

Verified: `q8_windowed_attn_test` + the windowed microbench (`WINDOW` knob); cohere2moe + qwen coherence gates clean; A/B empty-turn reproductions. Opt-outs: `HIPFIRE_C2M_EMPTY_TURN_GUARD=0` (the two guards); the kernel change is `window>0`-gated.

## Quality

- **KLD/PPL sweep** (bf16 oracle vs q8/mq6/mq4, 2048 wikitext tokens): **Q8 lossless** (PPL == bf16 @3dp), MQ6 +1.0%, **MQ4 +9.3%** (heavy p99 tail → prefer MQ6 for code-gen). See `docs/plans/kld-measurements-master.md` §3A.
- **EM-drop fix** (`19304fbb`): the forward diverged from the reference on two points — mean-centered LayerNorm instead of RMSNorm, and a missing layer-0 `force_rope`. Fixing both is a **−6.2% PPL** gain on held-out prose (6.824 → 6.398 on identical tokens), and restores long-context needle recall at greedy.

## Serving / agentic

- Daemon serving: list-form chat template (default/tool_use/rag), `<|END_OF_TURN_TOKEN|>` EOS, agentic tool-call marker state machine.
- cli serve honors the daemon's authoritative `reasoning` flag (visible answer was previously trapped in `reasoning_content`); per-model sampling recipes; interleaved-thinking reasoning preserved across multi-turn agentic turns; marker-leak hardening; oversized-prompt OOB guard + holistic reset.
- **Degenerate-output loop-guard** + env-gated long-context diagnostics (`longctx_probe`, `HIPFIRE_C2M_NORMDUMP`, `HIPFIRE_C2M_DUMP_PROMPT`).

## Status — DRAFT

**Owed before ready for review:**
- [ ] **Dual-arch coherence gate** (mandated): **gfx1100 (RDNA3) + gfx1201 (RDNA4)**. Currently validated only on gfx1151.
- [ ] Recorded clean run of `scripts/coherence-gate-cohere2moe.sh`.

**Under investigation (not blocking the arch, tracked here for transparency):**
- A `<PAD>` degenerate-output incident during agentic serving was traced to a **transient/operational serve state**, *not* a deterministic forward/model bug: the identical ~2500-token prompt generates coherently after a fresh restart; bf16 == mq4 and both attention kernels (batched + per-token) behave identically; unique content is healthy to 32K. The loop-guard converts any such attractor into a fast failure instead of a multi-minute hang. A daemon self-heal (auto-reload on repeated degenerate output) is a possible follow-up.

## Test plan

- [x] Coherence-validated forward (gfx1151)
- [x] KLD/PPL sweep across tiers
- [x] Agentic tool-calling end-to-end (multi-turn, tool results, reasoning preservation)
- [ ] Dual-arch gate (gfx1100 + gfx1201)


nwoolmer:
Addressed on `nw_cohere2moe_support` — benchmark/perf in `8b1c3194`, review fixes in `544f0af5`. Per finding:

**2 — Jinja fallback serves ChatML garbage** — fixed (`544f0af5`, `daemon.rs`). The cohere2moe serve path no longer falls through to a Plain ChatML frame. Both the render-failure arm **and** the `else` branch (jinja disabled via `HIPFIRE_JINJA_CHAT=0`, or a `.hfq` with no template) now `emit_error_with_id(...)` and `return`, so the client gets a `{"type":"error"}` instead of off-distribution decode. The `HIPFIRE_JINJA_CHAT=0` footgun is covered.

**3 — Silent OOB expert clamp** — fixed (`544f0af5`, `forward.rs::moe_per_expert`). `.min(m.experts.len()-1)` → explicit bounds check returning `Err("… router produced OOB expert id …")`.

**4 — `_qt_u` discarded in expert fuse** — fixed (`544f0af5`, `cohere2moe.rs`), one step past the suggested `debug_assert`: a hard load-time `Err` on `qt_g != qt_u`. A `debug_assert` wouldn't fire in release (where the models actually run), and a mis-tagged byte-fuse is unrecoverable garbage, so it should fail loudly at load in all builds. The check can only reject a genuinely mixed-dtype export, which the single-tag fuse can't represent anyway.

**6 — Doc mismatches** — fixed (`544f0af5`, `forward.rs`). Algorithm block now says `rmsnorm(...)` not `layernorm_meancentered(...)`; removed the stale "windowed-mask path … is a follow-up" lines (the windowed path is implemented and in use in this PR).

**5 — Perf claims unattestable** — addressed (`8b1c3194` + PR body). Replaced the bare ratios with a measured **prefill + decode throughput matrix** (MQ4/MQ6/Q8 × 512–32768) on gfx1151 — DPM-warmed (`HIPFIRE_DPM_WARMUP_SECS=10`), deterministic prompts with logged md5s, reproducible via `scripts/bench-cohere2moe.sh`. The `~2×/~9×/~15×` ratios are retained but now explicitly labeled dev-measured single-shell A/B at their commits (not `probe_commits.sh`-attested). The body also now documents *why* Q8 prefill is per-token (the grouped-MoE GEMM is MQ-family indexed-only; Q8 experts have no indexed kernel).

**1 — Coherence gate (blocker)** — partial. Re-ran `scripts/coherence-gate-cohere2moe.sh` on **gfx1151** against the post-fix daemon (`544f0af5`): **clean, exit 0**, all four rows coherent —
- cap → "Paris is the capital of France." (50.2 tok/s)
- reason → 9 sheep, correct reasoning
- code → `def square(n): return n * n`
- long-context → 5.7k-token prompt through the windowed flash path, OK, no error/marker leak (44.6 tok/s, 30.4 s prefill)

The mandated **gfx1100 (k9lin, RDNA3) + gfx1201 (hiptrx, RDNA4)** gate runs and `coherence-gate.sh` on gfx1100 still require those boxes — that's the standing owed item; the PR stays DRAFT until they're recorded.

Kaden-Schutt (approved):
Maintainer triage pass after merging current `origin/master` into the PR branch and fixing CI fallout.

Verified on head `362b108bff7f3c7e516292237bf07e21f1a96407`:
- `cargo check -p hipfire-loader`
- `cargo check -p hipfire-runtime --example daemon --features deltanet`
- `cargo test -p hipfire-runtime --example daemon toolcall_robustness_tests --features deltanet`
- `GITHUB_ACTIONS=1 GITHUB_EVENT_NAME=pull_request GITHUB_BASE_REF=master bash scripts/ci-rustfmt-changed.sh`
- `python3 scripts/registry_gen.py --check`
- `bun test cli/registry_loader.test.ts`
- `python3 scripts/check-env-docs.py`
- `bun test`
- `bun run typecheck`

GitHub required/no-GPU checks are green. Claude review is advisory and failed from `error_max_turns` without posting inline comments.

PR #450, opened 2026-06-13
fix(gpu-lock): flock-based stale-proof GPU mutex

fivetide:
## Problem

Validation gates and benches repeatedly hung waiting on a **stale** `/tmp/hipfire-gpu.lock`. Root cause in `scripts/gpu-lock.sh`: the lock was just a file holding a human-readable string, with **no owner identity and no liveness signal**, and `gpu_acquire` polled **forever** with zero staleness detection. Whenever a holder died without running `gpu_release` — `kill -9`, agent cancellation, crash, OOM, terminal close (none of which reliably fire the `trap … EXIT` the gates rely on) — the file persisted and every waiter blocked indefinitely.

Two existing workarounds confirmed the pain and were themselves buggy:
- `serve-restart.sh` blindly `rm -f /tmp/hipfire-gpu.lock`.
- `pp-gate.sh` skipped locking entirely when the file merely *existed* — unable to tell "my parent holds it" from "a rival agent holds it", so a stale file made it silently run **unlocked**.

## Fix: flock-backed mutex (stale-proof by construction)

`flock(1)` is now held on an open fd, so the **Linux kernel releases the lock automatically the instant the holder dies, for any reason**. Staleness becomes structurally impossible — no PID-reuse race, no manual recovery. PID/host/timestamp are still written into the file for `gpu_status`/diagnostics.

- **Bounded wait** — a genuinely-busy *live* holder makes waiters give up after `GPU_LOCK_TIMEOUT` (default 1800s, `0` = forever) and exit non-zero, instead of hanging.
- **Reentrant** (`HIPFIRE_GPU_LOCK_OWNER`) — nested gates (e.g. `pp-gate` under `speed-gate`) no longer self-deadlock, and a child never releases its parent's lock. Also fixes `pp-gate.sh`'s silent-unlocked-run bug.
- **Never deletes the lockfile** — unlinking an flock'd file lets a second acquirer lock a fresh inode (two holders). Dropped the dangerous `rm` from `serve-restart.sh`.
- `HIPFIRE_GPU_LOCKFILE` env override (used by the test to avoid touching the real lock).

The public API (`gpu_acquire`/`gpu_release`/`gpu_status`) is unchanged, so all 8 gate scripts (coherence, coherence-dflash, coherence-deepseek4-mtp, agentic, agentic-jinja-tools, pflash, pp, speed) and ~18 bench/probe scripts that already source `gpu-lock.sh` pick up the new behavior with **no changes**.

## Files
- `scripts/gpu-lock.sh` — rewritten (flock + reentrancy + timeout + lockfile override)
- `scripts/serve-restart.sh` — dropped the lockfile `rm`
- `scripts/pp-gate.sh` — unconditional acquire (reentrancy handles nesting)
- `scripts/test-gpu-lock.sh` — **new** no-GPU regression test
- `CLAUDE.md`, `docs/skills/serve-restart/SKILL.md` — doc updates

## Verification

`bash scripts/test-gpu-lock.sh` (no GPU required) — all 4 pass:

```
PASS: acquired immediately after holder was killed      (stale self-heal — the bug)
PASS: waiter returned 2 (timeout) instead of hanging
PASS: child reentrant no-op; ancestor stayed locked
PASS: file persists, status='gpu is free'               (no-rm invariant)
```

Real-environment smoke on the actual `/tmp/hipfire-gpu.lock` (gfx1151): cross-process `gpu_status` reports BUSY with metadata while held; a `kill -9`'d holder's lock auto-heals so a fresh acquire succeeds immediately. `bash -n` clean on both scripts.


fivetide:
Superseded by #451. This PR was accidentally based on the diverged fork `fivetide:master` (1491 fork-only commits), inflating the diff. #451 is the same single commit rebased cleanly onto `Kaden-Schutt:master` — 1 commit, 6 files.

PR #452, opened 2026-06-15
fix(dispatch): dead-gate Paro4G128T fused entries — token-0 attractor on PARO A3B

fivetide:
## Problem

`master` (via `50437c6b`, dispatch-unification lineage) is broken for PARO A3B models: `shisa-Qwen3.6-35B-A3B-PARO-packed` collapses to a token-0 (`!`) attractor from the first decoded token under greedy decode.

## Root cause

`50437c6b` wired three `FusedQkv` / `FusedQkvza` / `FusedGateUp` **Paro4G128T** dispatch entries whose guards match plain `DType::ParoQ4G128` `WeightTensor`s. But the underlying kernels (`paro4g128t_quad_rotate` / `fused_*_paro4g128t_*` in `gemv_paro4g128.hip`) compute `pairs`/`theta`/`channel_scales` offsets **inside** the weight buffer — they require the self-contained PARO4G128T blob layout (the removed `quant_type` 28/29 probe format).

Runtime `ParoQ4G128` tensors are HFQ4G128-repacked rows with the Givens rotation in **separate sidecar buffers** (`WeightTensor.paro`), so the rotate kernel read garbage rotation tables past the weight bytes. Every QKV/Z/A and gate/up projection garbled → logits collapsed → greedy argmax pinned at token 0.

Bisect (good `6d51065b`, bad `e2d21ae2`, 885 commits, GPU repro per step) landed on the `50437c6b..08f6be64` pair. `HIPFIRE_FORCE_UNFUSED=1` being coherent on the broken tip is confirming evidence.

## Fix

Dead-gate the three paro fused guards in `crates/hipfire-dispatch/src/pipeline/steps.rs` (always return `false`) until a loader actually produces genuine self-contained T-format blobs. `ParoQ4G128` then falls back to the correct per-op Givens-aware GEMV path (`run_auto`).

## Verification

- `cargo test -p hipfire-dispatch`: 144/144 pass (paro guard-rejection + table-existence tests green).
- Daemon repro on humaneval_0 (md5 `5333a1f7`): coherent 40-token continuation, ~39 tok/s decode (was 11× `tok_id:0` then stop).
- `coherence_probe` ×4 matrix (humaneval_0 / lloyd_long / tool_call_read_file / lru_cache_pep8_strict): all detectors OK, 0 hard fails, no attractors.

Cherry-picked from `b9375747` on `feature/paro-transparent-loading`; `steps.rs` is byte-identical between the fix's parent and current `master`, so this applies cleanly.


fivetide:
Superseded — commit 74be53bc already landed in master via PR #463 (feature/transparent-loading-all-models).

Issue #462, opened 2026-06-21
Serve regression: PR #455 left qwen3.5 DeltaNet/KV bundle migration incomplete → </think> attractor on multi-request serve


Kaden-Schutt:
## Summary
PR #455 (Unified model loading — carrier registry + WeightBackend, commit `980970fa`) moved qwen3.5 (arch 5/6) DeltaNet + KV recurrent state from `LoadedModel` direct fields (`m.dn_state` / `m.kv_cache`) into `ModelState::Qwen35(Qwen35Bundle{ kv_cache, dn_state })`, **migrated the forward/generate path to read the bundle, but did NOT migrate the daemon's reset / checkpoint / prefix-cache / abort state machinery** — those keep reading the now-always-`None` direct fields, silently no-op, so recurrent DeltaNet state **bleeds across requests**.

## Symptom
Multi-request serve produces a `</think>` thinking-loop attractor (garbage). **Catastrophic on the DFlash spec-decode path** (~6/8 requests garbled), **mild on AR** (~1/8). The *first* request after a restart is clean; contamination accumulates on the 2nd+ request.

## Why CI/gates missed it
Unit tests and `coherence-gate.sh` drive the daemon **single-request**, which is clean. Only **multi-request serve OUTPUT** exposes it. There is no multi-request serve-output test.

## Repro
Serve `qwen3.5:9b`, send repeated + distinct greedy (`temperature: 0`) `/v1/chat/completions` requests, and look at the **content**:
- **pre-455** (e.g. `5ca8ed83`): CLEAN — 8/8 correct, DFlash and AR.
- **master tip** (post-455): DFlash ~6/8 emit `</think>\n</think>\n…`; AR ~7/8.

## Fix
PR #460 (`feat/cli-serve-ux`) fix-forwards it: completes the migration across 16 daemon reset/checkpoint/abort sites to source `dn_state`/`kv_cache` from the bundle (no hoist — `unload_model` frees both the direct field AND the bundle, so aliasing double-frees; cache-hit continuation paths left untouched so multi-turn keeps working). Validated by serve output on gfx1100: **DFlash 8/8 + AR 8/8 + multi-turn continuation intact**.

## Action items
- [ ] Land #460 (fixes serve) — or temporarily revert master to pre-455 (`5ca8ed83`) until #460 lands.
- [ ] **Add a multi-request serve-output test to CI** — the missing guard that let this ship.

PR #455, opened 2026-06-15
Unified model loading: transparent quant + auxiliary-detail management (carrier registry + WeightBackend)

fivetide:
## Unified model loading — transparent quant + auxiliary-detail management

This branch reworks how hipfire loads a model **file** so that adding/serving a
model no longer means re-implementing quantization, FWHT rotation, PaRo sidecars,
or AWQ repacking per arch. The goal was a *reusable, transparent* loading path;
this PR lands the part of that goal that is built and verified, with an
**honestly-narrowed claim** and an explicit list of what reviewers with hardware
we lack need to validate.

> Scope note: this PR is about **loading the model file** (container → device
> weights + caches + teardown). It does **not** claim changes to inference/forward
> loops beyond keeping them byte-identical (proven below).

### What it achieves (the claim we assert)

**Per-tensor format transparency.** A per-arch loader now says:

```rust
wq: b.proj("self_attn.q_proj", q_dim, hidden)?,   // never sees quant_type / FWHT / PaRo
```

and the `WeightBackend` trait (`weight_backend.rs`) resolves dequant, FWHT
rotation, PaRo sidecar, and name layout behind it. Concretely:

- **One shared codec table** — `dequant_weight_raw` has a single definition driven
  by `RAW_CODECS`; a new quant format is one row and every arch gets it.
- **Container unified at the layer walk** — HFQ and safetensors/PaRo both run the
  same generic `load_layer<B: WeightBackend>`; the container difference collapses to
  which backend struct the carrier constructs.
- **PaRo is a transparent augmentor** (`augmentor.rs`, `DEFAULT_AUGMENTORS`).
- **Carrier registry** (`carriers.rs` + `lib.rs`) replaces the old `match arch_id`
  ladder; a vanilla dense HFQ arch is a one-line `HfqCarrier`.
- **Config via serde** (`RawConfig + finalize` per arch) and **compiler-enforced
  teardown** (`ModelState` exhaustive match → no leaked VRAM on unload).

### Key decisions (and why)

- **No single `ArchSpec` / unified step contract (N1 REJECTED).** Explored across 3
  review rounds + 2 ADHD passes; the divergence across spec/VL/EP pipelines is
  irreducible (2 spec result shapes × bespoke draft × transient-ModelSlot
  lifecycle), so a god-struct would *hide* cost, not remove it. We extracted the
  *per-tensor* cost and deliberately left *per-model* wiring explicit. Reasoning
[...]

PR #463, opened 2026-06-21
Unified model loading + KV-cache usage unification

fivetide:
# Unified model loading + KV-cache usage unification

Diverges from `master` at `5ca8ed83` (pre-#455). 237 non-merge commits,
343 files, +60.7k/−17.0k. Two headline themes — **source-transparent model
loading** and **KV-cache usage unification** — plus new arch support, a MoE
pruning crate, and a set of genuine pre-existing-bug fixes.

---

### What
The daemon's monolithic ~700-line `load_model` arch-id ladder (HFQ-only) is
replaced by a **carrier registry** in a dedicated `hipfire-loader` crate, and
loading is made *source-transparent*: every supported arch loads from hipfire's
native `.hfq` **or** directly from a raw HuggingFace safetensors directory (raw
FP / ParoQuant / AWQ), through one dispatch point.

Two abstractions carry it:
- **`ModelSource`** (`loader_api.rs:15`) — `Hfq(HfqFile) | Dir(SafetensorsSource)`,
  auto-detected by `from_path`. Both back-ends implement the `ModelSource`
  trait (`model_source.rs:36`), so arch weight-loaders consume `&dyn ModelSource`
  and never branch on the container.
- **`Carrier`** (`hipfire-loader/src/lib.rs:36`) — one object-safe trait per arch
  with a pure `claims_arch_id(arch_id, is_dir)` predicate. `load_model`
  (`lib.rs:691`) opens the source and filters a static `REGISTRY` of 8 carriers,
  collecting *all* matches so an accidental overlap fails loudly
  (`"ambiguous carrier dispatch"`) instead of silently resolving by registration
  order. Unknown `model_type` → sentinel `UNCLAIMED_ARCH_ID` → clean
  `"no carrier for …"` rather than mis-routing into Qwen35 and dying deep in
  weight loading.

### Before → after
Before (`5ca8ed83`), `load_model` was an HFQ-only ladder in `daemon.rs`:
```rust
let mut hfq = HfqFile::open(path)?;
if hfq.arch_id == 7  { /* qwen2: ~50 lines inline tokenizer+kv+state */ }
if hfq.arch_id == 10 { /* minimax ... */ }
if hfq.arch_id == 5 || hfq.arch_id == 6 { /* qwen35 ... */ }
// each block re-duplicated tokenizer/template/skeleton glue; no dir support
```
After — registry-driven and arch-agnostic (`lib.rs:790`):
```rust
let src = ModelSource::from_path(path)?;          // Hfq | Dir, auto-detected
let carrier = REGISTRY.iter().filter(|c| c.probe(&src)) /* exactly-one or fail */;
carrier.load(src, &mut ctx)
```
Each carrier reduces the source axis to a small seam: `resolve_source_meta`
(`carriers.rs:34`) folds the source-varying/arch-invariant work (tokenizer, chat
template, arch_id), leaving a `match src { Hfq => …, Dir => … }` that yields
`(config, weights)` plus a shared tail.

### Source × arch matrix
| arch (id) | carrier | `.hfq` | raw safetensors-Dir | note |
|---|---|---|---|---|
| LLaMA/Mistral (0), Qwen3 (1) | Llama | ✅ | ✅ FP/Paro | raw-FP via transparent paro; BF16-complete readers |
| Qwen3.5 dense/+VL (5), MoE (6) | Qwen35 | ✅ | ✅ ParoSource | Dir drops DFlash/CASK with a warning |
| Qwen2 (7) | Qwen2 | ✅ | ✅ | dir derives to **7** so attention biases load |
| dots.ocr (8) | DotsOcr | ✅ | ✅ | Dir byte-identical to q8-HFQ ref |
| DeepSeek-V4 (9) | Deepseek4 | ✅ | ⚠️ unvalidated (no local ckpt) | |
| MiniMax-M2 (10) | Minimax | ✅ | ✅ | |
| LFM2.5-MoE (11) | Lfm2Moe | ✅ | ✅ | |
| Cohere2-MoE (12) | Cohere2Moe | ✅ | ⚠️ loads, MoE forward follow-up | |

ParoQuant/AWQ are **transparent**: `ParoBackend::proj` tries `.qweight` sidecars
first and falls back to the raw `.weight`, so a checkpoint with no
`quantization_config` loads as plain FP (`hfq.rs:1385`).

### Benefit
One path to load any supported model from HF safetensors *or* hipfire's
quantized formats (no separate import step); the container and arch-matrix
decisions are hidden behind `ModelSource`/`Carrier` (deep modules); adding an
arch = one carrier + one registry line; adding a source kind (e.g. GGUF) = one
`ModelSource` arm + one `resolve_source_meta` arm.

---

### What
A three-layer unification of how every arch picks its KV storage tier **and** the
matching attention kernel. End state: a single sealed dispatch family
(`hipfire-dispatch/src/families/kv_tier.rs`) owns the bool→tier→kernel decision,
reached through one accessor `KvCache::tier_inputs()` (`llama.rs:5386`).
- **L1** `tier_inputs()` — turns the cache's ~9 quant-mode booleans into a
  `KvTierInputs` in one place, incl. the non-obvious `quant_q4_residual()` /
  `is_hfq8_kv()` derivations that used to be re-derived (or omitted) per site.
- **L2** sealed `KTier` + `classify()` (`kv_tier.rs:27`) — the one sanctioned
  bool→tier decode; ad-hoc `if kv.quant_q8 {…} else if kv.quant_asym3 {…}`
  ladders become `kv.k_tier()`.
- **L3** `KvTierPlan::derive()` (`kv_tier.rs:218`) — derives the **paired**
  `(write_key, attend_key)` in one pass so they can't drift on tier (the
  `tiers_match` debug-assert), folding in per-position heuristics (q8 flash
  selection, qwen2 GQA-flash, batched/tree remap, cohere2moe sliding-window).

### Before → after (minimax, `5a5c28f9`)
Before — concrete, Q8-hardcoded kernel calls:
```rust
gpu.kv_cache_write_q8_0(&state.kv.k_gpu[l], &state.fa_k, ...)?;
gpu.attention_q8_0_kv(&state.fa_q, &state.kv.k_gpu[l], &state.kv.v_gpu[l], ...)?;
```
After — derive from cache state, execute as a step:
```rust
let plan = KvTierPlan::derive(KvTierInputs { pos, ..state.kv.tier_inputs() })?;
execute_steps(gpu, &ctx, &[Step::Attend { plan, io }])?;
```
The qwen35 sites collapse a hand-copied 9-field literal into
`..kv_cache.tier_inputs()` plus only the genuinely per-call fields.

### Coverage & benefit
Now routed through `derive`: llama, qwen2 (F32 GQA-flash decode), qwen35 (3
sites incl. MTP head), lfm2moe, minimax, cohere2moe (decode + batched prefill).
minimax/lfm2moe — previously Q8-hardcoded — transparently gain whatever tier the
cache is built in. Benefits: adding/fixing a KV mode is one edit in
`classify`/`derive`; write/attend can't drift (single-pass + assert); pinned by
**byte-identical regression tests** (`tier_inputs_q8_is_byte_identical_to_legacy_literal`,
`…_asym3_has_no_residual_q4_flag`) proving the refactor is a no-op on kernel
selection; `KvTierInputs` is GPU-free scalars → pure and unit-testable.

---

## 3. New architecture & capability support
- **Cohere2-MoE (arch 12, North-Mini-Code)** — new crate: BF16 forward, WMMA Q8
  prefill (~9×), batched prefill + daemon serving (chat template, tool calls),
  O(1)-LDS windowed flash-Q8 attention, per-model sampling recipes, coherence gate.
- **lfm2moe / deepseek4 / minimax** — Architecture trait + config/weights from
  safetensors + daemon wiring.
- **`hipfire-reap`** — new MoE pruning + selective re-quant crate (ReapPlan,
  byte row-gather, `--reap-bake` standalone `.hfq`, HfqFile overlay, mixed-tier
  bucketed dispatch).
- **Misc:** native BF16 decode helpers; no-numpy Rust KLD/PPL sweep analyzer;
  shared `dense_forward` over llama+qwen2; declarative `WeightSpec`/`WeightBackend`
  loading substrate.

---

## 4. Pre-existing bugs fixed (present at/before `5ca8ed83`)
These repair long-standing master code (not code this branch introduced):
- **`56439e62`** — Llama flash-attention partials buffer hardcoded to
  `max_chunks=16` (max_seq 2048); any KV cap >2048 (daemon default 4096) silently
  overflowed → OOB. Now sized to the runtime KV cap.
- **`d06fd32c`** — single-token `AttnFlashAsym4` kernel produces garbage at
  head_dim=128 but constructors accepted it → silent garbage. Now gated to
  head_dim=256 with a fail-loud load error.
- **`74be53bc`** — never-GPU-validated `Paro4G128T` fused dispatch entries fed
  sidecar-rotation tensors to self-contained-blob kernels → token-0 attractor on
  *all* PARO models. Dead-gated to per-op Givens-aware GEMV (root-caused via bisect).
- **`c9607144`** — tokenizer never honored HF `add_bos_token`; models needing a
  leading BOS got a degenerate attractor. Now prepends `bos_id` when the model's
  config sets the flag (gated; ChatML/Qwen unchanged).
- *(borderline)* **`8401485a`** — CLI reasoning split keyed only on `<think>`
  tags and ignored the daemon's authoritative `reasoning` flag; long-standing CLI
  code, latent until a flag-driven thinking arch (Cohere2) exercised it.

*(Branch-local regressions — fixes to code this branch added — are intentionally
excluded from this list.)*

---

## 5. Validation
- **Raw BF16 safetensors run coherently** on gfx1151 (daemon, greedy):
  Qwen3-0.6B and Qwen2.5-0.5B (downloaded from HF). qwen3 previously emitted
  token-soup from incomplete BF16 reader coverage — now fixed (`92e3c2b0`).
- **Serve bundle-state migration** (`9363d3a2`, cherry-picked fix for the #455
  master serve regression #462): 16 daemon reset/checkpoint/abort sites now source
  DeltaNet/KV from `ModelState::Qwen35` instead of the now-always-`None` direct
  fields (which leaked recurrent state across requests → `</think>` attractor).
  Validated here by **AR multi-request serve on gfx1151** (4/4 distinct prompts
  coherent, no cross-request bleed); the identical commit was **DFlash 8/8 + AR
  8/8** validated on gfx1100 by the original author. (Local DFlash re-run was
  blocked by an *unrelated* pre-existing `dflash.rs:795` scratch panic on the
  27B/gfx1151 config — fails on request 1, not a state-bleed.)
- dots-ocr Dir byte-identical to its q8-HFQ reference; cohere2moe Dir load
  validated on North-Mini-Code-1.0-pa4.

---

## 6. Deferred / future work
**KV unification:**
- Llama "Site A" attention dispatch (`hipfire-arch-llama/arch.rs:242`) not
  unified — its literal is a Q8-shaped contract assertion; Dir-loaded llama with
  asym KV mis-dispatches (`None` givens). Needs its own fix, not a no-op refactor.
- L2 flag **seal** not enforced — `KvCache` quant flags stay `pub`; hand-rolled
  ladders can still be written outside `KvCache`.
- `KvModePlan` cold/hot stamp **rejected on perf** (`derive` ≈ 2 ns/call,
  ~1000× below noise) — do not rebuild.
- qwen35 multi-GPU (PP>1) attention ladder not migrated (not validatable on this
  single-GPU box — run on hiptrx/hipx).
- minimax batched/prefill and qwen2 F32 prefill attention left manual
  (full-attention, not paired write+attend).
- Attention-boundary **fusion declared dead** for perf (no kernel spans
  projection→attention; a real win needs new per-(dtype×tier×arch) megakernels).
- Non-goals: deepseek4 MLA, qwen35 DeltaNet layers, lfm2moe conv-mixer, vision towers.

**Loading:**
- Llama-family PARO loader (`load_weights_paroquant_llama`) not migrated onto the
  shared `WeightBackend` walk.
- Multi-GPU safetensors/PaRo loading — `load_weights_multi` is HFQ-only (needs a
  `ModelSource` page-drop/band-routing equivalent).
- **DeepSeek-V4 Dir arm unvalidated** (no local checkpoint).
- **Cohere2-MoE Dir loads but does not yet run** — MoE forward dispatch follow-up.

---

## 7. Reviewer caveats
- The `hipfire-runtime` lib-**test** target had a pre-existing E0063 break
  (`KvTierInputs` legacy literal missing `q8_windowed`/`window`); fixed in
  `c57a6d68` so the whole lib-test target compiles again.
- DeepSeek-V4 Dir and Cohere2-MoE Dir-run are explicitly unvalidated (above).
- PP>1 paths need real multi-GPU hardware to validate.

---


fivetide:
### Follow-up flags from #462 addressed

**1. Multi-request serve-output regression test** (`7946c35b`)
Added `scripts/serve-multiturn-gate.sh` — the gate that would have caught #462. It loads a qwen3.5 (DeltaNet) model **once** and sends several distinct prompts in the **same** session (no unload between), running attractor detection on **every** request. The single-request coherence gates (`load→ONE generate→unload` per row) structurally can't see cross-request recurrent-state bleed.

Proven on gfx1151: against the **pre-fix** daemon the AR session degrades req1→req4 (`uniq 0.75→0.47→0.26→0.25`, req4 emits the literal `</think><think>` loop) → **exit 1**; against the **fixed** daemon all requests pass → **exit 0**. AR is the always-on guard; DFlash is opt-in (`HIPFIRE_SERVE_GATE_DFLASH=1`). Registered in the GPU-gate suite (`ci.yml`) + `CLAUDE.md`.

**2. DFlash `mq_x_rot` scratch panic** (`cacb97d6`)
While validating, the 27B DFlash path panicked `"MQ4 dispatch requires mq_x_rot scratch"`. Root cause: the carrier refactor moved DFlash draft loading into `hipfire-loader` and regressed `DflashScratch::new_with_mq(..., draft_weights.has_mq)` → the `with_mq=false` `::new` shortcut, so MQ-quantized drafts got no FWHT rotation scratch. Restored `has_mq` detection. Validated: 27B DFlash multi-request 3/3 coherent, no panic.

### Separate pre-existing issue uncovered (NOT this PR)
On gfx1151, the 27B DFlash **prefill** also has a prompt-length-sensitive `dst_offset + size <= dst.size` d2d-copy bounds assertion (`hip-bridge/ffi.rs:746`) — a short 3-prompt session passes, a slightly-longer 4-prompt one panics during req-1 prefill. It's robust on gfx1100 (where DFlash was validated 8/8). Filed as a separate DFlash-scratch-sizing follow-up; that's why the gate's DFlash arm is opt-in rather than always-on.

fivetide:
### Update: the "separate pre-existing DFlash issue" is now fixed too (`cd3ae386`)

Root-caused the `dst_offset + size <= dst.size` (ffi.rs:746) panic via the backtrace: `generate_dflash → seed_target_hidden_from_prompt_abortable → forward_prefill_batch_with_pbs_opts → forward_prefill_chunk → write_rows_to_staging`.

`load_dflash_state` sized the `HiddenStateRingBuffer` **staging** buffers to `max_n = block_size+1` (the verify-cycle width), but the prompt seed prefills the prompt in chunks up to `PREFILL_MAX_BATCH` and captures each into staging. `write_rows_to_staging`'s `n <= max_batch` guard was a `debug_assert` — silent in release — so a prompt longer than `block_size+1` (~17) tokens overran the buffer. That's the exact "What is 2+2?" (passes) vs "...Answer briefly." (panics) threshold.

Fix: size staging to `max(max_n, PREFILL_MAX_BATCH)`, and upgrade the guard to a hard `assert!` so any future undersize fails loud and named.

Both gfx1151 DFlash scratch bugs (this + the `mq_x_rot` regression) are now fixed, so `serve-multiturn-gate.sh`'s DFlash arm is back to **auto-run**. Validated on gfx1151: the previously-panicking 4-prompt 27B DFlash serve session runs **4/4 coherent** (τ 7.3/5.08/7.89/5.08), and the full gate (AR 4/4 + DFlash 4/4) exits 0.

PR #466, opened 2026-06-22
fix(qwen2): apply chat template in generate_qwen2 for instruct models

fivetide:
## Summary

- `generate_qwen2()` (arch_id=7) called `tokenizer.encode(prompt)` on **raw user text**, bypassing all ChatML scaffolding that other arch paths apply
- Instruct-tuned Qwen2 models require `<|im_start|>…<|im_end|>` framing; without it the model sees plain text and falls into a space/noise attractor
- Fix: when `m.chat_template` is populated (loaded from `tokenizer_config.json` by the Dir/safetensors path), apply it via `JinjaChatFrame::render()` before encoding; falls back to raw encode for base weights

## Root cause

`generate()` short-circuits to `generate_qwen2()` at line 6140 **before** the Jinja/ChatFrame scaffolding block (line 6682+). The qwen2 path was wired as a bring-up stub and deferred sampling + prompt formatting — this PR closes the prompt-formatting gap.

## Validation

Tested with **WeiboAI/VibeThinker-3B** (Qwen2ForCausalLM, BF16 safetensors dir, transparent loading):

```
bun run cli/index.ts run /path/to/VibeThinker-3B -t 0.7 -n 512 "What is 2+2? Be brief."
# → "4"   [106 tok, 8.17 tok/s]  (thinking stripped by CLI)

bun run cli/index.ts run /path/to/VibeThinker-3B -t 0.7 -n 1024 "Give me just the Python code for fast-doubling Fibonacci."
# → correct O(log n) implementation  [882 tok, 8.18 tok/s]
```

GPU: gfx1151 (Strix Halo, 131 GB VRAM), ROCm / HIP 7.2

## Test plan

- [ ] `hipfire run <qwen2-instruct-dir>` produces coherent output (not spaces/noise)
- [ ] `hipfire run <qwen2-base-dir>` (no chat_template) still works via raw encode fallback
- [ ] Existing HFQ qwen2 models (paro-qwen25-packed, qwen25-0.5b-instruct.mq4) unaffected — they load via the same arch_id=7 path, `m.chat_template` was already populated and is now used

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #465, opened 2026-06-22
fix(arch-llama): unify Site A attention dispatch + expand llama HFQ KV-mode policy

fivetide:
## Summary

Deferred items from PR #463 (transparent loading). Three commits:

### 1. `fix(arch-llama)`: unify Site A attention dispatch (`arch.rs:242`)

`Llama::forward_scratch_layers` in `hipfire-arch-llama` had a hardcoded Q8-shaped `KvTierInputs` literal (`quant_asym*: false`, `v_mode_bits: 8`, `givens_cos/sin: None`). For asym or HFQ8 KV caches this would either panic or silently select the wrong kernel.

**Changes in `crates/hipfire-arch-llama/src/arch.rs`:**
- Replace the literal with `..kv_cache.tier_inputs()` struct update (all quantization flags now come from cache state)
- Thread `givens_cos/sin` from the cache into `AttnParams` (asym attend kernels unwrap these)
- Pass `flash_partials: Some(&scratch.attn_partials)` (required for flash attend tiles)
- Add conditional `(k_scales, v_scales)` for HFQ8 (attend kernel panics on `None` for that tier)  
- Use `gpu.graphs.capture_mode` instead of hardcoded `false`

Byte-identical no-op for Q8 caches.

### 2. `test(dispatch-tests)`: KV-tier dispatch sweep

9 GPU-free tier-routing tests in `hipfire-dispatch-tests/src/llama.rs` covering Q8, asym2/3/4, FWHT+asym4, HFQ4, HFQ8, Q4, F32. Each verifies `uses_givens` and the paired `write_key`/`attend_key`.

### 3. `feat(kv-mode)`: expand `LLAMA_HFQ_POLICY` + wire `kv_mode_override`

- `kv_mode.rs`: `LLAMA_HFQ_POLICY` now accepts `[Q8, Asym3, Asym4]` (was `[Q8]` only). Asym2/Fwht* have no Flat KV constructor and silently fall to Q8.
- `llama.rs::KvCache::from_mode`: new clean-error gate for Asym3 alongside the existing Asym4 gate — at `head_dim≠256` returns a user-readable error instead of panicking in `new_gpu_asym3_capped`.
- `carrier.rs::load_bundle`: was hardcoding `""` for `kv_mode_override`, silently ignoring explicit `--kv-mode` requests for HFQ llama loads since the carrier was introduced. Now passes `ctx.kv_mode_override`.

[...]

PR #468, opened 2026-06-22
fix(daemon): add generate_dots_ocr_text for arch_id=8 text-only path

fivetide:
Fixes t_8fd736aa — dots-ocr generate panic in daemon.

## Problem
dots_ocr (arch_id=8) loads with `state: None` (no ModelState bundle) because it's a stateless one-shot vision encoder — its text decoder state lives in `m.qwen2_state`. However, two code paths in the daemon silently assumed `m.state` is `Some(...)` for all architectures:

1. **`generate()`** (text-only path): fallthrough `else` branch unwrapped `ModelState::Llama` — panicked for arch_id=8
2. **`bench_prefill`**: same pattern — panicked for arch_id=8

The VL image path (`generate_vl_dots_ocr`) was already correctly wired with a dedicated function.

## Fix
- Added arch_id=8 dispatch in `generate()` — routes text-only requests through new `generate_dots_ocr_text()`
- Added arch_id=8 arm in `bench_prefill` — uses `m.qwen2_state` + `m.dots_ocr_config`/`m.dots_ocr_weights` directly
- `generate_dots_ocr_text()` mirrors `generate_vl_dots_ocr` but skips image preprocessing, vision forward, and IMGPAD-token splicing — text-only tokenization, embedding lookup, batched prefill, greedy decode

## Verification
- `cargo check -p hipfire-runtime --example daemon` passes
- Smoke tested: daemon loads `dots-ocr.q8.hfq` successfully, text-only generate enters correct code path, KV budget guard works — no panic

### DIFF
--- a/crates/hipfire-runtime/examples/daemon.rs
+++ b/crates/hipfire-runtime/examples/daemon.rs
@@ -2515,6 +2515,19 @@ fn main() {
                         }
                     }
                     ok
+                } else if m.arch_id == 8 {
+                    // dots.ocr: Qwen2 text decoder via qwen2_state + dots_ocr fields.
+                    let state = m.qwen2_state.as_mut().unwrap();
+                    let config = m.dots_ocr_config.as_ref().unwrap();
+                    let weights = m.dots_ocr_weights.as_ref().unwrap();
+                    let mut ok = true;
+                    for &tok in &synthetic {
+                        if qwen2::forward_step(&mut gpu, &weights.text, &config.text, state, tok).is_err() {
+                            ok = false;
+                            break;
+                        }
+                    }
+                    ok
                 } else {
                     let ModelState::Llama(b) = m.state.as_mut().unwrap() else {
                         unreachable!()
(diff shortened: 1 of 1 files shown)


fivetide:
Rebased onto current master (`d44f89e7`), conflict resolved.

**Conflict & resolution:** master (via the merged speculator-abstraction PR #477) added two *new* functions — `decode_vl_dots_ocr_ngram` and `run_dots_ocr_ngram_loop` (the post-vision-prefill n-gram spec decode path) — at the exact insertion point where this PR adds `generate_dots_ocr_text`. git's 3-way merge interleaved the two independent function bodies (they share coincidental lines like `let t_gen = Instant::now()`). I reconstructed the region authoritatively: master's two n-gram functions kept verbatim, followed by this PR's `generate_dots_ocr_text` verbatim. The other two hunks (the arch_id=8 synthetic-token loop and the `generate()` dispatch branch) applied cleanly.

The net diff vs master is **byte-identical to the original fix** (223 insertions, daemon.rs only) — verified by comparing added lines. The PR is now a single clean commit; the stale tui/main.rs churn and whole-file-rustfmt `style:` commit (already-upstream noise) were dropped.

**Validation:**
- `cargo build --release --example daemon` clean.
- End-to-end run via coherence_probe on `dots-ocr.q8.hfq`: the new arch_id=8 text path loads, runs, streams 96 tokens, and terminates **without panic**. This is the path's core value — on master a text request to a dots.ocr model falls through to the llama dispatch and hits `unreachable!()` (panic).

**⚠️ Caveat (needs maintainer judgment, NOT a rebase regression):** on a plain-text prompt the output is low-diversity (coherence_probe: unique-token ratio 0.15, a 4-gram repeating 8×). This is **not** introduced by the rebase — the added diff is byte-identical to the original commit. Two non-rebase factors: (1) dots.ocr is a vision-OCR model, so a plain-text instruction with no image is off-domain; and (2) `generate_dots_ocr_text` encodes the prompt raw (`tokenizer.encode(prompt)`) with no chat-template framing — the same class of issue #466 fixes for `generate_qwen2`. Also note master's qwen2 forward path evolved since this PR's base (qwen2.rs +262, `forward_prefill_batch_embeds` + `embedding_lookup_dispatch` changed). Recommend validating the path on an actual OCR/image flow (or a framed text prompt) before merge.

fivetide:
✅ Validated in a local integration stack (`#497→#473→#495→#468→#466→#465`) on gfx1151 — `cargo build --release --workspace --locked` green; `scripts/coherence-gate.sh` correctness **PASS** (no hard errors) and `scripts/qwen2-reset-gate.sh` **PASS** (0 ceiling resets, 0 panics). Merges clean onto current master with no conflicts. (The pflash perf stage false-fails on this box: gfx1100 baseline vs gfx1151 — untouched path.)

Kaden-Schutt:
Folded into #534 as a cleanly rebased/cherry-picked delta on `beta` for v0.3.0, preserving the newer Redline and modular-runtime line. Closing this source PR in favor of the consolidated release PR. Thank you for the contribution.

PR #470, opened 2026-06-22
fix(daemon): rewind live qwen2 bundle state on reset (arch_id=7, #462 class)

fivetide:
## Problem

The qwen2 (`arch_id=7`) daemon **`reset` command silently no-ops**: `next_pos` is
not rewound between requests, so prior-turn KV bleeds into new turns until the
`max_seq` ceiling auto-reset fires. This is the same **#462 bundle-migration
class** that `serve-loop-gate`/`serve-multiturn-gate` were created for — but those
gates only load qwen3.5/DeltaNet, so plain qwen2 slipped through.

Found while generating VibeThinker-3B drafter SFT rollouts (each request sends
`{"type":"reset"}` then `generate`): the daemon logged
`context full (12387/16384) — resetting Qwen2State.next_pos` repeatedly, which is
structurally impossible if the per-request reset worked — a single 4096-token
request can never reach the 16384 ceiling from `next_pos=0`.

## Root cause

`Qwen2Carrier` puts the live `Qwen2State` into `ModelState::Qwen2(bundle)`, and
`generate_qwen2` reads it from there. But the daemon's reset / VL-checkpoint /
bench-prefill handlers rewind the **legacy `m.qwen2_state` direct field**, which is
`None` for plain qwen2 (the `skeleton(...)` default) and is only populated by
**dots-ocr** (`arch_id=8`). So `if let Some(s) = m.qwen2_state { s.reset() }` never
runs for plain qwen2.

The bleed is invisible to attractor detection: each request re-sends the full
ChatML scaffold, so the model stays anchored to the latest turn and outputs remain
on-topic — only the `next_pos` accumulation (and eventual ceiling reset) reveals it.

## Fix

- Add `LoadedModel::qwen2_mut()` (mirrors `minimax_mut()` / `lfm2moe_mut()`).
- Rewind `b.state` at all three reset sites (reset command, VL checkpoint,
  bench-prefill), **preserving** the existing `m.qwen2_state` branch so dots-ocr
  keeps working.

## Regression guard

`scripts/qwen2-reset-gate.sh`: loads a small qwen2 model at a small `max_seq`, sends
N explicit `reset`+`generate` cycles, and fails if the daemon logs a ceiling
auto-reset (impossible under a working per-request reset). Wired into the
`.githooks/pre-commit` serve-gate block alongside the qwen3.5-only
`serve-loop-gate.sh`, closing the documented qwen2 coverage gap.

## Validation

| check | pre-fix | post-fix |
|-------|---------|----------|
| `qwen2-reset-gate.sh` (16 reset cycles, qwen2.5-0.5B, `max_seq=256`) | **7 ceiling resets → FAIL** | **0 → PASS** |
| `serve-loop-gate.sh` (qwen3.5-0.8B, 10 multi-req) | — | **PASS, 10/10 ok** (no regression) |

The added blocks are inert for non-qwen2 arches (`qwen2_mut()` → `None`), so
qwen3.5 / llama / etc. are byte-identical.

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -187,6 +187,30 @@ if echo "$CHANGED" | grep -qE "$SERVE_HOTSPOT"; then
         exit 1
     fi
     echo
+
+    # qwen2 (arch_id=7) per-request reset regression (#462 bundle class).
+    # serve-loop-gate above is qwen3.5/DeltaNet-only; this exercises the plain
+    # qwen2 reset path that the Qwen2Carrier bundle migration broke — the reset
+    # handler rewound the dead `m.qwen2_state` field, not the live ModelState::
+    # Qwen2 bundle state, so next_pos bled across requests. Only a real reset
+    # no-op (exit 1) blocks; SKIP (3, no qwen2 model) / infra (2) warn-only.
+    if [ -r ./scripts/qwen2-reset-gate.sh ]; then
+        echo "=== qwen2 per-request reset gate ==="
+        bash ./scripts/qwen2-reset-gate.sh; q2rc=$?
+        if [ "$q2rc" -eq 1 ]; then
+            echo
+            echo "========================================================================"
+            echo "COMMIT BLOCKED: qwen2 per-request reset no-op — next_pos bled across requests."
+            echo "========================================================================"
+            echo
+            echo "The daemon reset handler must rewind the live ModelState::Qwen2 bundle"
+            echo "state via m.qwen2_mut(), not just the dots-ocr-only m.qwen2_state field."
+            echo "  ./scripts/qwen2-reset-gate.sh   # re-run"
+            echo
+            exit 1
+        fi
+        echo
+    fi
 fi
 
 echo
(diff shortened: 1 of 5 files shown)


fivetide:
**This PR is fully superseded by master — recommend closing.**

I checked whether it could be rebased onto current master (`d44f89e7`). Cherry-picking the core fix (`4a26538d`) onto master produces an **empty diff** (`nothing to commit, working tree clean`) — every change is already present upstream, landed via the merged speculator-abstraction PR (#477):

- `LoadedModel::qwen2_mut()` — present at `crates/hipfire-loader/src/lib.rs:443`
- the three reset sites already rewind `m.qwen2_mut()` → `b.state.reset()` (daemon.rs ~1943 / ~2160 / ~2385)
- `scripts/qwen2-reset-gate.sh` — present on master
- `.githooks/pre-commit` serve-gate wiring — present on master

Master has in fact gone **further** than this branch: at the reset site, master applies the same #462-class bundle fix to deepseek4 (`m.deepseek4_mut()`), whereas this branch still has the older `m.deepseek4_state` direct field. So rebasing would only *revert* progress — there is nothing left to port.

No force-push made (the rebased branch would be identical to master). Suggest closing as already-merged. The two trailing `chore: rustfmt/clippy` commits were debt-clearing only and are also moot.

fivetide:
Superseded by master: rebasing `fix/qwen2-reset-noop-bundle` onto current `master` (d44f89e7) yields an empty diff. Master already contains the qwen2 `qwen2_mut()` bundle reset at all three reset sites plus `scripts/qwen2-reset-gate.sh`, and applied the equivalent #462-class fix to deepseek4 via `deepseek4_mut()`. The two remaining commits on this branch were pure rustfmt fmt-debt reformats. Closing as absorbed.

PR #482, opened 2026-06-26
feat(dispatch): QKV-bias fold across all per-row fused-QKV dtypes (default-on, byte-identical, +2.5–4.1% decode)

fivetide:
## QKV-bias fold: eliminate qwen2 per-layer `bias_add` launches (default-on, byte-identical, +2.5–4.1% decode)

Productizes the QKV-bias fold across **all per-row fused-QKV decode dtypes**. qwen2-family models (`attention_bias=true`) run 3 tiny, overhead-dominated `bias_add` kernels per layer right after the fused QKV projection. This folds the Q/K/V bias into the fused-QKV kernel's lane-0 store (`y[row] = acc + (bias ? bias[row] : 0.0f)`), so 3 separate launches/layer become 0 — a free fma at a store that already runs (no grid change, no barrier, no occupancy loss).

NULL bias pointers are an exact kernel no-op, so every existing (non-bias) caller stays byte-identical regardless.

### Mechanism
- **Flag** `fuse_qkv_bias` — env `HIPFIRE_FUSE_QKV_BIAS`, **default ON**, opt out with `=0` for bisection (resolved once at init; no per-launch env lock).
- **Kernels** gain 3 nullable `bias_q/k/v` params + the `+bias` epilogue (base + every arch sibling, so the kernarg ABI matches whatever the host dispatches).
- **Dispatch**: `FusedQkvParams.bias: Option<[&GpuTensor;3]>`; each per-row decode arm routes through a `gpu.fused_qkv_*_with_bias` variant (old methods delegate with nulls).
- **`execute_steps`**: a generalized peek-ahead — when a fold-supported 3-way QKV decode key is followed by 3 `BiasAdd` on the q/k/v outputs (ptr-identity checked), fold + skip them. dp4a archs (gfx906, no bias params in their kernel) are guarded off so the 3 bias adds run separately as before.

### Coverage
- **Family A** (clean byte-identical fold into existing per-row kernels): `fused_qkv_hfq4g256` (MQ4G256 + HFQ4G256), `mq4g256_lloyd`, `mq3g256_lloyd`, `q4k`, `q8_0` (+ gfx1100/wave64/v2 siblings).
- **Family B** (authored a new per-row kernel): `fused_qkv_hfq6g256` for HFQ6/MQ6, which previously had no per-row decode kernel (ran the n=1 GEMM). The GEMM→per-row decode swap is gated to only when the fold actually fires (bias present); fold-off keeps the historical GEMM.
- Correctly **skips** QKVZA/DeltaNet and qwen3/qwen35/MoE (no QKV bias), and research/unused formats (would be dead code).

### Validation (gfx1151)
**Correctness — bit-exact:**
- MQ4: three-way byte-identity — clean-master ≡ fold-off ≡ fold-on (fold fired 3060×).
- New GPU example `crates/rdna-compute/examples/test_fused_qkv_bias_parity.rs` (ALL bit-exact, max_abs=0): for every dtype `no-bias==0` and `with-bias==bias` (validates kernarg ABI + q/k/v→bias routing + epilogue); plus `fused_qkv_hfq6g256 == 3×gemv_hfq6g256` on finite weights (Family B's required gate).

**Coherent-model byte-identity** (Qwen2.5-1.5B-Instruct, dim 1536, a real coherent qwen2-bias model):

| dtype | output | distinct | fold-on ≡ fold-off |
|---|---|---|---|
| mq4-lloyd | coherent prose | 74 | ✅ byte-identical |
| mq3-lloyd | coherent prose | 59 | ✅ byte-identical |

**No regression:** `scripts/coherence-gate.sh` exit 0 across all present qwen3.5 models (qwen35 has no QKV bias → fold never fires there; confirms the dispatch additions don't regress).

### Performance (decode tok/s, fold-on vs fold-off; warm + 5 fresh procs each; byte-identical prompts; non-overlapping ranges = real signal)

| model / dtype | fold-off | fold-on | Δ |
|---|---|---|---|
| vibethinker-3b.mq4 (HFQ4/MQ4, 36L) | 91.70 | 94.03 | **+2.54%** |
| qwen2.5-1.5b.mq4-lloyd (28L) | 132.80 | 136.99 | **+3.16%** |
| qwen2.5-1.5b.mq3-lloyd (28L) | 148.59 | 154.68 | **+4.10%** |

The win scales with how launch-bound the decode is (smaller/faster models recover proportionally more launch tax) — matching the PoC's +2.6% and the thesis that the recoverable tax on gfx1151 dense decode is in the *tiny* ops.

### Notes
- q4k / q8_0 / hfq6 are validated by the GPU parity example; no dense qwen2 model in those formats was available for a model-level coherence run (the parity test is the hardware validation there).
- The original PoC test model (vibethinker-3b) produces degenerate greedy output; this was investigated and found to be **unrelated to this change** (identical on clean-master; a near-lossless Q8 re-quant also reproduces it, ruling out quantization) — a separate model/engine-interaction issue, tracked independently.

### DIFF
--- a/crates/hipfire-arch-qwen35/src/qwen35.rs
+++ b/crates/hipfire-arch-qwen35/src/qwen35.rs
@@ -7204,6 +7204,7 @@ fn run_fused_gate_up_key(
         k,
         rot_scratch: &[],
         batch_size: Some(n),
+        bias: None,
     };
     hipfire_runtime::llama::fused_qkv_family()
         .run(&ctx, gpu, &params)
@@ -7253,6 +7254,7 @@ fn run_fused_qkv_key(
         k,
         rot_scratch: &[],
         batch_size: Some(n),
+        bias: None,
     };
     hipfire_runtime::llama::fused_qkv_family()
         .run(&ctx, gpu, &params)
@@ -7300,6 +7302,7 @@ fn run_fused_qkvza_key(
         k,
         rot_scratch: &[],
         batch_size: Some(n),
+        bias: None,
     };
     hipfire_runtime::llama::fused_qkv_family()
         .run(&ctx, gpu, &params)
--- a/crates/hipfire-dispatch/src/families/fused_qkv.rs
+++ b/crates/hipfire-dispatch/src/families/fused_qkv.rs
@@ -20,6 +20,13 @@ pub struct FusedQkvParams<'a> {
     /// for `x_rot_up` internally). Empty slice for non-Paro keys; existing arms
     /// ignore it.
     pub rot_scratch: &'a [GpuTensor],
+    /// Optional Q/K/V bias tensors for `HIPFIRE_FUSE_QKV_BIAS`. When `Some`, the
+    /// bias is folded into the per-row fused-QKV decode kernel's lane-0 store
+    /// (eliminating 3 separate `bias_add` launches). Only the 3-way QKV decode
+    /// arms whose kernel has a `_with_bias` variant read this field; all other
+    /// arms (QKVZA, gate+up, Paro, prefill GEMM) ignore it. `None` (the default)
+    /// means no fold — existing callers pass `None` and stay byte-identical.
+    pub bias: Option<[&'a GpuTensor; 3]>,
     /// Batched-prefill row count (`#397 Ship 5.2 slice 2`). `None` = single-token
     /// DECODE: gate+up arms dispatch to the `gpu.fused_gate_up_*` kernels (the
     /// historical behavior; the decode pipeline in `pipeline::steps` passes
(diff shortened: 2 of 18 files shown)


fivetide:
Rebased onto current master (`d44f89e7`), conflict resolved.

**Conflict was the redundant tui commit.** The only collision was `crates/hipfire-tui/src/main.rs` — from the trailing `fix(tui): resolve clippy::never_loop` commit, which master already has (the `if let Some(arg) = args().nth(1)` form is on master at main.rs:30). The 3 real QKV-bias-fold commits touch none of the conflicting files, so I cherry-picked them onto fresh master (all applied **cleanly**) and dropped the redundant tui commit. Net diff: 18 files / 1113 insertions — exactly the fold work (dispatch tables, the fused_qkv HIP kernels, gemm/gemv, feature_flags, and the parity example).

**Validation:**
- `crates/rdna-compute/examples/test_fused_qkv_bias_parity` (GPU, gfx1151): **ALL PASS** — every `fused_qkv_*` folds bias bit-exactly (`no-bias max_abs=0.000e0`, `with-bias max_abs=0.000e0`) across mq4g256-lloyd, mq3g256-lloyd, q4k, q8_0, and the new Family-B hfq6g256; plus `hfq6 fused == 3×gemv_hfq6g256` bit-exact.
- Full workspace build (lib + bins + examples + kernels): clean.
- End-to-end coherence: `qwen25-0.5b-instruct.mq4` (qwen2 arch_id=7, exercises the mq4 fold) with the fold default-ON generates a clean, coherent haiku (coherence_probe verdict OK, 0/0). A coherent fold-ON output rules out a double-bias dispatch bug — double-applying bias would corrupt Q/K/V into garbage, not produce coherent text.

**Scope note:** a strict byte-identical fold-on-vs-fold-off token A/B wasn't cleanly capturable through the probe tooling on this box (the committed-token-id stream isn't emitted on the qwen2 .mq4 path, and the one quantized qwen2 model that would best exercise hfq6 — vibethinker .mq6 — degenerates to 0 tokens at greedy, a pre-existing issue). Per the original PR's methodology, correctness for this fold rests on the bit-exact GPU parity suite (which passes here), backed up by the coherent default-ON run. Default-off escape hatch `HIPFIRE_FUSE_QKV_BIAS=0` preserved.

Now `MERGEABLE`.

Kaden-Schutt:
Folded into #534 as a cleanly rebased/cherry-picked delta on `beta` for v0.3.0, preserving the newer Redline and modular-runtime line. Closing this source PR in favor of the consolidated release PR. Thank you for the contribution.

PR #483, opened 2026-06-27
feat(spec): qwen35 DDTree distribution-correct temp>0 spec decode (SWOR) + target-generic DFlash for LLaMA/Qwen3

fivetide:
## ⚠️ Update (2026-06-30) — dense-DFlash seed-anchor bug fixed; DDTree gated behind `--ddtree`

> These two changes land on top of everything below and **supersede the "tree-SWOR is the default" and "z-lab Qwen3-8B drafter is τ≈0.2 / DFlash is acceptance-bound" framing** in the original description. Verified via a torch-reference diff against z-lab's own `spec_generate`.

### 1. Root-caused + fixed a real dense-DFlash bug — the Qwen3-8B drafter was never weak (`961a94ef`)

`GenericDflashSpeculator` built its noise block by broadcasting `embed(MASK)` to **all** `b` slots. The z-lab reference (`dflash.py:235-237`, `embed_tokens([seed, MASK, …])`) puts **`embed(seed)` — the committed token — at slot 0**, the anchor every drafted slot attends to in the bidirectional block. Without it the drafts collapse into degenerate repetition and acceptance craters.

Localized by diffing against z-lab's torch `spec_generate` on **bf16 Qwen3-8B** (reference **τ=6.34** on the LRU code prompt; hipfire was 1.46): hipfire's captured `target_hidden` per-layer RMS matched the reference *exactly* (`[0.27, 22.8, 23.3, 23.5, 26.7]` vs `[0.27, 23.1, 23.5, 23.8, 26.9]`), but drafts were degenerate given identical conditioning → the bug was in the **draft input**, not the capture.

The old **"τ≈0.2"** was two stacked artifacts: **(a)** the bench ran the daemon with thinking **on** (Qwen3-8B is a thinking model), so the whole window was `<think>` reasoning the drafter never trained on — out-of-distribution; `/no_think` alone recovered τ 0.27→1.46. **(b)** this seed-anchor bug.

**Impact (qwen3-8b mq4, in-distribution, temp 0, same prompt):**

| | chain tok/s | τ |
|---|---|---|
| before | 20.9 | 1.46 |
| **after** | **35.3 (+69%)** | **3.27 (+124%)** |

Greedy-lossless (verify untouched; output token-identical to AR; code/prose/fact coherent). Dense-path only — the 27B qwen35 DeltaNet DFlash builds `[seed, drafts.., bonus]` per-slot and never had the bug (gate green).

[...]

PR #484, opened 2026-06-29
feat(dspark): DeepSeek-V4 DSpark draft-module speculator for deepseek4 (code beats MTP, stacked on #483)

fivetide:
## What this is

Ports DeepSeek's **DSpark** draft module (from `deepseek-ai/DeepSeek-V4-Flash-DSpark`) into hipfire's `hipfire-arch-deepseek4` as a drop-in alternative to the in-trunk MTP speculator. DSpark is a **3-stage MTP chain** (`mtp.0/1/2`) that drafts a 5-token block per window from a multi-layer target-context (`main_hidden` mean-pooled over the HC residual at target layers `[40,41,42]`), a vanilla rank-256 markov head, and a confidence head — *not* dense DFlash. It rides the existing unified `MtpSpeculator` seam: the only thing that changes is the draft **source** (one block-batched `dspark_forward` instead of K iterations of `mtp_forward`); verify + greedy-accept reuse the shared trunk-forward + `accept_greedy_prefix` core.

The line-accurate blueprint extracted from the reference `inference/model.py` is committed at `docs/design/2026-06-28-dspark-deepseek4-forward-spec.md`.

## Builds on

**Stacked directly on #483** (`feat(spec): qwen35 DDTree distribution-correct temp>0 spec decode (SWOR) + target-generic DFlash`). This branch's merge-base with #483 is `d7dffb1f`, which is exactly #483's head — i.e. #483 is a clean ancestor with zero divergence. The DSpark work is the **17 commits on top** of #483.

Because #483 is not yet merged, the diff against `master` shows all 52 commits (35 from #483 + 17 DSpark). **Reviewers should either merge #483 first, or review only `d7dffb1f..HEAD`.** The DSpark-specific change set is:

```
17 files changed, 3392 insertions(+), 56 deletions(-)
```

| Area | File |
|---|---|
| Loader / config | `carriers.rs`, `arch.rs` (DsparkConfig parse, sidecar wiring) |
| Forward | `forward.rs` (3-stage `dspark_forward`), `deepseek4.rs`, `dspark_speculator.rs` |
| GPU kernel | `kernels/src/dspark_stage_kv.hip`, `rdna-compute/{attention,kernels}.rs` |
| Quantize | `scripts/quantize-dspark.sh`, `hfq_dump.rs` |
| Examples | `dspark_{bench,forward_smoke,load_smoke,parity}.rs` |

## Correctness

- **Numeric parity (novel head surface):** `dspark_parity` validates the GPU kernels with **no trunk reuse** — `main_proj` gemv, `main_norm` RMS, `markov_w1` embed, `markov_w2` bias — each against an independent CPU reference derived from the reference `model.py` on the real quantized weights. **All PASS at cosine 1.000000** (markov embed bit-exact). This is the runnable slice of the mandated numeric-parity spike (the full fp8 `model.py` can't run on RDNA) and closes the review's biggest exposure — correctness no longer rests on coherence alone.
- **On-GPU KV staging:** the per-window attention KV assembly was moved fully on-GPU (`dspark_stage_kv` kernel); `HIPFIRE_DSPARK_VERIFY_STAGE=1` asserts byte-identity to the old host path — 0 mismatches / 340480 elems across all 3 stages.
- **Coherence:** PASS across capital / haiku / physics / 200-tok binary-search prompts on gfx1151. 200-tok run: first128 unique-ratio 0.55 / last128 0.66 / 3gram density 0.00 — no attractor or structural loop. A 5-agent fidelity review against the reference `model.py` rated the port faithful.

## Performance (gfx1151 / Strix Halo, MQ2-Lloyd V4-Flash sidecar)

Fresh-process, post-warm, byte-identical prompts (token-md5 recorded), `dspark_bench` driving the exact generic `Speculator` path the daemon uses. The session's job was to rule out the wrong levers and find the right one — the **confidence head was the lever**:

| Lever | Verdict |
|---|---|
| Warm stage KV rings in prefill | **Loss** — measured, documented, not shipped (the trained draft attends better to a sparse recent window under mq2; τ 3.24 → 2.55) |
| Fold bootstrap into verify | **Loss** — reverted (dropping the free bonus token costs ≈1.4× more verify forwards than the bootstrap saves) |
| Cold-ring zero-key masking | **Loss** — −6–8% tok/s (under 2-bit quant the zero-key denominator inflation is a net acceptance *benefit*, not a bug) |
| GPU staging kernel (drop per-window host d2h/h2d) | **Shipped** — bit-identical, +4% prose |
| On-GPU markov argmax | **Shipped** — token-identical, perf-neutral; removes ~5 MB/window dead host traffic |
| **Confidence-truncated verify (T0.5)** | **The win** |

**Confidence truncation** (`5f34a214`) is DSpark's own mechanism — computed by the draft but previously unused. It truncates the draft block to its confident prefix (first slot where `sigmoid(confidence) < threshold`) *before* the heavy 43-layer / 256-expert trunk verify runs. Coherence-safe by construction: committed tokens stay greedy-verified; it only skips verifying slots the draft is unsure of (likely-rejected anyway). Default `0.5`, env-overridable; detector battery clean throughout.

Reproducible threshold sweep vs in-trunk MTP (tok/s; default **T0.5**):

| Prompt | MTP | DSpark off (T0.0) | **DSpark T0.5** | T0.7 |
|---|---|---|---|---|
| prose | 14.0 | 10.7 | **12.9** (τ 2.96) | 11.4 |
| code  | 15.1 | 14.0 | **15.6** (τ 3.64) | 14.2 |

- **T0.5 makes code BEAT MTP** (+3% tok/s, τ 3.64 vs 2.50) and lifts prose +20% (closing the MTP gap from −24% to −8%).

**Honest verdict:** DSpark **wins on code/structured generation**, is **within ~8% on prose**, and has higher acceptance/τ everywhere. It does *not* dominate MTP on prose — and that remaining gap is structural, not a tuning miss: DSpark verifies `block+1 = 6` positions per window vs MTP's `K+1 = 3`, so its heavier per-window machinery only pays back when τ is high enough (code). Fully closing prose would need a smaller block or a cheaper verify, with diminishing returns. Shipped as the faster choice for code/structured generation, coherence-parity elsewhere.

## Configuration surface

DSpark is configured entirely through the **unified `speculation` selector** and one tuning knob — there are no DSpark-specific env gates. Everything resolves through the standard config ladder (**env > CLI flag > per-model config > global config**), the same as the other speculation mechanisms.

### 1. Mechanism selection — `speculation`

`speculation` is the single canonical selector; exactly one drafter ever runs. DSpark joins `off|auto|ngram|dflash|mtp` as a sixth value. For a deepseek4 target with a `-dspark` sidecar next to it:

| `speculation` | DSpark sidecar loaded? | Drafter that runs |
|---|---|---|
| `auto` *(default)* | yes, if the sidecar file exists | **DSpark** (auto-wins over in-trunk MTP — preserves the prior default-on) |
| `dspark` | yes (forced; warns + falls back to AR if absent) | **DSpark** |
| `mtp` | **no** — load skipped | in-trunk MTP head |
| `dflash` / `ngram` / `off` | **no** — load skipped | dflash / n-gram / none (AR) |

Set it three ways (highest priority first):

```bash
HIPFIRE_SPECULATION=dspark hipfire run <model>        # env (top of ladder)
hipfire run <model> --spec dspark                     # CLI flag (also --speculation)
hipfire config <model> set speculation dspark         # per-model config
hipfire config set speculation dspark                 # global config
```

**The selector now also gates the *load*.** Previously the 3×MoE DSpark sidecar was paged into VRAM whenever the file existed; now `--spec mtp/dflash/ngram/off` skips loading it entirely. This is what makes the old `HIPFIRE_DEEPSEEK4_LOAD_DSPARK` knob unnecessary — load and use are decided by one mode.

### 2. Tuning — `dspark_conf_threshold` (confidence-truncated verify)

The single DSpark tuning knob is the confidence-truncation cutoff (the headline perf lever above). Default `0.5`; range `[0, 1]`. Only consulted when DSpark runs.

```bash
HIPFIRE_DEEPSEEK4_DSPARK_CONF_THRESHOLD=0.6 hipfire run <model>   # env
hipfire run <model> --dspark-conf-threshold 0.6                  # CLI flag
hipfire config <model> set dspark_conf_threshold 0.6            # per-model
hipfire config set dspark_conf_threshold 0.6                    # global
```

Resolved **once at speculator build** (env > flag/param > `0.5`); the per-step env read of the original implementation is gone. It also appears in the `hipfire config` TUI with a `[0,1]` slider.

### Migration from the prototype env gates

The three env vars the feature shipped with are replaced — no user action needed for the default path (`auto` reproduces the old default-on behavior):

| Removed env (prototype) | Now |
|---|---|
| `HIPFIRE_DEEPSEEK4_DSPARK=0` (force MTP) | `--spec mtp` |
| `HIPFIRE_DEEPSEEK4_LOAD_DSPARK=0` (skip load) | folded into the mode (any non-dspark `--spec` skips the load) |
| `HIPFIRE_DEEPSEEK4_DSPARK_CONF_THRESHOLD` | kept as the top-of-ladder override for `--dspark-conf-threshold` |

> The standalone `dspark_bench` example keeps its own `HIPFIRE_DEEPSEEK4_DSPARK` A/B switch — it builds the speculator directly and never goes through the loader, so it is a bench-local harness knob, not a production gate.

### Validation

On gfx1151 (deepseek-v4-flash mq2lloyd) through the real CLI→daemon path:
- `--spec dspark` → sidecar loaded, *"DSpark speculator enabled (block=5)"*, coherent.
- `--spec mtp` → **no** "opening DSpark sidecar" line (6 GB sidecar skipped, VRAM saved), *"MTP speculator enabled (in-weights, K=3)"*, coherent.
- `--dspark-conf-threshold` rejects out-of-range / non-numeric before any load.
- CLI `tsc` clean + 281 bun tests pass (the config-meta completeness gate is green, so the new `dspark_conf_threshold` key is fully wired: default + validation + TUI meta).


nwoolmer (commented):
## Code review — DSpark drafter for deepseek4

Reviewed the incremental slice `d7dffb1f..HEAD` (the 17 DSpark commits on top of #483), across `forward.rs`, `dspark_speculator.rs`, `arch.rs`, `deepseek4.rs`, `carriers.rs`, the `dspark_stage_kv.hip` kernel, and the CLI/loader wiring.

**Verdict: mergeable. No CRITICAL or HIGH correctness bugs.**

The central safety property holds: **committed-token correctness does not depend on any of the novel DSpark math.** The shared `accept_greedy_prefix` over a *greedy* trunk verify guarantees the emitted stream equals greedy AR regardless of whether the draft forward, mean-pool capture, KV staging, or confidence head are numerically perfect. The worst failure mode of a bug in the novel path is *low acceptance / slow*, never *wrong output*. That's the right architecture for a speculator, and it's why nothing here rises to HIGH/CRITICAL.

### Verified correct
- **Confidence-truncated verify** (the headline feature) is correct — truncation shortens the *draft proposal* before `accept_greedy_prefix`, so it can only reduce eligible slots, never inject an unverified token. The committed stream is identical to the untruncated block. (`dspark_speculator.rs:242-320`)
- **`dspark_stage_kv.hip`** is clean — one thread per `(d,b)`, disjoint writes, no races, `size_t` offsets, and the `HIPFIRE_DSPARK_VERIFY_STAGE=1` host cross-check is bit-exact. Index math (block positions `P+1..=P+block`, the `n_valid ≤ 1024` guard) checks out.
- **Load-gating selector** traced end-to-end (CLI → daemon → carriers): `off/dflash/mtp/ngram` skip the sidecar; `auto/dspark` load it; DSpark wins over in-trunk MTP when both are present. No path drives the wrong drafter.

### Follow-ups (non-blocking)
- **MEDIUM** — `dspark_conf_threshold` has **no `.clamp(0, 1)` at the Rust consumption point** (`build_deepseek4_dspark_speculator`). The TS validates it, but the `HIPFIRE_DEEPSEEK4_DSPARK_CONF_THRESHOLD` env var and the daemon JSON path bypass that validation. Out-of-range isn't *unsafe* (output stays greedy-correct — `>1` truncates to len 1 ≈ AR, `<0` keeps the full block) but it silently degrades. A one-line clamp in the builder turns a silent-degradation mode into an enforced contract.
- **MEDIUM** — no validation that `target_layer_ids < num_hidden_layers`. An out-of-range target layer never matches in the capture check, leaving `main_hidden` stale/zero → silent draft-quality collapse (acceptance drops, output still correct). A load-time bounds check would make this loud instead of silent.
- **LOW** — per-call scratch tensors in `dspark_forward` (`staged`, `main_kv`, markov/head buffers) leak on early `?` error returns (freed with `let _ = gpu.free_tensor(...)` only on the success path, no RAII guard). Only bites error paths.
- **LOW** — the quantize script writes `...-dspark.mq2lloyd`, but sidecar discovery expects `<trunk_stem>-dspark.<trunk_ext>` next to the trunk; the operator must place/rename it (ops, not code).
- **LOW** — stale doc comment at `spec.rs:876` claims DSpark samples its verify; DSpark verify is greedy (`requires_greedy() == true`).

### Note on verification limits
I can't run the GPU here, so the on-device numerics (draft forward, mean-pool capture, confidence head) rest on the `dspark_parity` harness (cosine ≥ 0.9999 vs a CPU reference) and the coherence gate — both present in the diff. The `dspark_parity` addition is the right call given the fp8 `model.py` reference can't run on RDNA; it isolates the novel linear heads where a transpose/dequant bug would otherwise hide.

---
🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

PR #492, opened 2026-07-03
feat(dspark): qwen3 DSpark speculative decode + GPU-resident perf + τ-adaptive block controller

fivetide:
**Base:** `master` (stacked on #484 — the DeepSeek-V4 DSpark port). This PR is the work on
`feature/dspark-qwen3` since it diverged from #484 (`4e7d3143..HEAD`, ~147 commits).

## TL;DR

Generalizes DSpark speculative decode into a reusable, arch-agnostic engine, ports it to
**qwen3-dense**, makes the drafter substantially faster (GPU-resident decode path,
accepted-prefix hidden reuse, batched lm-head, lazy prefix-stop), enables and wins
**temperature > 0** serving via distribution-preserving sampled verify, and adds a
**default-on τ-adaptive draft-block-size controller** that auto-picks the block per arch.
Everything is validated with per-arch coherence gates + byte-identical parity harnesses;
negative results are committed, not hidden.

## Headline results (gfx1151, warm; committed byte-identical prompts)

| Change | Measured |
|---|---|
| Accepted-prefix hidden reuse (qwen3) | **+31.7% prose / +25.3% code** tok/s (bootstrap share 10.4%→0.4%) |
| Accepted-prefix hidden reuse (deepseek4) | **+67%** tok/s (9.9→16.6; bootstrap 36.9%→2.8%) |
| On-GPU verify argmax (deepseek4) | **+37–40% code** (5.3→7.3–7.9), byte-identical τ/accept |
| Batched F16 lm-head GEMM | `run_heads` −31%, **+10%** qwen3 code & prose |
| Lazy prefix-stop, greedy verify | **+16% qwen3 code / +46% ds4 code**, byte-identical output |
| Lazy prefix-stop, temp>0 verify (qwen3) | code 20→**29.6 tok/s**, **+24% vs AR**, now served |
| τ-adaptive controller (deepseek4) | code **19.6 tok/s @ block 2** (+46% vs fixed-b5, +60% vs AR) |
| τ-adaptive controller (qwen3) | climbs block 2→3 (+40% over stuck-block-2) |

## 1. DSpark engine generalization + qwen3 port

- **Arch-agnostic `dspark-core`.** The deepseek4-specific drafter was refactored into a
  `DsparkBody` trait + generic `DsparkDrafter` (`crates/hipfire-runtime/src/dspark_core.rs`,
  `2c624f29`/`515b9867`). The core owns the whole window loop (bootstrap → `draft_block` →
  `run_heads` → `verify_block` → `accept_greedy_prefix`); arches implement only the
  attention body and the `SpecTarget` verify/commit/capture seams. **deepseek4 migrated onto
  the shared core byte-identical** (`f63798fc`: τ/accept/windows unchanged, parity cosine
  1.000000, coherence 6/6).
- **qwen3-dense port — quantize → load → forward → dispatch.** `--format qwen3-dspark-q8`
  emits a 5-layer dense drafter sidecar (`f4abbaf4`), discovered at `<stem>-dspark.<ext>`
  and loaded via `load_qwen3_dspark` (`3c16c169`). `dspark_qwen3_block_forward` implements
  the bidirectional `[ctx++block]` GQA masked attention faithful to the reference
  (`54af5391`), wired into `LlamaCarrier` with DSpark > DFlash > n-gram precedence
  (`cfaac0d8`). Two correctness bugs were caught by the parity harness *before* any perf
  measurement: a head-path double-norm (`de28aef9`) and an arch-specific confidence-norm
  (`845d515b`). Parity on gfx1151: x_head cosine 0.9997, markov token-identical, confidence
  cosine 0.9999. E2E coherence gate `scripts/coherence-gate-qwen3-dspark.sh` + a tok/s bench
  added (`b1f39f6f`).

## 2. Performance — GPU-resident decode path

Five passes (P1–P5) move per-window data off the host round-trip. On gfx1151 UMA (coherent
memory) several are neutral; they are real bandwidth wins on discrete RDNA (gfx1100/1201):

- **On-GPU attention KV assembly** (`8388d5b3`) — new `dspark_stage_kv` kernel, −6 stream
  syncs/window.
- **On-GPU verify argmax** (`d63bd790`/`b1adf411`) — download token ids, not the
  `[block+1 × 200k]` logit tensor. **+37–40% deepseek4 code**, byte-identical.
- **GPU-resident confidence head** (`3fa96a30`), **markov head chain** (`d2d640cc`, P4), and
  **accepted-prefix hidden reuse** (`b6f1db96`, P5, new `verify_block_capture_gpu` seam) —
  byte-identical, dGPU-portability wins.
- **Accepted-prefix hidden reuse as multi-slot context** (`3ccaaa8c`/`a5350141` qwen3;
  `3a89aab6` deepseek4) — bootstrap runs once per generation, not per window: **+31.7%
  prose / +25.3% code** (qwen3), **+67%** (deepseek4).
- **Batched F16 lm-head GEMM** (`7eb2acec`) — the drafter lm-head was ~21% of the window;
  a K2-pipelined MW16 WMMA kernel that preserves the trunk F16 cache cuts `run_heads` −31%,
  **+10%** qwen3 both genres. (The naive batched variant was tried first and regressed
  verify −41% by evicting the trunk cache — reverted.)
- **Lazy prefix-stop, greedy verify** (`0709898e`) — acceptance is a strict prefix, so stop
  the per-position lm-head at the first mismatch (`lazy` flag; only DSpark capture paths take
  it, DFlash/n-gram stay eager). **+16% qwen3 code / +46% ds4 code**, byte-identical.

## 3. temp > 0 sampled verify + serving

- **Distribution-preserving sampled verify** (`bc4df7c2`): at temp>0 the target draws
  `t_i ~ p_T(temp, top_p, top_k)` per position and the point-mass draft is accepted iff it
  equals that sample — exact temp-T speculation, committed token is always the target's own
  sample. `temp ≤ 1e-6` stays on the byte-identical greedy path.
- **GPU-fused sampler + lazy prefix-stop turned it into a win** (`adb90438` → `0f007882`):
  reuses the same `sample_top_p_pf` kernel AR uses (so committed tokens are distribution-
  identical to AR), and lazily stops at the first mismatch so the 152k-vocab lm-head GEMV
  runs ~τ times/window instead of `block` times. qwen3 code temp=0.7 **20.3 → 29.6 tok/s**
  (+24% vs AR 23.8). A **fused on-GPU sample+accept kernel** (`2f744c11`) folds the host
  sampler loop into one launch (parity 60/60; UMA-neutral, dGPU win).
- **qwen3 temp > 0 enabled in serving** (`af6e7ff5`): the daemon chain-routes a sampling-
  capable non-SWOR drafter to the spec loop instead of AR. deepseek4 temp>0 is **gated off
  in serving** (`3fe37f27`, bench-only) because prose consistently loses (−14%) — it awaits
  the fallback in §4.
- **conf-threshold tuning + CLI-shadow fix** (`7aa7a77e`/`066d084a`/`a85a072d`): qwen3
  default 0.5→**0.1** (+16.6% prose); a CLI bug that unconditionally forwarded 0.5 and
  clobbered it (−15% code/−33% prose) fixed by making the CLI default `null`; deepseek4
  default 0.5→**0.3** (lazy verify makes a higher draft budget profitable: code +5.8%,
  prose +20%). **DSpark promoted to a first-class `--spec dspark` mode** with
  `--dspark-conf-threshold` (env > CLI > per-arch carrier default) and VRAM-saving
  conditional sidecar load (`813bfe7a`).

## 4. τ-adaptive draft-block-size controller (default-on)

`BlockController` (`crates/hipfire-runtime/src/dspark_block_controller.rs`, no-GPU,
unit-tested) picks the draft block per window by a throughput cost model:

> `block = argmax_N  τ(N) / (t_ar + N·Δt)`,  where `τ(N) = 1 + Σ_{k≤N} S(k)`,
> `S(k) = P(accept_len ≥ k)` is the acceptance-depth survival, and `t_ar`/`Δt` are the
> live-calibrated single-window / marginal-per-position cost.

- **Auto-adapts, no per-arch knobs.** DeepSeek4 (expensive MoE verify, shallow survival)
  settles at **block 2**; qwen3 (cheap dense verify, deeper survival) climbs. The controller
  reads each arch's own cost curve.
- **Ramp-bootstrap breaks a calibration deadlock** (`645ba717`) — a per-request sweep
  `min→max` seeds ≥2 distinct block timings + the survival counts, then the argmax takes
  over. Cost is fit on **full per-window wall-time** (draft+heads+verify), not verify-only
  (which mis-charges large blocks); survival is count-based so deep entries are *held*, not
  decayed — fixing a cap-trap (`4e6c8138`). `p* = Δt/t_ar` is a same-session, thermal-
  invariant ratio, preserved across `reset()`.
- **Default-on**, `HIPFIRE_DSPARK_ADAPTIVE_BLOCK=0` restores fixed block. Shipped result:
  deepseek4 code **19.6 tok/s** (beats fixed-b5 13.4 and AR 12.2); qwen3 climbs off the
  stuck block-2 (+40%). **8 no-GPU unit tests** (`cargo test -p hipfire-runtime
  dspark_block_controller`).

### spec↔AR fallback — explored, reverted

A "block 0 = one plain target forward when no spec block beats AR" fallback was implemented
(`57658693`) and **reverted** (`bc84eb51`). It delivered a large qwen3 temp>0 win (falls to
AR: prose 20→37, code 30→37 tok/s) and the AR step was proven state-clean (deepseek4 block-2
acceptance unchanged: 0.506 vs 0.530). But it cost deepseek4 code **~7% (warm)** — the ramp's
mandatory AR-cost-measurement windows commit only 1 token/window — and deepseek4 is the
flagship spec arch. It is reverted in favor of block-modulation-only, and preserved in
history for a future **arch-gated** revisit or a zero-DS4-cost AR-cost estimator that avoids
the ramp AR window.

## Validation & test plan

- **Coherence gates** per arch: `coherence-gate-deepseek4-mtp` (6/6, --full), the new
  `coherence-gate-qwen3-dspark.sh` (4 rows, Tier-1/2 attractor thresholds), greedy coherence
  gate for served qwen3 temp>0.
- **Parity harnesses**: cosine 1.000000 for byte-identical changes; CPU Python reference
  (`qwen3_dspark_cpu_ref.py`) for the qwen3 forward (ctx_len 1 + 3); deepseek4 τ/accept/
  windows regression sentinel (τ=1.818/accept=0.178/windows=110 @ block 5 on lru-code).
- **Unit tests**: 8 no-GPU controller tests; workspace `cargo build --release --workspace
  --locked` + `cargo test --lib` green (358+ lib tests at the temp>0 milestone).
- **Negative results committed** (not shipped): naive batched lm-head (verify −41%),
  ring-warm / bootstrap-fold levers (measured losses), ds4 greedy-lazy is ~neutral (MoE
  trunk dominates; head <1% of window) — see `.agent-memory/notes/dspark-*.md`.

## Notes

- Stacked on #484; if #484 merges first, rebase drops its commits from this diff.
- All new serving behavior is default-on with `=0`/`--spec` opt-outs; no `HSA_OVERRIDE`.

---

## Update — deepseek4 temp>0 DSpark **un-gated** + CACTUS (follow-up in this branch)

**Review response (#492's one MEDIUM):** the CACTUS GPU rejection path sampled the
raw target instead of the h-distribution the host builds. Fixed in-kernel
(`chain_accept_spec.hip`), byte-identical at δ=0; validated on gfx1151 (c8
60/60 at δ=0; δ>0 matches theory, TV=0.004 vs 0.65 for the raw target). Review
LOW nits also addressed (ddtree assert → 1024 blockDim cap; forced `--spec
dspark` absent-sidecar warning; BlockController interior-argmax test).

**deepseek4 temp>0 DSpark serving is now un-gated.** It was hardcoded greedy-only:
`generate_deepseek4_spec` called the unified `generate_spec` with `temp=0.0` and
never `set_sampling`, reached only at `temp<=1e-6`; `carriers.rs` built the
speculator with `supports_temp=false`. The τ-adaptive block controller already
makes temp>0 DSpark beat AR — the daemon just never routed to it. Now temp>0
routes to the spec loop when the drafter samples (`!requires_greedy()`, the same
predicate arch_id=7 uses), threading temp/top_p/top_k/cactus through.

**CACTUS** (acceptance boost) is ported to the deepseek4 DSpark lazy sample-accept
kernel and exposed as a **per-request opt-in** (`cactus_delta`, default `0.0` =
lossless/distribution-preserving; never default-on). Deliberately lossy at δ>0 —
trades KL-bounded distribution fidelity for higher acceptance τ.

### Daemon-path benchmark @ temp=1.0 (gfx1151, warm, decode tok/s)

| config | code | prose | math |
|---|---|---|---|
| AR | 14.7 | 14.7 | 14.7 |
| DSpark δ=0 (lossless) | **25.2** (+72%) | 14.7 (≈AR) | 14.9 (≈AR) |
| DSpark δ=0.5 | 26.2 (+78%) | 18.8 (+28%) | 16.5 (+13%) |
| DSpark δ=1.0 | **27.8** (+89%) | **20.7** (+41%) | **20.4** (+39%) |

Lossless DSpark already beats AR on code; CACTUS lifts prose/math (where the
drafter aligns poorly at temp=1.0) to +39–41%. All cells coherent — greedy
coherence gate green, and the temp=1.0 outputs are correct across genres
(iterative Fibonacci + a 2nd variant; Rayleigh-scattering prose; a correctly
set-up trapezoidal-volume math solution).

### Reproduce

```bash
cargo build --release --example daemon --features deltanet -p hipfire-runtime
MODEL=~/.hipfire/models/deepseek-v4-flash.mq2lloyd   # its -dspark sidecar auto-loads

HIPFIRE_EMIT_TOKEN_IDS=1 HIPFIRE_DEEPSEEK4_SPEC_DECODE=1 \
  ./target/release/examples/daemon <<JSONL
{"type":"load","model":"$MODEL","params":{"max_seq":4096,"dspark_mode":"on"}}
{"type":"generate","id":"warm","prompt":"Say hello.","temperature":1.0,"max_tokens":128,"cactus_delta":0.0}
{"type":"generate","id":"code","prompt":"Write a Python function that returns the nth Fibonacci number, with a short docstring.","temperature":1.0,"max_tokens":200,"cactus_delta":1.0}
{"type":"unload"}
JSONL
```

- **AR baseline:** `"dspark_mode":"off"`.
- **Lossless spec:** `"cactus_delta":0.0` (distribution-preserving).
- **Opt-in boost:** `"cactus_delta":0.5`–`1.0` (lossy; +τ). `dspark_bench` sweeps the
  same via `HIPFIRE_DEEPSEEK4_CACTUS=<δ> HIPFIRE_DEEPSEEK4_TEMP=1.0`.


nwoolmer (commented):
## Code review — generic DSpark engine + qwen3 port + GPU sampler + τ-adaptive block

Reviewed the incremental slice `#484..HEAD` (~148 commits, ~19k lines) across four axes: the arch-agnostic engine core + block controller (`dspark_core.rs`, `dspark_block_controller.rs`), the qwen3 port (`dspark_body.rs`, qwen35 refactor), the GPU sampler/attention/GDN kernels + `rdna-compute` wrappers, and the serving/loader/CLI wiring.

**Verdict: mergeable after deciding on the CACTUS item below. No CRITICAL or HIGH correctness bugs.**

### Verified correct
- **Distribution-preserving temp>0 verify is sound.** The committed token is *always* the target's own draw `t_i ~ p_T(temp, top_p, top_k)`; accept-iff-`draft == t_i` is pure KV/hidden-reuse bookkeeping and never changes what's emitted. This is the point-mass degenerate case (not Leviathan accept-ratio) and is correct precisely *because* the target's sample — not the draft — is what gets committed. Greedy path (`temp ≤ 1e-6`) stays byte-identical.
- **Lazy prefix-stop / `accept_greedy_prefix`**: `u32::MAX` padding is never accepted, the stop fires only at the first mismatch, the bonus slot is always a real token, and `lazy=true` is correctly restricted to the two capture callers (grep-verified). No off-by-one in prefix length or τ counting.
- **Accepted-prefix hidden reuse**: position algebra traced — `capture[0..k+1]` maps to absolute positions `P..P+k` using the *current* window's `P` (not stale), next window's block starts contiguously at `P+k+1`, and bootstrap runs once per generation. Faithful to the reference `_update`.
- **`BlockController` math**: the `τ(N)` recurrence, argmax bounds (arrays sized 9, indices ≤ 8), divide-by-zero guards (`t_ar > 0` calibration gate), count-based survival (the cap-trap fix), and `reset()` preserving calibration all hold. 8 no-GPU unit tests pass.
- **qwen3 `dspark_body.rs`**: both pre-merge bugs are genuinely fixed — the head-path double-norm (single `output_norm` downstream) and the arch-specific confidence-norm (`confidence_uses_normed=true` for qwen3 vs raw for ds4). GQA grouping, RoPE positions, and the `[ctx++block]` bidirectional mask are cited against modeling.py and consistent.
- **Config ladder & serving gates**: the CLI-shadow fix (`dspark_conf_threshold` default `0.5 → null`, forwarded only when non-null) is real and complete — no other knob (`--spec`, `--draft-max`, block size) has the same pattern. **deepseek4 temp>0 is double-gated OFF** in serving (both a `temp ≤ 1e-6` routing guard and `supports_temp=false`). Sidecar precedence DSpark > DFlash > n-gram yields exactly one drafter; a missing sidecar warns + falls back to AR, never panics.
- **Mechanical multi-file changes are consistent**: the `eff_block_start` hipGraph-safety fix is applied correctly across all ~10 `attention_flash_*_tile_batched.hip` variants; the GEMM `__launch_bounds__` occupancy bump is functionally inert across its 4 files. The chunked GDN kernel (chunk math, cross-chunk state carry, race-freedom) and `dspark_sample_accept_lazy` are correct and have real parity harnesses.

### The one substantive finding
- **MEDIUM — CACTUS rejection samples the wrong distribution on GPU** (`kernels/src/chain_accept_spec.hip:171-210`). On rejection with `cactus_delta > 0`, the host rewrites `target_probs` into the CACTUS h-distribution *before* `sample_residual` (confirmed at `speculative.rs:3661-3678`); the GPU kernel skips that rewrite and samples `residual(p_t, p_d)` from the raw target. So committed rejection-bonus tokens are drawn from a **different distribution** than the host reference for the same inputs. **Scope: opt-in only** — `cactus_delta` defaults to `0.0` and is documented as deliberately lossy, and the accept *decision* is equivalent (the `.min(1.0)` vs `.min(p_d)` clamp difference doesn't change accept prob in the region where they diverge). Recommend either implementing the h-distribution rewrite in-kernel or asserting `cactus_delta == 0` on the GPU accept path. Also, **the kernel comment (lines 64-68) is factually wrong** — it claims the host uses the same simplified approximation and cites `speculative.rs:3726-3727`, but those are the accept-path lines; the actual host rejection code at `3661-3678` builds the full h-distribution.

### Quality follow-ups (non-blocking)
- **MEDIUM (test coverage)** — there is **no parity harness for `chain_accept_spec_f32` / `batched_categorical_sample_f32`** against the host accept loop. These are the highest-risk distribution kernels (they decide the committed token) and are validated only indirectly; a dedicated byte/statistical accept-parity gate would have caught the CACTUS divergence. `BlockController` tests also only assert monotone extremes — an interior-argmax case (where the cost model should pick a specific N=3) would lock the actual selection logic.
- **LOW** — `softmax_temp_batched.hip` reductions use the power-of-two tree pattern, but the wrappers clamp `block = min(256/1024, vocab)`; a small non-pow2 `vocab` would silently drop the orphan tail element. Latent (all production vocab ≫ 1024). `ddtree_build_attn_mask_f32` asserts `big_n ≤ 4096` but launches `blockDim.x = big_n` (AMD caps at 1024) — tighten the assert. `--spec dspark` with a totally-absent sidecar is a silent AR fallback (deserves a warning). `HIPFIRE_DDTREE_BUDGET=0` now reads as *unset* rather than force-disable (documented, benign).

### Two diffstat framings worth correcting for reviewers
- **`ddtree.rs` is not deleted** — it's still ~1927 lines and heavily used; the "360 ----" in the diffstat is a *net* per-file change (removed experimental "path C"), not a file deletion.
- **`speculative.rs` is a net −674 consolidation** (dead experimental paths removed, sampling helpers de-duped to `pub(crate)`), not a rewrite. All affected crates + the daemon example compile clean; grep confirms zero dangling references to any removed symbol.

### Note on verification limits
I can't run the GPU here, so on-device numeric byte-identity (the deepseek4 migration onto the shared core, qwen3 attention numerics) rests on the parity/coherence gate scripts (`qwen3_dspark_parity.rs`, `coherence-gate-qwen3-dspark.sh`) — present in the diff but not executable in this review. Confidence is high on the control-flow/gate findings and the sampled-verify distribution argument, medium only where on-GPU execution would be needed to confirm numeric equivalence.

---
🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

PR #449, opened 2026-06-12
Reconcile PP+MTP daemon path from #382

Kaden-Schutt:
## Summary

Surgical current-master reconciliation of the useful PP+MTP pieces from #382.

This intentionally does **not** import the old `SpecPath` / `GenerateCtx` dispatcher rewrite or the full old branch history. It adds the narrow pieces needed to load a compressed MTP head on the PP output device and run greedy PP+MTP decode through the daemon.

## What changed

- Add `forward_prefill_batch_multi_with_caps` so PP verify can capture output-band hidden rows and elide last-token logits when requested.
- Add PP-safe compressed MTP stepper with output-device MTP head, mirrored token embeddings, and multi-device DeltaNet snapshot/restore.
- Add daemon load/generate wiring for `pp > 1` + explicit compressed `.mtp` head.
- Add `hipfire_runtime::mtp_mirror` tensor clone helpers.

## Scope intentionally left out

- Old #382 `SpecPath` / `GenerateCtx` dispatcher rewrite.
- Old PP GDN tape-shard fast replay. This branch uses full accepted-prefix replay on partial accepts for correctness first.

## Validation

Local:

```bash
cargo check -p hipfire-runtime --example daemon
git diff --check
```

hiptrx rerun after avoiding a midflight collision with another agent:

- Commit: `6d86816f`
- Hardware: `ROCR_VISIBLE_DEVICES=2,3`, 2x Radeon AI PRO R9700 (`gfx1201`), HIP 7.2
- Trunk: `/home/kaden/.hipfire/models/qwen3-35b-a3b.mq4-awq`
  - md5 `edde51ec1dac0f2bd42cff5ef1cb8944`
  - sha256 `1dc1c7964de415e0040a540a4300b9518e11b00c13d99c23f576f2b9fe1e8bca`
- MTP head: `/home/kaden/.hipfire/models/qwen3.6-35b-a3b.moe-mtp-mq4-cvs16384.mtp`
  - md5 `51076bfb5b489832f2f6c4191b9799b0`
  - sha256 `1e11a06d1946e1e5711d6692894e917a61d5f360d4f3508c8372d49e97c912c1`

Load params:

```json
{
  "max_seq": 4096,
  "pp": 2,
  "kv_mode": "q8",
  "mtp_mode": "on",
  "mtp_head": "/home/kaden/.hipfire/models/qwen3.6-35b-a3b.moe-mtp-mq4-cvs16384.mtp",
  "mtp_k": 5,
[...]

PR #527, opened 2026-07-12
[Superseded] refactor: unify device mesh, model parallelism, and generation dispatch

fivetide:
## Status: superseded by upstream-native port plan

**Do not merge or rebase this PR wholesale.** The active replacement plan is tracked in **#666**.

This PR remains open temporarily as immutable forensic provenance for the original device-mesh implementation, review discussions, physical-GPU reports, parity hashes, and architecture decisions. It is no longer the active implementation or status authority.

## Why the pivot

The feature work is original and remains relevant, but the branch history diverged from current upstream:

- GitHub reports 3,109 changed files, 906,419 additions, and 106,919 deletions against current master.
- The branch/upstream common base is old (`78065c084c1c67f64d282bd904b1b28cd94aca42`).
- Branch merge `bac02a1a2` includes fork-mainline parent `919f80d9...`, which is not an ancestor of current upstream.
- A trial merge against current upstream produced hundreds of genuine add/add and content conflicts and was aborted safely.
- Current upstream has newer ownership and runtime seams—`ArchModel`, carriers, `hipfire-generate`, engine scheduler/terminal, current Qwen TP/VL work, and merged LFM2-VL PR #653—that must be reused instead of overwritten.

Therefore the implementation will be ported from current upstream master as grouped, reviewable change sets rather than replaying this branch history.

## Replacement delivery model

Issue #666 defines the approved grouped approach:

- fine-grained proof obligations remain independently visible;
- delivery is grouped into G0-G15 coherent change sets;
- dependencies are executable seam gates rather than one serial foundation chain;
- reset/lifecycle, model-loading manifests, and the interface-safe part of MoE-to-Step development run in parallel;
- each grouped change set has one sole owner, production route, evidence packet, and revert identity;
- every positive route is paired with its exact fail-closed or rollback route;
- historical, semantics-only, emulated, failed, or host-local evidence cannot become current upstream acceptance without rerun;
- physical PP/TP/EP obligations still require qualifying distinct-GPU evidence.

The machine-checkable working authority is pinned to the public `fivetide/hipfire` branch `feat/device-mesh-port-tracker`; **PR #668** proposes importing it into upstream but does not block G1-G5 implementation. Upstream implementation PRs mirror their carried obligation subset and reference the exact fork tracker commit. `docs/VALIDATION.md` remains the validation-route authority and `docs/admissions.yml` remains the admission authority.

## Grouped change sets

- **G0:** upstream port authority and PR #527 supersession
- **G1:** sole hardware owner plus named DeviceMesh topology
- **G2:** source-aware capability policy and effective-mesh admission
- **G3:** manifest planning and transactional model-loading pilot
- **G4:** reset, abort, parser, and terminal lifecycle contracts
- **G5:** collective-aware MoE Step architecture and sealed expert ownership
- **G6:** Gemma typed cutover
- **G7:** MiniMax lifecycle and typed cutover
- **G8:** Qwen35 state ownership and typed cutover
- **G9:** DeepSeek4 typed cutover
- **G10:** LFM2 typed cutover using merged PR #653 seams
- **G11:** Muse/Glimmer semantic design and typed cutover
- **G12:** generation and batching consolidation
- **G13:** dense parallel axes
- **G14:** expert parallel axes
- **G15:** vision parallel axes
- topology-class physical evidence campaigns and final aggregate closure follow the implementation groups

Full dependencies, evidence rules, parallel lanes, and stop conditions are in #666.

## What remains authoritative here

The following are retained as historical evidence inputs, not current upstream pass claims:

- original `DeviceMesh`, manifest, placement, ownership, Step, PP/TP/EP, and generation design decisions;
- branch tracker and inventory snapshots at their recorded commits;
- deterministic raw-versus-Step, Single/PP/TP/EP, reset, and lifecycle reports;
- contributor comments and physical R9700 findings, including the unresolved illegal-access and token-58 divergence failures;
- reviewed local STEP-005 Gemma/MiniMax work and exact parity evidence;
- explicit blocked/deferred tasks and failed probes.

No old checkbox is copied as complete into the upstream tracker. Each obligation receives an evidence disposition and must pass its current upstream route.

## Provenance anchors

- Remote forensic head: `5c6f5cc6b7107a09559ae3fe94a12cccafc97c3c`
- Latest reviewed local forensic head: `396aa7191ba53cf6c7c23e8827b0add5f79a2f45`
- Upstream base observed at pivot: `6163369329d3b076376286a00c13cadbae069ecc`
- Replacement planning issue: #666

This PR will be closed after the upstream authority lands and #666 contains durable links to every replacement change set. Until then, discussion should treat this PR as read-only historical context.


taniguchi-taku-softm:
# Physical PP Validation Findings on 2× R9700 — Partial, Blocked

Date: 2026-07-14
PR: https://github.com/Kaden-Schutt/hipfire/pull/527
Commit: `4df035373669369484797abdd274f3f710c4c061`

---

## Environment

| Item | Value |
|---|---|
| GPU | 2× AMD Radeon AI PRO R9700 (gfx1201, RDNA4) |
| Interconnect | PCIe Gen5 x8 (2 hops, weight 40), same NUMA node |
| P2P | ✅ Enabled (bidirectional, verified via `rocm-smi --showtopoaccess` and `gpus.enable_peer_all()` in `pp_parity`) |
| VRAM | 32 GiB each (64 GiB total) |
| ROCm | 7.2.4 |
| amdgpu kernel driver | 6.16.13 |
| Kernel | 6.17.0-35-generic |
| RCCL | 2.27.7.70204 |
| HIPFIRE branch | `feature/device-mesh` (PR#527) |

### Artifact fingerprints

| File | Size | SHA-256 |
|---|---|---|
| `qwen3.5-9b.mq4` (arch 5, 32 layers, DeltaNet) | 5.3 GB | `ba83acf5bfd5d4e334b0afc26d779734e31623bb7f74e807c3581dfecb3128ad` |
| `qwen3-0.6b.hf4` (arch 1, 28 layers, plain Qwen3 via LLaMA carrier) | 436 MB | `7760b19dfb940f8b33078eb524602b4f2b5e6825c6e10c466e6e99bcfc133838` |

**Note:** HW-003's requirement specifies `qwen3-0.6b-llama.mq4` (arch 0, pure LLaMA). That exact artifact is not on disk — tested with `qwen3-0.6b.hf4` (arch 1, plain Qwen3 via LLaMA carrier) instead.

---

## Results

### 1. VRAM split: `pp2_vram_probe` ✅ PASS (Qwen35, physical PP-2)

```
qwen3.5-9b.mq4: layers=32, dim=4096, hidden=12288, vocab=248320, kv_heads=4, head_dim=256

── per-card VRAM deltas (PP=2 split) ──
        baseline:  dev0= 0.000 GiB  dev1= 0.000 GiB
       Δ weights:  dev0= 2.638 GiB  dev1= 2.638 GiB
       Δ scratch:  dev0= 0.013 GiB  dev1= 0.013 GiB
            Δ KV:  dev0= 0.134 GiB  dev1= 0.134 GiB
      Δ DN state:  dev0= 0.006 GiB  dev1= 0.006 GiB
───────────────────────────────────────
           total:  ~2.8 GiB / card    (8.7% of 32 GiB)
```

- Equal per-stage weight, KV, and DeltaNet state allocation confirmed

### 2. Physical peer access ✅ Confirmed

`rocm-smi --showtopoaccess` reports bidirectional P2P. `gpus.enable_peer_all()` in `pp_parity` succeeds without error. Actual stage-boundary data transfer (via `boundary_copy` in the PP forward path) executed implicitly during `pp_parity` run — no explicit transfer-trace log was captured.

### 3. Token parity: `pp_parity` ❌ (58/100 match, Qwen35 PP=1 vs PP=2)

PP=1 (single GPU) vs PP=2 (real 2 GPU) greedy decode (temp=0):

- **First 20 tokens**: identical
  `[5328, 1301, 279, 2614, 1414, 321, 4087, 279, 3296, 3559, 13, 271, 1138, 25, 198, 760, 2614, 369, 264, 11782]`
- **58/100 tokens match** → first divergence at index 58
  - PP=1: `[... 3694, 314, 29687, 13, 198, 760, 3788]`
  - PP=2: `[... 3694, 314, 6824, 11, 30954, 11, 321]`

**Working hypothesis:** numerical differences introduced around the DeltaNet stage-boundary state transfer, which may be amplified by recurrent state evolution. This has not yet been isolated from possible synchronization, copy, or state-ownership defects.

**Note on HW-004 closure:** HW-004 requires PP=2 output to byte-match single-device cold generation **and** a two-turn fixture by hash. This result does not satisfy either criterion — the divergence is unresolved, not "expected."

### 4. Load/unload cycles: `pp_unload_reload` ✅ PASS (100 cycles, emulated PP)

LLaMA model (`qwen3-0.6b.hf4`, arch 1, **emulated** PP-2 via `HIPFIRE_EMULATE_GPUS=2`), 100 load→unload cycles with per-cycle VRAM monitoring:

```
VRAM baseline: 0.101 GiB
cycle    0: loaded=0.487 GiB  after_unload=0.231 GiB  drift=0.130 GiB
cycle   49: loaded=0.487 GiB  after_unload=0.231 GiB  drift=0.130 GiB
cycle   99: loaded=0.487 GiB  after_unload=0.231 GiB  drift=0.130 GiB
───────────────────────────────────────────────────────────
VRAM final:   0.231 GiB
total drift:  0.130 GiB (zero growth after cycle 0)
```

- **130 MiB drift occurs on first cycle only, then stays flat for 100 cycles** — one-time HIP runtime context + kernel code object overhead, not a leak
- Zero panics or crashes across 100 cycles

**Caveat:** This test exercises the `load_model_pp` + `unload_model` code path under emulation, not physical per-device teardown. HW-003 requires physical per-GPU VRAM recovery to within 64 MiB of post-first-unload baseline with no monotonic increase across cycles 2–4. That has not been tested.

### 5. Manifest placement parity: `store-pp-gate.sh` ❌ Forward failed (emulated PP)

`llama_store_pp` example under `HIPFIRE_EMULATE_GPUS=2`:

```
llama_store_pp: PP-2 placement OK — 155 tensors on stage 0, 156 on stage 1
  (embed→0, output-norm+lm_head→1, 28 layers banded by stage_for_layer)
```

- **Placement**: ✅ Correct manifest-driven stage split confirmed (bit-exact tensor placement)
- **Forward**: ❌ `hipMemcpy D2H: an illegal memory access was encountered` at `forward_logits` → `scratch.logits` download failed
- **Root cause**: LLaMA forward path (`forward_scratch_compute` + `LlamaWeights`) has a gfx1201 (RDNA4) kernel compatibility issue. Qwen35 path (`forward_scratch_multi`) works fine on the same hardware.

### 6. Daemon serving parity: ⏳ Not tested

HW-003 requires end-to-end daemon serving with committed-token hash matching. Not yet run.

---

## Issues Found

### Issue A: gfx1201 × LLaMA forward illegal memory access

LLaMA-family (arch 0/1) `forward_scratch_compute` crashes on gfx1201 with an illegal memory access during logits download. Qwen35 (arch 5+, DeltaNet) works correctly on the same hardware.

**Impact:** Blocks all LLaMA-based PP forward testing on gfx1201 (HW-003 dense, plus any arch 0/1 coverage). Likely a missing gfx1201 kernel variant or ISA compatibility gap.

### Issue B: 58/100 token divergence between PP=1 and PP=2 (Qwen35)

DeltaNet PP-2 output diverges from single-device output at token 58. Root cause not yet isolated — candidates include FP accumulation differences across stage boundary, synchronization defects, copy-size mismatch, or state-ownership bugs.

---

## Summary

| Requirement | Status | Notes |
|---|---|---|
| Qwen35 physical weight/state split | ✅ | Equal per-stage allocation (2.638 GiB/card) |
| Physical peer access | ✅ | Bidirectional P2P enabled + verified |
| Physical stage-boundary execution | ⚠️ Partial | Qwen35 forward ran; explicit boundary-copy trace not captured |
| Qwen35 token parity (PP=1 vs PP=2) | ❌ | Diverges at token 58; cause unconfirmed |
| Dense LLaMA physical placement | ❌ Blocked | Placement confirmed under emulation only; real forward crashes gfx1201 |
| Dense LLaMA forward parity | ❌ Blocked | gfx1201 illegal memory access in `forward_scratch_compute` |
| Physical unload/reload stability | ⏳ Partially tested | 100 emulated cycles clean; physical per-GPU teardown not done |
| PP-only capacity gain | ⏳ Not demonstrated | Test model (5.3 GB) fits on one GPU — no PP-required workload tested |
| Daemon serving parity | ⏳ Not tested | Required for HW-003 closure |

### Remaining work

1. **Fix gfx1201 LLaMA forward crash (Issue A)** — the PR blocker for HW-003 dense
2. **Isolate DeltaNet PP divergence cause (Issue B)** — synchronization, accumulator, or state-ownership
3. Run physical per-GPU VRAM teardown test (HW-003 § load/unload criterion)
4. Run HW-003 manifest gate (`store-pp-gate.sh`) on physical PP after Issue A fix
5. Run end-to-end daemon serving with token hash comparison
6. Retest with the canonical `qwen3-0.6b-llama.mq4` artifact (pure LLaMA, arch 0)


Sorry, i dont have ai pro x4...

fivetide:
Thank you—two real R9700s are exactly the validation we needed. The balanced PP allocation and verified P2P result are valuable evidence.

We’re recording the gfx1201 LLaMA illegal-access result as a separate classification blocker; the HF4 artifact and emulated `llama_store_pp` path do not yet prove a physical PP regression. We’ll reproduce it with the canonical fixture and capture the first failing HIP launch before assigning a cause.

The Qwen35 PP divergence remains a hard blocker for HW-004, not expected numerical drift. After the pending generation-path work, we’ll re-run deterministically with pinned artifacts/prompts and capture boundary, recurrent-state, and copy-order diagnostics.

Thanks again for the thorough environment, artifact, and VRAM data—and no apology needed; a real two-GPU report is precisely what this draft needed.

Kaden-Schutt:
Carry-forward notes from two PRs being closed in favor of this device-mesh ownership spine:

- **#382 / narrow replacement #449 (`6d86816f`) — PP+MTP:** do not inherit the original sprawling historical branch. The reviewable slice was `forward_prefill_batch_multi_with_caps`, a PP-safe compressed MTP stepper, `mtp_mirror` tensor-clone helpers, daemon load/generate wiring for `pp > 1` plus an explicit compressed `.mtp` head, and the end-of-generation cycle-depth guard. That slice was validated on hiptrx with a 2-way PP split and emitted `spec_path:"pp-mtp"`. Please carry the capability forward within the unified mesh/generation architecture if it still fits.

- **#495 — Qwen PP rotated-KV Givens ownership:** `forward_prefill_chunk` must prefer the per-device `PrefillBandCtx.givens_cos/sin` tensors for FWHT/asym KV under PP rather than reading the global `kv_cache.givens_*` fields, which are intentionally `None` in the multi-GPU constructors. The patch is tiny, but the acceptance test needs a real multi-GPU rotated-KV prefill.

Neither stale branch is being merged wholesale into beta; this comment preserves their intent where the new ownership model can absorb it cleanly.

fivetide:
## COR-005 Progress Update — 2026-07-19

**Status: complete.** Commit `688bb42e` makes generic LLaMA/Qwen3 speculative-target and DFlash construction transactional across target loading, scratch/KV allocation, PaRo/AWQ/DSpark paths, malformed-HFQ handling, and post-allocation helper cleanup. It adds deterministic, feature-gated fault injection and records the acceptance evidence in the authoritative tracker.

**Hardware evidence:** on UMA gfx1151 (HIP 7.2), using `qwen3-8b.q8f16.hfq` and `qwen3-8b-dflash.hfq`, the feature-gated generic DFlash rollback gate passed in 1154.75s. It swept every generic target/DFlash fault stage plus repeated load/prefill/step/unload cycles. The gate synchronizes and drains the pool; its 64 MiB VRAM envelope accounts for observed ROCm graph/allocation accounting variance.

**Validation:** `cargo test` passed; the targeted GPU gate passed; changed Rust files pass `rustfmt --check`; and `git diff --check` passed. Workspace-wide `cargo clippy --all-targets -- -D warnings` remains blocked by unrelated existing warnings in `redline`, `hsa-bridge`, `hip-bridge`, `hipfire-detect`, `hipfire-atlas`, and unmodified regions of `rdna-compute`; no unrelated lint cleanup was included.

nwoolmer:
### Integration points with #563 (routed-expert paging) — for later, no action needed here

#563 pages routed experts from the HFQ into bounded per-layer slot pools. Both PRs touch
expert weights, so I checked the overlap and pre-built the seams on our side. **#527 needs
no changes** — the only shared surface is a `pub fn` this branch already exports.

Recording it here so whoever merges second has the map.

**Already compatible, verified against this branch:**

- **Pointer table** — unchanged. `ptr_tensor` → `ptr_slot` is a local rename; still
  `[2·n_exp]` F32 with the `to_ne_bytes` u64 encoding #563 pins in a contract test.
- **Per-rank weights** — `load_weights_sharded` / `load_weights_tp` each return their own
  `DeepseekV4Weights`, so #563's pager (a `Mutex` field on `Weights`) is per-rank. No
  shared-pager or cross-deviceproblem.
- **Loader entry** — #563's paging block lives inside `load_weights_inner`, so the
  `tp_slice` parameter added here is already in scope at the point it's needed.
- **`select_grouped_lloyd_variant` / `dispatch_grouped_lloyd` → `pub(crate)`** helps:
  #563 mirrors that selection for its banded prefill and should call it instead (it
  already drifted once against beta's i8-WMMA arm).

**TP expert slicing — the one real interaction.** Under TP a paged fill must READ the full
expert and WRITE a row-gathered subset, so read length ≠ write length. #563 has landed that
seam (`Transport::read_batch`, `ExpertFillTransform`, packed-stride pool sizing) and proves
it with a synthetic 1/tp transform, so only the concrete gather is missing. To enable once
this lands:

1. thread `tp_slice` into `alloc_paged_layer_expert_pool`, sizing the pool with the packed
   stride (`stride / tp`) to match what this branch writes to `expert_*_stride`;
2. `set_fill_transform(Box::new(TpRowGather { .. }))` wrapping
   `weight_store::expert_tp_row_gather`;
3. delete #563's TP guard.

The expert catalog needs no change: it records **full HFQ ranges**, and the on-disk layout
is TP-independent.

**EP — the seam is now in too** (`a2c5a940`). #563 has an `ExpertOwnership` (`All | Subset`)
and `plan_dispatch` skips non-owned experts in all three places that matter: the pool
capacity count, the catalog validation, and the resolve loop. So a non-owned expert is never
resolved, never read, and never repointed — it keeps aiming at this branch's zeroed dummy and
contributes zero.

That last point is the whole correctness argument, and it has no runtime error to catch it:
repointing a non-owned expert makes the rank contribute a value the OWNING rank also
contributes, and the all-reduce double-counts. It is pinned by a test that fails with exactly
that message if the filter regresses (verified by neutering the filter — 3 of 5 new tests
fail).

Worth knowing: **paging replaces EP's compaction rather than composing with it.** This branch
packs owned experts densely via `local_of_global`; paging assigns slots dynamically. Both
yield `base + index·stride`, so the pointer-table format is identical — which is why it drops
in rather than fighting the existing scheme.

To enable EP + paging: have `alloc_paged_layer_expert_pool` build the zeroed dummy and aim
non-owned experts at it (it currently points everything at base), call
`set_ownership(Subset(..))` from `owns_expert(rank, e)`, size the pool against `n_owned`, and
lift #563's EP guard. The guard stays for now only because EP cannot be tested on a single
GPU — the seam is ready, the activation is not claimed.

**Known gaps, flagged not fixed:**

- The safetensors path (`upload_layer_routed_experts_from_source`) has no paging; the
  catalog is built from the HFQ tensor index, so paging is HFQ-only by construction.
- `load_weights_inner`'s expert section will need a hand merge — architecturally
  compatible, textually overlapping.
- #563's TP detection is duck-typed on `HIPFIRE_TP` / `HIPFIRE_TP_EXPERT_SLICE`, inferred
  rather than confirmed against this branch's activation path. Worth a look when retargeting.
- The gather would run **per fill** under paging, not once at load. On a ~4.7 MB expert
  that's a strided host copy on a path already read-dominated; TP+paging would be *correct*
  but its cost is unmeasured. A TP-aware layout at quantise time (slice contiguous on disk)
  may be the better answer.

One process note: #527 targets `master`, #563 targets `beta`, and `beta` is ~488 commits
ahead. Whichever lands second inherits a larger merge than the file overlap suggests.

Kaden-Schutt:
Triage: HOLD_WIP. DeviceMesh/ModelParallel remains architecturally relevant but the author checklist, PP correctness, and hardware axes are incomplete; it also overlaps active loader placement work.

fivetide:
## STEP-005 Gemma4 increment complete

Branch head: `58320b336` (pushed to `feature/device-mesh`).

The Gemma4 family slice of STEP-005 is complete and independently reviewed:

- removed Gemma4 production `SuperOp`/`run_layer_program` execution;
- added typed borrowed Steps for the required dense/MoE operations;
- closed eager/lowered terminal, rollback, cancellation, exact-prefix cache, reset, and ownership contracts;
- dense Q8 sequential/batched parity passes B=1/2/4/8;
- exact E2B/E4B parity passes;
- official HF MoE reference established; admitted Q8-expert/F32-KV artifact matches canonical argmax `107`;
- production battery and chain are coherent with `empty=0`, `attractor=0`, and `retrieval_miss=0`;
- historical MQ4 MoE artifact is explicitly rejected as over-quantized.

Validation at the closure head: `cargo build && cargo test` passed **4,024 tests**, **75 ignored**, **0 failed**. Final Step-contract and quality reviewers both approved with no findings.

The dense Gemma Q8 batched path intentionally keeps scalar batched GEMM instead of WMMA. The rejected WMMA route failed KV-conditioned next-token parity; the source now records the exact failure, the measured B=2 cost (56.629 ms WMMA vs 93.811 ms scalar, +65.7%), and the evidence required before re-enabling it.

Authoritative evidence: `.agent-progress/step-005-production-inventory.md`. The tracker is synchronized in `.agent-progress/device-mesh-refactor-tracker.md`.

**Status precision:** this closes the **Gemma4 increment**, not aggregate STEP-005. LFM2, MiniMax, Qwen35 Single/EP, and DeepSeek4 Single/EP remain open under STEP-005, so the PR checklist stays unchecked until those family rows and the final no-SuperOp inventory close.

PR #532, opened 2026-07-17
fix: remove noisy RDNA2 GEMV variant eprintln

### LINKED ISSUE #531: Remove noisy 'RDNA2 GEMV variant' eprintln from gemv_hfq4g256_for_arch
## Problem

`gemv_hfq4g256_for_arch` in `crates/rdna-compute/src/kernels.rs` prints:

```
RDNA2 GEMV variant: v1 (baseline-rdna2)
```

on every GEMV kernel selection for gfx1030/gfx1031 (RDNA2). This produces hundreds of identical lines in `serve.log` and bench output — pure noise with no actionable information.

## Fix

Remove the `eprintln!` call. The variant info is already available via `HIPFIRE_RDNA2_VARIANT` env var and the pre-compiled kernel cache directory name (`.hipfire_kernels/gfx1030/`).

rakaarwaky:
## Summary

Removes the `eprintln!("  RDNA2 GEMV variant: v{variant} ({name})")` call in `gemv_hfq4g256_for_arch` (`crates/rdna-compute/src/kernels.rs`).

## Problem

On RDNA2 (gfx1030/gfx1031), every GEMV kernel selection prints this line to stderr:

```
RDNA2 GEMV variant: v1 (baseline-rdna2)
```

Since a single decode step involves many GEMV calls, this produces hundreds of identical spam lines in `serve.log` and `hipfire bench` output. The variant information is not actionable for users — it's only useful for kernel developers who can check the variant via `HIPFIRE_RDNA2_VARIANT` env var or the pre-compiled cache directory name.

## Testing

- Verified the `eprintln!` string no longer appears in the compiled binary via `strings`
- Tested `hipfire bench qwen3.6-35b-a3b.mq2` — no more RDNA2 GEMV variant spam in output

Closes #531

### DIFF
--- a/crates/rdna-compute/src/kernels.rs
+++ b/crates/rdna-compute/src/kernels.rs
@@ -2346,7 +2346,7 @@ pub fn gemv_hfq4g256_for_arch(
                 "cache-aggressive",
             ];
             let name = names.get(variant as usize).unwrap_or(&"baseline-rdna2");
-            eprintln!("  RDNA2 GEMV variant: v{variant} ({name})");
+            // eprintln!("  RDNA2 GEMV variant: v{variant} ({name})");
             match variant {
                 2 => (GEMV_HFQ4G256_GFX1030_V2_SRC, "gemv_hfq4g256_rdna2v2"),
                 3 => (GEMV_HFQ4G256_GFX1030_V3_SRC, "gemv_hfq4g256_rdna2v3"),


Kaden-Schutt:
Thanks @rakaarwaky — this is already fixed on the release line, so I'm closing this out.

Your change landed on `beta` as [`0432bac25`](https://github.com/warpfront/hipfire/commit/0432bac25) ("fix: remove noisy RDNA2 GEMV variant logging (#532)"), with your authorship on the commit. The version that landed goes slightly further than the diff here: rather than commenting out the `eprintln!`, it removes the surrounding dead block entirely — 10 lines deleted — so there's no commented-out code left behind in `kernels.rs`.

Verified against the current `beta` tip: the noisy per-dispatch log is gone.

Closing as superseded, not rejected. The fix is shipping in v0.3.0.

PR #553, opened 2026-07-29
perf(kernels): query-tiled Q8 flash prefill attention (scalar + WMMA)

nwoolmer:
## Summary

Query-tiled Q8 prefill attention for RDNA3, in two variants: a scalar
FlashAttention-style kernel and a WMMA matrix-core kernel. **Default-on for
gfx11xx**, off elsewhere; `HIPFIRE_FLASH_PREFILL=0` disables.

**End-to-end prefill: 1.91× at 14.7K context (295.5 → 563.3 tok/s), 1.56× at
11.5K, 1.24× at 7.7K.**

## Why

Prefill on gfx1151 / Qwen3.6-35B-A3B MQ4R halved by ~10K context (1015 tok/s @1K
→ 316 @14.4K). A rocprofv3 sweep pins that entirely on attention: every other
kernel scales with a fitted exponent of 1.00–1.05, while attention scales
**N^2.19** and grows from 12.3% of prefill GPU time at N=2048 to **64.3% at
N=12288**.

One design choice caused it. Both existing Q8 prefill attention kernels launch
**one workgroup per individual query**, so K/V is re-streamed per query with zero
reuse (~490 GFLOP/s measured), and `scores[]` LDS scales with context — which is
what forces the crossover at 8192 to `attention_flash_q8_0_tile_batched`,
measured **2.3× less efficient per unit of attention work**.

## What changed

`attention_q8_0_flash_prefill{,_wmma}` stage a tile of query rows and stream K/V
in tiles held in LDS as native Q8_0 blocks, with online softmax. **LDS depends
only on tile size, never on context** — no capacity ceiling, no occupancy decay,
and the 8192 crossover becomes unnecessary.

The WMMA variant is not a port of the `attention_dflash_wmma_*` family: those all
hard-guard `head_dim == 128`, and the one head_dim-generic variant needs ~84 KB
of LDS at head_dim=256. This fits in ~18 KB (3 WG/CU) by keeping LDS in f16 and
never staging K — each B fragment is dequantised straight from the cache, and the
16×16×16 fragment shape reuses it across 16 query rows, amortising the dequant.

## Performance (gfx1151, nh=8 nkv=2 hd=256, N=256), ms

| CTX | tiled | scalar | **WMMA** | legacy | WMMA/legacy | WMMA/tiled |
|---:|---:|---:|---:|---:|---:|---:|
| 2048 | 10.08 | 5.84 | **2.45** | 3.68 | 1.50× | 4.42× |
| 8192 | 45.34 | 25.00 | **10.28** | 22.02 | 2.14× | 4.44× |
| 12288 | 68.51 | 39.76 | **14.21** | 48.23 | 3.39× | 4.96× |
| 16000 | 89.88 | 52.59 | **20.29** | N/A (>64 KB LDS) | — | 4.32× |

WMMA beats the legacy kernel at every context, so no crossover or MIN_CTX gate is
needed. (The scalar variant loses below ~10.2K — 8× fewer workgroups — so it
keeps a break-even gate.) Per-position measurement confirms **the 8192 step is
gone**: 2.15 → 4.07 ms/tok across it before, 1.91 → 2.23 smooth after.

Two optimisations were measured and **rejected**: staging K in LDS (costs 8 KB,
occupancy 3→2 WG/CU, 2.26× *slower*) and keeping scores in f32 (58% slower for
zero accuracy gain). Occupancy dominates on this kernel.

## Correctness and quality

`test_q8_flash_prefill` vs `attention_q8_0_kv_batched`: 12 shapes (incl.
`seq_len % BC != 0`, `seq_len < BC`, `batch % BR != 0`, ragged `positions[]`) ×
GQA kv_group 1/2/4/8, plus head_dim=128. The gate rejects non-finite and all-zero
output — it previously passed an all-zero kernel vacuously.

The scalar kernel holds the fp32-reassociation bar. WMMA computes in f16 and is
held to relative L2 ≤ 5e-3 / cosine ≥ 1−1e-5; measured 5.3e-4 to 1.6e-3.

Quality vs the f32 path, **2048 paired positions** through
`forward_prefill_batch` (`flash_prefill_quality.rs`; note `perplexity.rs` scores
at batch_size=1 and never reaches these kernels):

| | base | split-Q |
|---|---:|---:|
| ppl vs f32 7.4812 | 7.5310 (+0.67%) | 7.4910 (+0.13%) |
| paired ΔNLL | +0.0066 nats | +0.0013 |
| significant at 0.05 | no (t=0.76) | no (t=0.27) |
| top-1 agreement | 95.17% | 95.46% |

No detectable regression; bounded below +2.4% ppl at 95% confidence. Top-1
divergence is confined to near-ties — median NLL 2.81 nats at disagreeing
positions vs 0.30 where the arms agree, and the f32 top-1 stays inside the f16
top-8 in 99.76% of cases.

`HIPFIRE_FLASH_PREFILL_SPLITQ=1` carries Q as a double-single pair (Q's f16
rounding is 98% of the error), cutting relative L2 4.5–5× for 1.6× time. Default
off: it narrows the distributional gap 5× but does not change which tokens are
chosen.

## Scope and safety

Causal, non-tree, non-windowed prefill only; tree-verify, sliding-window and
decode are untouched. Default-on is gated to gfx11xx — the WMMA intrinsic is
RDNA3+ and gfx12 needs the `_gfx12` variant this kernel does not use. Verified
inert when disabled (committed token stream identical to baseline, 3/3 runs), and
order-balanced replication found no added degeneracy (0/24 default-on vs 8/24
control).

Not included: raising `Q8_BATCHED_LDS_CROSSOVER` itself (separately validated
bitwise-identical at 12288; the real LDS bound is ~15.8K at head_dim=256).

Full design, plan and results — including an intermittent long-context
degeneracy that predates this branch and remains unattributed after four tested
hypotheses — are in `docs/perf-checkpoints/2026-07-29-*`.


Kaden-Schutt:
Merged into `beta` at bcd1ef008 (via 1b906a7fb). All 19 commits went in unrewritten, so authorship and PR linkage are preserved. One header fix: five of the new files had copied the house template and carried `Copyright (c) 2026 Kaden Schutt` for code you wrote — corrected to you in bcd1ef008. SPDX stays Apache-2.0 per CONTRIBUTING.md:319.

This composed cleanly with our in-flight attention work, which is worth spelling out because the sibling PR (#554) did not, and the difference is instructive. Your change **restructured** the algorithm — query tiling, online softmax, context-independent LDS — so it lands beside a load-widening hoist we had in flight. #554 **re-optimized the same inner loop** that hoist had already fixed, so the two competed and it needed a port to compose. Structural changes compose; point optimizations on the same loop collide.

Independent numbers on our hoisted tree, A/B/B/A with hashed binaries against the pinned 3.6-27B trunk:

| arch | pp 4096 | pp 8192 | pp 14336 |
|---|---|---|---|
| gfx1100 (RX 7900 XTX, 960 GB/s) | 1.0849× | 1.1294× | 1.0466× |
| gfx1151 (Strix Halo 8060S, 256 GB/s) | 1.1155× | 1.1861× | 1.1421× |

Spreads 0.29–0.45%. It wins **more** on the bandwidth-starved part, which is coherent with the mechanism. Our LDS-staged GEMM is unaffected by it being active (deltas +0.05 / −0.80 / +0.74 / +0.55 points, all inside spread), so the two are orthogonal.

Two findings that may save you time:

**The scalar path is a large loss on gfx12, so keeping it gfx11-gated was the right call.** `wmma_ok` requires gfx11, but the fallthrough runs the scalar variant on any arch above `MIN_CTX`, so on gfx1201 we measured **0.8986 / 0.7894 / 0.7329×** at pp 12288 / 16384 / 20480. No port was needed to find this — it's reachable today on gfx12 above 10240. Worth an arch guard on the scalar fallthrough rather than relying on the WMMA gate alone.

**The WMMA path is not a one-symbol port to gfx12.** I assumed it was, twice, before compile-probing: `cannot initialize a parameter of type 'vector of 8 _Float16' with an lvalue of type 'half16_t'` at both call sites. gfx12 takes `half8_t`, gfx11 `half16_t`. Real cost is 3 call sites plus fragment repack plus accumulator-map verification plus a parity run. Deferred, but its structural value on gfx12 is real — it would remove the 4096 crossover and the 15,872-token LDS ceiling.

One untapped axis if you revisit these kernels: neither of your two has widened K/V loads. The hoist that gave us 1.585× on the batched tile applies the same way here — `d0 = tid*dpt`, `dpt = head_dim/32`, `head_dim % 32 == 0` means a thread's dims never straddle a 32-quant block, so the scale is loop-invariant and the quants are contiguous. Bit-exact, since values and accumulation order don't change. Part of that gfx12 scalar deficit may be the missing hoist rather than the query-tiling structure.

Also: `scripts/refresh-credits.sh` only scans `--base master`, so beta-based PRs never reach `CREDITS.md`. Generator bug, flagged separately.

PR #554, opened 2026-07-30
perf(gfx12): reuse Q8 KV across four prefill queries, ~1.7x prefill

HUSRCF:
## Summary

This adds an opt-in gfx12 Q8 prefill attention tile that processes four adjacent causal queries per wave and reuses each dequantized K/V value across those queries. It preserves the existing global partials ABI and established reduce kernel, so the change is isolated to the bandwidth-dominant tile stage.

**No ready and wait in status of draft since gfx11 verification is in progress. And PR #553 had already had similar realization. Thus this PR is frozen to seek more aggressive performance pursuit.**

The route remains default-off behind `HIPFIRE_Q8_PREFILL_M4=1`. It is restricted to gfx12, causal Q8 prefill, head dimensions 128/256, M4-aligned batches, and sufficient partial capacity. Tree attention, sliding-window attention, unsupported shapes, and other architectures continue to use the existing kernels.

No CK shared object or external binary is linked or redistributed. This is a native hipfire kernel using the query-tiling/KV-reuse principle; the existing hipfire/Kaden provenance and SPDX convention are preserved.

- [x] `kernels/` (HIP source)
- [x] `crates/rdna-compute` (kernel dispatch / RDNA arch routing)
- [ ] `crates/hipfire-arch-toy`
- [ ] examples / daemon
- [x] docs / CI / scripts

## Performance

Hardware and configuration: Radeon AI PRO R9700 (`gfx1201`, 34.2 GB), HIP 7.14, Q8 KV, context limit 32,768, temperature 0, DFlash/MTP off, fresh daemon per condition, byte-identical prompts inside every A/B cell.

The full LongBench-v2 hard20 matrix used 20 complete 20.5K–30.2K-token contexts:

| Model / mode | Baseline prefill | M4 prefill | Speedup | Prefill reduction | End-to-end reduction |
|---|---:|---:|---:|---:|---:|
| 27B / no-think | 3,315.7 s | 1,895.5 s | 1.749x | 42.83% | 38.07% |
| 27B / think | 3,359.0 s | 1,943.2 s | 1.729x | 42.15% | 25.86% |
| 35B-A3B / no-think | 1,284.7 s | 707.2 s | 1.817x | 44.96% | 42.53% |
| 35B-A3B / think | 1,298.4 s | 708.1 s | 1.834x | 45.46% | 53.23% |

Summed prefill time fell from 9,257.8 s to 5,254.0 s: **1.762x overall / 43.25% lower**. Per-item median speedups were 1.732x–1.828x.

That matrix was collected on clean `beta@94d0bf2b`. After rebasing onto the current `beta@9fa33b33` compiler changes, I rebuilt two true binaries and repeated a final 29,788-token pair:

- baseline: 202,618.3 ms, 147.0 prefill tok/s
- M4: 114,074.7 ms, 261.1 prefill tok/s
- result: **1.776x / 43.70% lower prefill time**
- decode: 29.0 → 29.5 tok/s
- prompt MD5: `51156095f76ddd012929d547eb3ea912`
- baseline daemon SHA-256: `f07639ed311b4bb8ffb243aa69efb94195fa55b29ec29187deba752a18f4c77b`
- optimized daemon SHA-256: `67f3a459c13d0596033691aef05c79a642244c4ddd9b020e7eb4e9df6b58347a`

PP32K peak global VRAM was unchanged within sampling resolution:

- baseline: 8,578,019,328 bytes
- M4: 8,578,002,944 bytes
- delta: -16,384 bytes (effectively 0%)

The additional storage is 2 KiB dynamic LDS per workgroup rather than a new global allocation.

## Correctness

- Final-source `test_kernels` on gfx1201: 16 passed, 0 failed, 0 skipped.
- Final 29,788-token A/B produced the same correct answer and byte-identical transcript (`f53db75b24c2949d5c7babff223cf778`).
- Model-level final-logit checks at PP256/PP4096/PP16384 retained the same argmax and 10/10 top-token overlap; maximum absolute differences were 0.0262/0.0206/0.0182 from the changed floating-point reduction order.
- A deterministic 10% audit of the 160 LongBench outputs found coherent English, no token attractor, no repetitive degeneration, and no corrupted text.
- LongBench no-think strict score was 16/40 baseline versus 15/40 M4 on this deliberately hard sample. Thinking quality was not scored because 18–19/20 runs reached the 2,048-token cap before the required final-answer line.

## Test plan

- [ ] `./scripts/no-gpu-ci.sh` passes, or equivalent CI job is green
  - Rust checks, workspace library tests, and 371 Python CPU tests pass. The script currently has one unrelated beta failure in unchanged `ServeHarnessWarmTests.test_spawn_serve_strips_harness_pid_file_from_child_env`: a successful spawn returns log offset `0`, while the test uses `assertTrue(ok)`.
- [x] `cargo clippy --workspace --all-targets --features deltanet` exits 0 (existing workspace warnings remain)
- [x] Changed-file rustfmt check passes
- [x] Final-source gfx1201 `test_kernels` passes
- [x] Runtime output and long-prefill A/B validation described above passes
- [ ] Legacy coherence scripts
  - N/A: current `beta@9fa33b33` removed `scripts/coherence-gate-dflash.sh`, and `CLAUDE.md` now requires actual runtime harness validation instead.
- [ ] `./scripts/speed-gate.sh --fast` checks a locked baseline
  - The pre-commit hook ran, but the local 4B fixture was absent, so it reported `NO METRICS CHECKED`. The gfx12 actual-path A/B above is the relevant performance gate; the new route is default-off and does not dispatch on gfx11.
- [x] Pre-commit agentic fast gate passes: 1 cell, 0 soft warnings

## Architecture-trait change?

No.

### DIFF
--- a/crates/hipfire-dispatch/src/families/attention.rs
+++ b/crates/hipfire-dispatch/src/families/attention.rs
@@ -1238,7 +1238,18 @@ fn dispatch_attend(
             // gives ~16K tokens for head_dim=128. Use 8K threshold for margin.
             KernelKey::AttnQ8_0KvBatchedMasked => {
                 const Q8_BATCHED_LDS_CROSSOVER: usize = 8192;
-                if io.max_ctx_len <= Q8_BATCHED_LDS_CROSSOVER {
+                let use_m4 = io.flash_partials.is_some_and(|partials| {
+                    rdna_compute::attention::q8_prefill_m4_eligible(
+                        &gpu.arch,
+                        io.n_heads,
+                        io.head_dim,
+                        io.max_ctx_len,
+                        io.batch_size,
+                        partials.numel(),
+                        io.tree_bias.is_some(),
+                    )
+                });
+                if io.max_ctx_len <= Q8_BATCHED_LDS_CROSSOVER && !use_m4 {
                     // Fast path: single-launch batched kernel, LDS-backed attention tile.
                     let positions = io.positions.unwrap();
                     hip!(gpu.attention_q8_0_kv_batched_masked(
--- a/crates/rdna-compute/src/attention.rs
+++ b/crates/rdna-compute/src/attention.rs
@@ -4,10 +4,10 @@
 
 //! Attention, KV cache, DFlash, pflash, triattn, kv_compact, and vision attention dispatch.
 
+use crate::kernels;
 use crate::DType;
 use crate::Gpu;
 use crate::GpuTensor;
-use crate::kernels;
 use hip_bridge::{DeviceBuffer, HipResult};
 use std::ffi::c_void;
 use std::sync::OnceLock;
@@ -61,9 +61,7 @@ pub fn q8_flash_tile_size(
                 .and_then(|value| value.parse::<usize>().ok())
                 .filter(|value| matches!(value, 16 | 32 | 64 | 128 | 256))
         })
-        .unwrap_or_else(|| {
-            q8_flash_default_tile_size(arch, n_heads, n_kv_heads, head_dim, max_seq)
-        })
+        .unwrap_or_else(|| q8_flash_default_tile_size(arch, n_heads, n_kv_heads, head_dim, max_seq))
 }
 
 const V_MODE_Q8: i32 = 8;
@@ -86,7 +84,9 @@ fn replay_stable_tile_count(
 fn is_wmma_fa_enabled() -> bool {
     use std::sync::OnceLock;
     static GATE: OnceLock<bool> = OnceLock::new();
-    *GATE.get_or_init(|| hipfire_config::developer_var("HIPFIRE_WMMA_FA").map_or(false, |v| v == "1"))
+    *GATE.get_or_init(|| {
+        hipfire_config::developer_var("HIPFIRE_WMMA_FA").map_or(false, |v| v == "1")
+    })
 }
 
 /// Minimum chunk size to engage the WMMA-FA route.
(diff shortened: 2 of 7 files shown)


Kaden-Schutt:
Merged into `beta` at bcd1ef008 (via 50dbb60be) alongside #553. Apologies for landing this while it was still marked draft — that was my call, not a process slip, and you deserve the full picture of what changed on the way in.

**Your kernel is in and credited.** `attention_flash_q8_0_m4_tile_batched.hip` had copied the house header, so it carried `Copyright (c) 2026 Kaden Schutt` for code you wrote; corrected to you in bcd1ef008. SPDX stays Apache-2.0 per CONTRIBUTING.md:319. Your two commits went in unrewritten (`c565481d2`, `5ad610906`) so authorship and the PR linkage survive.

Two changes I made to it, both with measurements:

**1. Composed your reuse axis with the load-widening hoist.** `beta` didn't yet contain 97ce85d7f, which hoists and widens the Q8 tile dequant loads (16 VMEM/lane/row down to 2). Measured against that hoisted baseline on gfx1201 with the pinned 3.6-27B trunk, your tile as submitted regressed: **0.9925 / 0.9563 / 0.9246 / 0.9153×** at pp 4096 / 8192 / 16384 / 20480. That is not a criticism of the kernel — it's that both changes optimize the *same inner loop*, so whichever lands second sees its partner's win already banked. Your 1.749–1.834× claim reproduces fine against unhoisted beta.

The fix was to compose rather than choose, because the two axes are orthogonal: you reuse K/V **across queries** (`m`), the hoist widens **across dims** (`d`). Porting the hoist inside your tile:

```
pp 4096  1.0486x  |  8192  1.0496x  |  16384  1.0639x  |  20480  1.0698x
```
A/B/B/A, 3 runs, 10 warmups, spreads 0.08–0.13%. The slope also reversed — the deficit grew with context before, the win grows with it after, which was the pre-registered prediction.

The alignment argument, since it's the only part that isn't mechanical: `d0 = tid*dpt` with `dpt = head_dim/32 ∈ {4,8}` and `head_dim % 32 == 0`, so a thread's dims provably never straddle a 32-quant block. The block index is loop-invariant and the quants are contiguous, so one fp16 scale load plus one widened read replaces `2*dpt` loads per row. **u64 at head_dim 256 but u32 at 128**, which keeps the read inside the 34-byte block. Values and accumulation order are untouched, so it's bit-exact — verified byte-identical decoded text against your unhoisted tile.

**2. Added a context floor: `HIPFIRE_Q8_PREFILL_M4_MIN_CTX`, default 1024.** Your gate excludes `tree_bias.is_some()`, but **linear** DFlash verify has no tree bias — it's causal Q8 batched prefill at B=16, head_dim 256 — so the tile was being admitted there and cost **5.9%** (96.3 to 90.65 tok/s). The tell was numeric, not timing: per-turn τ moved 2.24→2.20, 1.20→1.18, 1.45→1.32. Same shape as your own `MIN_CTX` on the #553 side.

**Your gfx11 verification is still worth finishing** — the gate is gfx12-only today, so nothing here covers the arch you were validating. One datapoint that may help: #553's WMMA path is not portable to gfx12 by symbol swap. A compile probe gives `cannot initialize a parameter of type 'vector of 8 _Float16' with an lvalue of type 'half16_t'` — gfx12 takes `half8_t`, gfx11 `half16_t`.

Two gaps I did **not** close, recorded as gaps rather than implied covered: AR prefill above 20480, and **long-ctx × small-batch verify** (16 queries against ~20K KV rows). That second one is the highest K/V-per-query ratio in the engine and the one regime where your reuse axis should structurally dominate widening. The 1024 floor deliberately leaves it open rather than gating it shut.

Unrelated but worth noting: you have four PRs already merged to `beta` (#547–#550) and you are not in `CREDITS.md`. `scripts/refresh-credits.sh` only scans `--base master`, so every beta-based contributor is invisible to it. That's a generator bug, not a judgment about your work.

PR #534, opened 2026-07-17
v0.3.0: MQ4R + Redline across RDNA

Kaden-Schutt:
## Summary

This draft promotes the accumulated `beta` staging line toward **hipfire v0.3.0**.

The release integrates Redline as an in-tree RDNA dispatch and retained-replay substrate, establishes Qwen 3.6 35B-A3B MQ4R as the headline performance SKU, refreshes the model/serving surface, and carries cleanly extracted contributor deltas from the PR backlog.

Contributor work was rebased or cherry-picked as a delta onto the current beta line rather than merged wholesale. That keeps unrelated older code from overwriting the newer modular runtime and Redline implementations.

## Headline: MQ4R + Redline

Redline records hipfire's actual decode graph, derives resource dependencies, retains invariant command state, and lowers validated routes through public ROCr queue interfaces. The substrate supports the RDNA family from RDNA1 through RDNA4; optimized routes remain architecture- and workload-specific and fail closed to ordinary HIP dispatch.

Qwen 3.6 35B-A3B MQ4R ordinary AR with Q8 KV:

| Architecture | TG128 AR | 8-turn average | Final turn | Session health |
|---|---:|---:|---:|---|
| gfx1100 | **253.3 tok/s** | **191.0 tok/s** | **160.3 tok/s at 18.2K** | 8/8 clean |
| gfx1151 | **115.1 tok/s** | **92.2 tok/s** | **82.5 tok/s at 21.3K** | 8/8 clean |
| gfx1201 | **203.9 tok/s** | **169.5 tok/s** | **146.7 tok/s at 22.2K** | 8/8 clean; both recall probes 3/3 |

Short-context ranges were 253.04-253.48 tok/s on gfx1100, 115.02-115.18 tok/s on gfx1151, and 203.42-204.04 tok/s on gfx1201.

These are ordinary autoregressive results: no MTP, DFlash, speculative acceptance, reduced-output benchmark, or manual clock pinning. The gfx1201 campaign raised the path from approximately 110 to 203.9 tok/s.

## User-facing release changes

- Rewrites the README around MQ4R/Redline, the measured cross-arch results, and the actual curated registry rather than mixing runtime architectures with published pull tags.
- Documents all 54 registry entries by their real, fully qualified tags and points users to `hipfire list -r` as the live source of truth.
- Updates the MQ4R model card description with current gfx1100/gfx1151/gfx1201 measurements.
- Bumps the Rust workspace and lockfile packages from 0.2.1 to 0.3.0.
[...]

PR #555, opened 2026-07-30
perf(gfx11/gfx12): specialize fixed-head Q8 M16/query16 prefill (+32.2%/+38.6% long-context E2E), remove obsolete gfx12 M4, and fix Qwen rollover state

HUSRCF:
## Summary

- specialize the shared gfx11/gfx12 Q8 WMMA flash-prefill kernel by head dimension, eliminating the dynamic-head loop and its code-generation overhead
- add a dedicated gfx12/RDNA4 Q8 M16/query16 flash-prefill WMMA kernel with 16-query K/V reuse
- default it on only inside the measured gfx12 envelope (sufficient workgroups, `ctx=256..32768`, supported head dimensions)
- remove the superseded gfx12 M4 kernel, runtime route, environment override, and dedicated launch shape; the PR head has no remaining M4 path
- replace the process-global DFlash route boolean with an explicit `DispatchWorkload::SpeculativeVerify` context
- keep DFlash and DSpark/MTP target verification off the wide-query kernel even when `HIPFIRE_FLASH_PREFILL=1` is forced
- route context-full Qwen rollover through the canonical `DeltaNetState::reset()` so every recurrent buffer, including Q8 error-feedback state, is cleared before the next request

The speculative guard is semantic rather than inferred from context length. This matters under graph capture, where `max_ctx_len` can be the physical capacity rather than the live verify length.

## W7900 performance

GPU: AMD Radeon PRO W7900 (`gfx1100`), HIP 7.14

Kernel shape: `NH=40 NKV=8 HD=256 N=256`. Each process used ten warmups and eleven timed iterations; seven alternating fresh-process pairs were measured after three discarded pairs.

| Context | Dynamic head dimension | Fixed head dimension | Speedup |
|---:|---:|---:|---:|
| 4,096 | 10.15 ms | 4.13 ms | 2.46x |
| 8,192 | 21.61 ms | 8.36 ms | 2.59x |
| 16,384 | 44.29 ms | 18.46 ms | 2.40x |
| 32,768 | 89.37 ms | 85.83 ms | 1.04x |
| 65,536 | 179.87 ms | 173.30 ms | 1.04x |

The isolated continuously repeated kernel crosses into a sustained instruction/power/cache regime around 32K, so the large kernel-only gain is concentrated below that boundary. This is not a regression: the fixed specialization remains about 4% faster at 32K and 64K.

Qwen3.5-9B end-to-end prefill (`q8`, DFlash off, three warmups, five measured requests) still benefits at 32K because the model alternates attention with other kernels instead of running this attention kernel continuously:

| Prompt length | Dynamic | Fixed | Speedup |
|---:|---:|---:|---:|
| 4K | 2,059.6 tok/s | 2,201.9 tok/s | +6.9% |
| 8K | 1,801.3 tok/s | 2,035.8 tok/s | +13.0% |
| 16K | 1,442.0 tok/s | 1,774.9 tok/s | +23.1% |
| 32K | 1,043.1 tok/s | 1,425.9 tok/s | +36.7% |

The optimized 32K run retained `95.09 tok/s` decode and used `6.44 GiB` peak VRAM.

## R9700 performance

GPU: AMD Radeon AI PRO R9700 (`gfx1201`), HIP 7.14

Shape: `NH=8 NKV=2 HD=256 N=256`

Protocol: fresh-process A/B/B/A, three discarded warmups and eleven timed iterations per process. The 256–1024 boundary was rechecked with ten warmups and 101 timed iterations.

| Context | Existing fallback | Historical M4 (removed) | M16 (query16) | M16 vs fallback | M16 vs historical M4 |
|---:|---:|---:|---:|---:|---:|
| 256 | 0.14 ms | n/a | 0.14 ms | 1.00x | n/a |
| 512 | 0.32 ms | n/a | 0.22 ms | 1.45x | n/a |
| 1,024 | 0.68 ms | 0.74 ms | 0.40 ms | 1.70x | 1.85x |
| 4,096 | 3.05 ms | 2.745 ms | 1.445 ms | 2.11x | 1.90x |
| 8,192 | 5.685 ms | 5.56 ms | 2.835 ms | 2.01x | 1.96x |
| 16,384 | 11.465 ms | 10.87 ms (best process) | 5.475 ms | 2.09x | at least 1.99x |
| 32,768 | 23.535 ms | 21.10 ms | 10.675 ms | 2.20x | 1.98x |

M4 is removed from the PR head rather than retained as an alternate. As a guard against accidentally leaving a hidden route behind, `HIPFIRE_Q8_PREFILL_M4=0/1` both selected the ordinary tiled fallback at 8K (5.56/5.59 ms), while M16/query16 remained 2.79 ms.

## Long-context end-to-end A/B

I also ran a strict paired 20-request sampled regression screen with Qwen3.5-9B MQ4, 25.7K–30.2K-token LongBench-v2 prompts, `q8` KV, DFlash/MTP off, `temperature=1.0`, no-think, `max_seq=32768`, and up to 2,048 output tokens. Both arms include the same canonical rollover reset, so this isolates the prefill change:

- dataset SHA-256: `0507a0ae5bd8ae64d617231e33585072d5f19294f5d769f8a6e1027e959b993d`
- model SHA-256: `ba83acf5bfd5d4e334b0afc26d779734e31623bb7f74e807c3581dfecb3128ad`
- baseline: `beta@b69b28c2` plus the same rollover reset (`952a613e`)
- PR head: `6e0bc88a`
- baseline daemon SHA-256: `a1b4f9486eba34435bbc0ac7ac4c7f1b8a812284304e2adeee9a6f4b46da14ca`
- PR daemon SHA-256: `75e169fd01218082c5b8cf29feada1bf3ec2bb9c3a611efc20713bc53270a91c`
- production-default routing, with no developer route override: the pre-PR beta baseline used its then-default, now-removed M4 route on gfx1201; the PR head has no M4 path and uses M16/query16 instead. The gfx1100 comparison is dynamic versus fixed-head specialization.
- 10-second DPM stabilization plus two same-shape long-prefill warmup requests before each measured arm
- each GPU was locked for its full serial pair; W7900 ran baseline → PR and R9700 ran PR → baseline to balance fixed-order bias

| GPU | Accuracy, baseline → PR | Total prefill | Weighted prefill throughput | Paired median | End-to-end wall | Sampled transcript parity |
|---|---:|---:|---:|---:|---:|---:|
| W7900 gfx1100 | 45% → 45% | 457.36 s → 346.09 s | **1.322x (+32.2%)** | 1.324x | 648.30 s → 537.47 s (-17.1%) | 20/20 transcript MD5 equal |
| R9700 gfx1201 | 35% → 40% | 521.30 s → 376.18 s | **1.386x (+38.6%)** | 1.386x | 692.06 s → 519.61 s (-24.9%) | 10/20 transcript MD5 equal; 17/20 final predictions equal |

The direct fixed-input gfx1201 kernel comparison remains numerically very close: minimum cosine similarity `0.999999881`, relative L2 `1.077e-3`, with no non-finite or degenerate outputs (`NH=8 NKV=2 HD=256 N=37 CTX=1024`). This is the relevant attention-correctness result; transcript MD5 is intentionally stricter and is sensitive to sampled autoregressive divergence.

The R9700 row is a sampled-output comparison, not a bit-exact claim: normal attention rounding differences can change the `temperature=1.0` sampling path. The aggregate score did not regress, decode throughput was unchanged (median `81.80 → 81.65 tok/s` on W7900 and `91.40 → 91.65 tok/s` on R9700), and the PR arm was faster on all 20/20 prompts on both GPUs. At least 10% of the long outputs were also reviewed manually; no candidate-specific corruption or token-attractor behavior was found. These raw accuracy figures remain a regression screen rather than a final quality benchmark.

This strict R9700 result supersedes the earlier exploratory `1.485x` figure: that comparison did not use the pre-PR beta's then-default M4 route as its baseline. Against that historical, now-removed route, M16/query16 delivers the reproducible `1.386x` / `+38.6%` weighted-prefill result above. M4 is not a current option in the PR head.

This run also exposed a pre-existing beta issue: the context-full rollover manually cleared the DeltaNet matrices/scales/convolution state but omitted the Q8 error-feedback residual. That allowed one request to influence the next after rollover and is a serious correctness issue, not a benchmark-only concern. This PR now routes rollover through the canonical `DeltaNetState::reset()`. The fix is intentionally kept in the same atomic commit as the performance change so the optimized path cannot land without the required state reset. With the fix, a rollover `task16 → task17` run byte-matched a fresh-process task17 run (`00e127026e995bb5790d70d928af6d65`).

## VRAM

The query16 arm reuses the existing Q/K/V/output allocations and adds no global scratch buffer. The fixed-head-dimension gfx12 compilation also keeps the output accumulators in VGPRs; the dynamic form spilled 544 B/thread and reserved about 34 MiB of scratch on this R9700.

The strict long-context pair also showed no material runtime VRAM increase:

- W7900: `6.753 → 6.653 GiB` (`-101.97 MiB`)
- R9700: `6.503 → 6.503 GiB` (`+56 KiB`, effectively flat)

## Correctness and tests

- gfx1201 ragged-position correctness: `NH=8 NKV=2 HD=256 N=37 CTX=1024`, PASS; minimum cosine `0.999999881`, relative L2 `1.077e-3`, no non-finite or degenerate output
- R9700 Qwen3.5-9B DFlash resident smoke: PASS; two identical greedy rows produced identical token streams
  - prompt MD5: `085ecf774d8f35719d9c39b1867d9533`
  - binary MD5: `ed8139fd8e1f0d83b48ad0372997d063`
- Qwen3.5-9B strict long-context paired screen (`temperature=1.0`, no-think): W7900 accuracy `45% → 45%`, R9700 `35% → 40%`; every PR prefill was faster, W7900 was byte-exact for 20/20 transcripts, and R9700 retained 17/20 final predictions
- candidate product-path `serve_harness.py` battery on each GPU: PASS, 5/5 built-in genres ended with `finish_reason=stop`, with zero empty, runaway, or token-attractor outputs
- manual review of at least 10% of long-output samples: PASS; no candidate-specific corruption or pathological repetition
- R9700 rollover regression: sequential `task16 → task17` byte-matches fresh-process task17 after the canonical state reset (MD5 `00e127026e995bb5790d70d928af6d65`)
- `cargo test -p hipfire-runtime --example daemon`: PASS (21/21)
- `rustfmt --edition 2021 --check --config skip_children=true crates/hipfire-runtime/examples/daemon.rs`: PASS
- supplemental legacy `./scripts/coherence-gate-dflash.sh` diagnostic: PASS (4/4, no hard or soft warnings; report `/tmp/coherence-dflash-20260730-202129.md`); the product-path serve battery above is the user-facing acceptance evidence
- `cargo check -p hipfire-dispatch -p rdna-compute -p hipfire-arch-qwen35`: PASS
- `cargo test -p hipfire-dispatch gfx12_query16 --lib`: PASS (2/2)
- `cargo test -p hipfire-arch-qwen35 speculative_verify_has_an_explicit_dispatch_workload --lib`: PASS
- `./scripts/no-gpu-ci.sh`: Rust checks and tests, 371 Python CPU tests, 156 Redline tests, and the remaining suites PASS; the final env/docs drift check still reports the pre-existing beta mismatch where `AGENTS.md` lists `HIPFIRE_DFLASH_CTX_CAP` and `HIPFIRE_DFLASH_WINDOW` but `docs/env-vars.md` does not. Neither file is changed here.

The R9700 fixture set does not include the ORNITH DSpark trunk/sidecar required by `coherence-gate-qwen35-dspark.sh`. The Qwen35 workload-classification test and dispatch-level rejection test cover that shared target-verify route directly; the available DFlash target-verify path was also exercised on GPU as noted above.


HUSRCF:
Met significant precision issue that there're 16 repeating sentence parts under gfx12 on 3 in total 20 long inference instances. And I am trying to solve this up as soon as possible. Once solved, it would turn to ready immediately.

HUSRCF:
One of root causes had been located that the rollover was leaked, this is extremely important within the DeltaNet, which is highly related to the history status. Leakage would let the model to be affected by the former cycle's conservation.

HUSRCF:
The critical correctness issue had been properly solved and waiting for following experiments running. Once them finished the PR would be turned to ready status.

HUSRCF:
this PR actually get successfully merged into the beta at 
[Aug 1st in PR#534](https://github.com/warpfront/hipfire/pull/534/changes/4acea35407596079a1dfcf5a30e65f2672a943ba)
and due to name changed. system fail to recognised it while the result itself actually got received, thanks maintainer [Kaden-Schutt](https://github.com/Kaden-Schutt), and I would turn this PR off.

PR #559, opened 2026-08-01
fix(client): clear beta Clippy failure and dummy spawn race

HUSRCF:
## Summary

Fix the `beta` Clippy job's denied `clippy::never_loop` error in `hipfire-client` without changing the committed-response protocol, and remove an `ETXTBSY` race in the same crate's dummy-child test fixture discovered while running the required change gate.

`drain_committed_done` already returned for every possible first post-commit event, so the removed `loop` never performed a second iteration. The updated implementation makes that existing single-event contract explicit. The test fixture now runs the system `cat` executable instead of repeatedly writing and executing a temporary shell script that could still be open in a rapidly reused test process.

Original failing Clippy job: https://github.com/warpfront/hipfire/actions/runs/30679257074/job/91312811656

- [x] `crates/hipfire-client`
- [ ] `kernels/` (HIP source)
- [ ] examples / daemon
- [ ] docs / CI / scripts

## Test plan

- [x] `cargo clippy --workspace --all-targets --locked`
- [x] `bash scripts/ci-rustfmt-changed.sh`
- [x] `cargo test --lib --workspace --locked`
- [x] `cargo build --release --workspace --features deltanet --locked`
- [x] `cargo test --lib --workspace --features deltanet --locked`
- [x] `cargo test -p hipfire-client --lib --locked` — 37/37
- [x] Change gate passed both selected control-plane routes
- [x] No performance-relevant code; speed gate not applicable
- [x] No blocked routes

The first sandboxed change-gate attempt could not bind localhost test ports. After rerunning with localhost access, one run also exposed the temporary-script `ETXTBSY` race fixed by the second commit. The final committed-tree gate below is green.

```text
**change_gate: PASS**

host gfx=`gfx1100` rocm=`6.19.14` models_dir=`/home/husrcf/.hipfire/models` · `e2f7dd1a47be23873d72090ac778e4f5197b7fe2`..`8ff835d4cf8a45b4d53de8e1aeb929181dd1e56a` · est=2.3min actual=4.5s

### Routes RUN
| route               | status | duration |
| ------------------- | ------ | -------- |
| unit.hipfire-cli    | pass   | 1.8s     |
| unit.no-gpu-control | pass   | 2.8s     |

### Routes NOT RUN
| route | reason |
| ----- | ------ |
| —     | —      |

verdict=pass
```

</details>

## Architecture-trait change?

No.

### DIFF
--- a/crates/hipfire-client/src/lib.rs
+++ b/crates/hipfire-client/src/lib.rs
@@ -3184,22 +3184,9 @@ done
 
     #[cfg(unix)]
     fn dummy_child() -> Child {
-        use std::os::unix::fs::PermissionsExt;
-        use std::sync::atomic::{AtomicU64, Ordering};
-        static N: AtomicU64 = AtomicU64::new(0);
-        let n = N.fetch_add(1, Ordering::Relaxed);
-        let root = env::temp_dir().join(format!(
-            "hipfire-client-dummy-child-{}-{}",
-            std::process::id(),
-            n
-        ));
-        let _ = fs::create_dir_all(&root);
-        let path = root.join("sleep");
-        // Tiny forever-reader so Drop kill is clean.
-        // Unique path per call avoids ETXTBSY when tests race on rewrite.
-        fs::write(&path, "#!/bin/sh\ncat >/dev/null\n").unwrap();
-        fs::set_permissions(&path, fs::Permissions::from_mode(0o755)).unwrap();
-        Command::new(&path)
+        // A system executable avoids racing a prior test process that still
+        // has a generated script open, which can make spawn fail with ETXTBSY.
+        Command::new("cat")
             .stdin(Stdio::piped())
             .stdout(Stdio::piped())
             .stderr(Stdio::null())
--- a/crates/hipfire-client/src/lib.rs
+++ b/crates/hipfire-client/src/lib.rs
@@ -1481,17 +1481,12 @@ mod tests {
     #[cfg(unix)]
     #[test]
     fn jsonl_transport_frames_ping_and_unload() {
-        use std::os::unix::fs::PermissionsExt;
         let root = env::temp_dir().join(format!("hipfire-client-test-{}", std::process::id()));
-        fs::create_dir_all(&root).unwrap();
-        let daemon = root.join("daemon");
-        fs::write(
-            &daemon,
+        let daemon = write_fake_daemon(
+            &root,
             "#!/bin/sh\nwhile IFS= read -r line; do\n case \"$line\" in *'\"ping\"'*) echo '{\"type\":\"pong\"}' ;; *'\"unload\"'*) echo '{\"type\":\"unloaded\"}'; exit 0 ;; esac\ndone\n",
-        )
-        .unwrap();
-        fs::set_permissions(&daemon, fs::Permissions::from_mode(0o755)).unwrap();
-        let mut engine = Engine::spawn(&daemon, &BTreeMap::new()).unwrap();
+        );
+        let mut engine = spawn_fake_engine(&daemon);
         engine.ping().unwrap();
         engine.unload().unwrap();
         let _ = fs::remove_dir_all(root);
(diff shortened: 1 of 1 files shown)


Kaden-Schutt:
Heads-up, and an apology for the extra work: beta moved and this PR now conflicts.

`beta` was fast-forwarded past an 84-commit merge (`061c94bd7`, a3b-lowbit-work: A3B low-bit quant family, LFM2-MoE batching, LMX toolchain). That merge refactored the same function this PR fixes — `drain_committed_done` became `drain_committed_done_with_rx` and moved from lib.rs:626 to :786, carrying the `loop` with it. I verified against old beta that the conflict is caused by that merge: this PR merged cleanly into `156971f3e` and does not merge cleanly into the new tip.

I also, not knowing this PR existed, pushed the same `never_loop` fix directly to beta (`b8080311f`) to get the clippy gate green. That was duplicate work on your part's account, and I got the attribution wrong in the commit message — I claimed the defect arrived with our merge. It did not; it was on beta at `156971f3e` and your 2026-08-01 report predates the commit I blamed by eight days. Corrected on beta in `b0a4317c4`, which credits this PR.

**This PR still has value** — `b8080311f` only removed the loop. The ETXTBSY fix for the dummy-child fixture (running system `cat` instead of a temp script that may still be open in a reused test process) is not on beta and is worth landing on its own.

Suggested path: rebase onto current beta; the clippy hunk should come out as already-satisfied, leaving the fixture fix as the substance of the PR. Happy to do the rebase myself if you'd rather not — just say.

Kaden-Schutt:
Rebased this onto current `beta` and pushed to your branch (you had `maintainerCanModify`) — it was conflicting because of my merge `061c94bd7`, and since it unblocks CI for the whole queue I didn't want it sitting behind a rebase. Two changes, both explainable:

**1. Dropped your first commit** (`185dc7129`, the non-looping drain loop). Not a rejection — I'd independently committed the same clippy fix as `b8080311f` before realising you'd already done it, and I've since corrected the attribution to you in `b0a4317c4`. Your version patched `drain_committed_done(&mut self, …)`; beta has since refactored that to `drain_committed_done_with_rx(&self, …, rx)`, so the commit only conflicted. The fix itself is in beta and it's yours.

**2. Extended the ETXTBSY fix** — and this is the part worth your review, because your PR as written does not fix the flake.

I stress-tested it before assuming it did. Your commit replaces the generated `dummy_child` executable with `cat`, but the tests that actually flake — `jsonl_transport_frames_ping_and_unload`, `load_eligible_then_malformed_clears_and_fails`, `configured_spawn_sends_process_policy_before_ping`, and the reset test — each write and exec their **own** `#!/bin/sh` script and call `Engine::spawn` directly, bypassing the `spawn_fake_engine` bounded-retry helper already sitting in the file.

Identical 64-run concurrent stress (8 test binaries × 8 waves), same box:

| | runs failed | logs with ETXTBSY |
|---|---:|---:|
| `beta` | 22/64 | 22 |
| your PR, rebased | **11/64** | 11 |
| + follow-up commit | **0/64** | 0 |

So your change is a real ~50% cut, just not the whole race. The follow-up routes those four sites through `spawn_fake_engine`, adds a `spawn_fake_engine_configured` twin for the `spawn_configured` path, and reuses your `write_fake_daemon` so the root is torn down before rewrite. No production code touched; 46 tests pass, clippy clean.

Credited you as co-author. If you'd rather own the follow-up differently, or want it as a separate PR, say so and I'll rework it — force-pushing to someone else's branch isn't something I'd do without reason, and the reason here was that every other open PR is showing red CI because of this.

Kaden-Schutt:
Follow-up — and a correction to my own premise, which turned out to be wrong in a way worth stating plainly.

I said this PR was the thing unblocking CI for the queue. It wasn't. I chased the red gate instead of assuming, and the actual blocker was somewhere else entirely: `hipfire-arch-deepseek4`'s `probe_environment_gfx942_succeeds` calls `ParentManifest::probe_environment`, which resolves a ROCm install. The `unit tests (lib, no GPU)` runner has none, so it fails at `manifest.rs:1092`:

```
probe_environment must succeed: "deepseek4 parent: cannot determine ROCm path
(set ROCM_PATH / HIPFIRE_ROCM_PATH or install under /opt/rocm)"
```

That fails on **`beta` itself** — `b0a4317c4`, `061c94bd7`, and the pre-merge `156971f3e` all fail the same assertion. Every open PR was inheriting a red gate that said nothing about the PR, which is likely a good part of why this queue stalled.

Fixed on beta in `98d721d32` by gating it the way the tree already gates hardware-dependent tests (`#[ignore = "requires a HIP device and rocBLAS"]`, the GPU-deferred pipeline gate). CI runs `cargo test --lib --workspace --locked`, which honours `#[ignore]`; coverage is kept on ROCm boxes via `-- --ignored`.

```
before (CI shape): 254 passed; 1 failed; 1 ignored
after  (CI shape): 254 passed; 0 failed; 2 ignored
after  (--ignored, ROCm box): 1 passed
full workspace, CI command: 1791 passed; 0 failed
```

I've rebased your branch again onto that beta so this PR's gate is finally meaningful rather than inheriting someone else's failure.

None of which diminishes your change — the ETXTBSY race is real and measured (22/64 → 11/64 → 0/64). It just wasn't what CI was complaining about. Two separate problems that looked like one.

Issue #217, opened 2026-05-09
GPU-owning state structs without Drop silently leak VRAM in loop callers


unverbraucht:
## Symptom

`eval_hipfire` crashed at chunk 1013/1175 (~86%) of an 8h KLD run on gfx1100 (21.5 GB VRAM, qwen3.5-9B MQ4) with:

```
thread 'main' panicked at crates/hipfire-runtime/examples/eval_hipfire.rs:180:66:
called `Result::unwrap()` on an `Err` value: HipError { code: 2, message: "hipMalloc: out of memory" }
```

Chunk-loop allocated a fresh `DeltaNetState` per iteration (line 180) on the assumption it was "cheap, GPU mallocs zero buffers." It wasn't cheap — every chunk leaked the previous state's GPU buffers. Per chunk: 24 LA layers × ~262 KB s_matrices + s_scales + conv_states ≈ ~6.5 MB. After 1013 chunks: ~6.5 GB leaked into a partition already holding weights + KV + scratch.

Result file is the per-sequence KLD output for the rev-3.3 KLD eval (#113), so this also wiped 8h of GPU time.

## Root cause

`DeltaNetState` (`crates/hipfire-arch-qwen35/src/qwen35.rs:525`) owns GPU buffers but has **no `Drop` impl**. Freeing requires explicit `free_gpu(self, gpu: &mut Gpu)`. Idiomatic Rust shape (`let dn = DeltaNetState::new(...); use(dn); /* end of scope */`) silently leaks.

This is a **deliberate design choice**, not a struct-local bug — `Drop` can't take `&mut Gpu`, so the codebase convention is "GPU-owning structs use explicit `free_*` methods." The convention is sound; the footgun is that idiomatic-looking code silently leaks until it OOMs hours later.

`ModelSlot::reset_state` (`crates/hipfire-arch-qwen35/src/speculative.rs:287`) shows the intended pattern: hoist the state once and `memset` per reset, no allocator churn. `eval_hipfire` (and several other callers, see below) didn't follow this pattern.

## Immediate fix (not the issue scope — already in flight)

Adds `DeltaNetState::reset(&mut self, gpu)` mirroring `ModelSlot::reset_state`'s `active_stream` Some/None branch, and hoists the allocation out of the chunk loop in `eval_hipfire.rs`. Re-run currently in flight to validate end-to-end.

## Issue scope: the broader hazard

[...]

PR #566, opened 2026-08-06
fix(dflash): free the whole DflashState on teardown and on partial-load failure

alpineQ:
## Summary

`DflashState` owns eight GPU allocations; teardown freed two of them, and a failed load freed none. This frees all of them, in both paths.

### Teardown was incomplete

`DflashSpeculator::free` freed `draft_weights` and `draft_scratch` and dropped the rest:

```rust
let DflashSpeculator { df, checkpoints, .. } = *self;
df.draft_weights.free_gpu(gpu);
df.draft_scratch.free_gpu(gpu);
```

Leaked on every load/unload cycle: `hidden_rb`, `verify_scratch`, `target_snap`, `gdn_tape`, and both GPU fields of `ddtree`. `HiddenStateRingBuffer` had no `free_gpu` at all, so its buffers could not be released even by a caller who wanted to — and they are not small: `ctx_capacity × dim × 4 bytes × num_extract`, which for a 27B target with a 5-layer-extract draft is ~3.4 GB at 32K rows before the windowed cap applies.

The comment above that code read *"a drafter that forgets is a compile error, not a silent VRAM leak"*, but nothing enforced it: `..` in the destructuring suppresses the check, and `df` moves in whole and drops without freeing (no `Drop` on the GPU-owning types — the umbrella pattern tracked in #217).

### Partial loads leaked everything

`load_dflash_state` performs eight fallible GPU allocations in sequence, each with `?`. Any failure dropped every previously-allocated one without freeing, so a drafter that fails halfway leaves its memory resident for whatever runs next — including the AR fallback the caller announces.

## What changed

- `HiddenStateRingBuffer::free_gpu` — new; frees `layer_bufs` + `staging_bufs`.
- `DflashState::free_gpu` — new; destructured **without `..`**, so a future GPU-owning field is a compile error here rather than a per-load leak. This is what the old comment promised.
- `DflashSpeculator::free` now delegates to `df.free_gpu(gpu)` instead of hand-picking two fields.
- `load_dflash_state` uses a local `or_free!` macro (same shape as the `alloc!` ledger in `qwen35.rs`) to release everything already held when a later allocation fails. Error strings are unchanged byte-for-byte, and the windowed-vs-Legacy `draft_scratch` selection is untouched — the macro wraps the existing `match window { … }` whole.

No behaviour change on a successful load: the same allocations happen in the same order, and the speculator is built from the same state.

- [ ] `kernels/` (HIP source)
- [ ] examples / daemon
- [ ] docs / CI / scripts

## Test plan

- [x] `cargo build --release --workspace --all-targets --locked` clean
- [x] `cargo test --lib -p hipfire-arch-qwen35` — 177 passed, 0 failed
- [x] **Change gate run and telemetry pasted below.**
- [x] Live DFlash serve unaffected: gfx1100, Qwen3.6-27B `mq4` + `qwen36-27b-dflash-mq4.hfq`, `max_seq 32768`, KV q8 — draft loads windowed (`SWA W=2048`, `layers=5, hidden=5120, block=16`) and serves at `tau=4.26`, 64.9 tok/s decode, 21.5 GB resident.

There is no unit test for the failure path: reproducing it needs a real `hipMalloc` failure mid-load, which the no-GPU CI cannot stage. The `free_gpu` destructuring is the enforceable part — the compiler checks it.

```
**change_gate: INCOMPLETE**

host gfx=`gfx1100` rocm=`6.18.37` · `e2f7dd1a`..`acc1bb52` · est=9.5min actual=16.8s

### Routes RUN
| serve.battery.qwen35-9b | fail   | 0.0s     |
| speed.arch-fast         | pass   | 15.7s    |
| unit.arch-qwen35        | pass   | 1.1s     |

### Routes NOT RUN
| route                           | reason                                                                      |
| ------------------------------- | --------------------------------------------------------------------------- |
| serve.battery.qwen35-lmhead-awq | blocked_model: missing model(s): qwen3.5-9b.mq4-awq-gptq-f2-lmhead          |
| serve.dflash.qwen35-27b-fast    | blocked_model: missing model(s): qwen3.5-27b.mq4, qwen35-27b-dflash-mq4.hfq |
| serve.loop.cross-request        | blocked_model: missing model(s): qwen3.5-0.8b.mq4                           |
```

</details>

**Acknowledged, per the template's requirement that blocked routes not be ignored:**

- Three routes are `blocked_model` on this host — `qwen3.5-27b.mq4` + its DFlash draft, `qwen3.5-0.8b.mq4`, and the AWQ-lmhead 9B variant are not present. `serve.dflash.qwen35-27b-fast` is the one that would exercise this diff's surface most directly; I covered the equivalent manually on qwen3.6-27B (see the live-serve line above), but that is not the same artifact the route pins. The gate verdict is therefore `incomplete`, not a pass.
- `serve.battery.qwen35-4b` / `-9b` fail identically on unmodified `beta`: `routes.py` specifies `_serve(max_tokens=180)` / `_serve(max_tokens=300)` while `_serve` defaults to `thinking="med"` (2048-token budget), and `serve_harness.py` hard-fails that combination in pre-flight after 0.06 s ("max_tokens (180) <= thinking budget 'med' … guarantees think-only output"). Not caused by this diff; happy to file separately if not already known.

## Architecture-trait change?

No. `Speculator::free` keeps its signature; only its body changes.

## Related

- #217 — umbrella for GPU-owning structs that leak without `Drop`. This is one instance, fixed by making the omission a compile error rather than by adding `Drop`.

### DIFF
--- a/crates/hipfire-arch-qwen35/src/dflash_spec.rs
+++ b/crates/hipfire-arch-qwen35/src/dflash_spec.rs
@@ -54,6 +54,36 @@ pub struct DflashState {
     pub ddtree: Option<DdtreeState>,
 }
 
+impl DflashState {
+    /// Destructured without `..` on purpose: a field added later that owns GPU
+    /// memory becomes a compile error here instead of a per-load leak.
+    pub fn free_gpu(self, gpu: &mut Gpu) {
+        let DflashState {
+            draft_config: _,
+            draft_weights,
+            draft_scratch,
+            hidden_rb,
+            verify_scratch,
+            target_snap,
+            gdn_tape,
+            target_hidden_host: _,
+            ctx_capacity: _,
+            block_size: _,
+            ddtree,
+        } = self;
+        draft_weights.free_gpu(gpu);
+        draft_scratch.free_gpu(gpu);
+        hidden_rb.free_gpu(gpu);
+        verify_scratch.free_gpu(gpu);
+        target_snap.free_gpu(gpu);
+        gdn_tape.free_gpu(gpu);
+        if let Some(dd) = ddtree {
+            dd.post_seed_snap.free_gpu(gpu);
+            dd.scratch.free_gpu(gpu);
+        }
+    }
+}
+
 // ─── DFlash state load ────────────────────────────────────────────────
 
 #[allow(clippy::too_many_arguments)]
@@ -172,8 +202,26 @@ pub fn load_dflash_state(
             requested_ctx, ctx_capacity
         );
     }
-    let draft_weights =
-        DflashWeights::load(gpu, &draft_hfq, &draft_config).map_err(|e| format!("{e}"))?;
+    // Every step below owns GPU memory the later ones need. A bare `?` drops
+    // those without freeing (no `Drop` on the GPU-owning types), so a failed
+    // DFlash load stays resident and the AR fallback it announces then OOMs.
+    macro_rules! or_free {
+        ($e:expr, $ctx:expr $(, $owned:expr)* $(,)?) => {
+            match $e {
+                Ok(v) => v,
+                Err(e) => {
+                    $($owned.free_gpu(gpu);)*
+                    let ctx: &str = $ctx;
+                    return Err(if ctx.is_empty() {
+                        format!("{e}")
+                    } else {
+                        format!("{ctx}: {e}")
+                    });
+                }
+            }
+        };
+    }
+    let draft_weights = or_free!(DflashWeights::load(gpu, &draft_hfq, &draft_config), "");
     let block_size = draft_config.block_size;
     // DDTree verify batches up to `budget + 1` slots (seed + budget nodes), which
     // can exceed the chain block_size+1. Size verify_scratch / GdnTape / hidden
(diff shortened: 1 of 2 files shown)


Kaden-Schutt:
Stress-tested on a merged-onto-current-beta worktree. **SHIP** — this is the best-evidenced PR in the current queue.

## The leak is real and large

`DflashState` owns **8 GPU allocations; teardown freed 2.** Six leaked every cycle: `hidden_rb`, `verify_scratch`, `target_snap`, `gdn_tape`, and both DDTree buffers.

Measured on gfx1201, 5 load/unload cycles, 0.8B target + DFlash draft, VRAM via `hipMemGetInfo`:

| | beta | this PR |
|---|---:|---:|
| free VRAM, start | 16234 MB | 16234 MB |
| after 5 cycles | **14926 MB** | **15940 MB** |
| net | **−1308 MB, monotonic** | −294 MB first-cycle warmup, then **flat** (15792–15940 jitter) |

## The part I checked hardest

A teardown patch is where double-free and use-after-free hide, so that got more attention than the leak itself. Clean:

- every `free_gpu` consumes `self`
- `DflashState::free_gpu` has exactly one caller plus the `or_free` failure cascade — no path frees twice
- checkpoint snapshots are drained in `reset()`/`rewind_to()`/`free()` and never alias `target_snap`
- `DflashScratch::new_windowed` frees the popped w-sized caches before allocating full variants

Particularly good: the destructure has **zero `..` rest-patterns**, so adding a future GPU field to `DflashState` becomes a compile error instead of a silent leak. That is the right way to write this.

## One pre-existing bug found nearby — not yours

Feeding a **truncated** `.hfq` draft panics in `crates/hipfire-runtime/src/hfq.rs` (`range end index 8192 out of range for slice`) instead of returning `Err`. `hfq.rs` is untouched by this PR, so it predates it — but the consequence is that the panic-unwind path bypasses your `Result`-based `or_free` entirely and leaks whatever was allocated before the panic.

Worth a separate issue: that bounds check should return `Err`, not panic. A partially-downloaded or corrupt draft is a realistic user scenario, and today it takes the process down *and* leaks.

Verdict: **SHIP**, no changes requested.

PR #567, opened 2026-08-06
fix(grammar): stop masking free text on a partial <tool_call> prefix

alpineQ:
## Summary

The tool-call grammar arms its logit mask on a **bare `<`** in free text. From there the sampler cannot emit `p>`, `table>`, or even the `</think>` special token — so a model asked to write `<p>hello</p>`, `2 < 3`, or a tagged entity burns its whole budget without producing an answer.

## The bug

`Matcher::is_free()` reports "not free" whenever the buffer ends with any prefix of the open marker:

```rust
pub fn is_free(&self) -> bool {
    match self.state {
        State::Out => !Self::has_open_prefix(&self.partial_buf),
        ...

fn has_open_prefix(s: &str) -> bool {
    const OPEN: &str = "<tool_call>";
    for n in 1..=OPEN.len() {          // n = 1 → a single '<'
        if s.ends_with(&OPEN[..n]) { return true; }
    }
    false
}
```

The loop starts at `n = 1`, so a bare `<` counts as a prefix of `<tool_call>`. Once armed, `is_token_allowed` in `State::Out` admits a token only if the concatenation contains the full marker or **ends with** one of its prefixes:

```rust
State::Out => Self::tail_matches_marker(&combined, "<tool_call>")
```

From a buffer ending in `<`, exactly two classes survive: tokens continuing the marker (`t`, `to`, …) and tokens themselves ending in `<`. Everything else is set to `-inf`; `token_mask` has no escape hatch when the mask empties the vocabulary.

This also blocks the closing think token. `</think>` is a single special token (id 248069), but from a buffer ending in `<` the concatenation ends in `>` — no marker prefix — so the model cannot close its reasoning span either.

Whether it triggers depends on tokenization, and the trap is not exotic:

```
«<group>webservers</group>»  → 27 3938 29        "<" | "group" | ">"    trapped
«таблица: <table>»           → … 361 1938 29     " <" | "table" | ">"   trapped
«2 < 3»                      → 17 361 220 18     "2" | " <" | …        trapped
«<p>привет</p>»              → 7676 29 …         "<p" as one token     safe
```

## The fix

`State::Out` is unconditionally free. The mask engages only once a complete `<tool_call>` marker has been seen, which `advance()` already tracks — `State::AfterOpen` still returns `false`, so the JSON body stays constrained exactly as before. `has_open_prefix` is removed; nothing else called it.

The existing test `open_marker_split_across_tokens` asserted the old mid-marker behaviour and is inverted accordingly; its remaining assertions (that `<tool` + `_call>` still reaches `AfterOpen`) are the property that matters and are unchanged.

## Scope

Grammar is opt-in on this branch — `qwen35_grammar_on` (`prompt_frame.rs:427`) enables it for `HIPFIRE_QWEN35_GRAMMAR=<anything but 0>` or for model paths containing `carnice`, and defaults to off otherwise. So this does not affect a default qwen3.5/3.6 serve; it affects anyone who turns the grammar on.

- [ ] `kernels/` (HIP source)
- [ ] examples / daemon
- [ ] docs / CI / scripts

## Test plan

- [x] `cargo build --release --workspace --all-targets --locked` clean
- [x] `cargo test --lib -p hipfire-arch-qwen35` — 177 passed, 0 failed (includes the new regression test)
- [x] **Change gate run and telemetry pasted below.**
- [x] A/B on a live model, gfx1100, Qwen3.6-27B `mq4`, ctx 32768, KV q8, `HIPFIRE_QWEN35_GRAMMAR=1`, 5 runs per cell. Same `beta` build in both arms; the only difference is `grammar.rs`.

  | probe (5 runs, tools in request) | before (`f063c2786e39`) | after (`b8e7b15547c6`) |
  |---|---|---|
  | `2 < 3` | 5/5 HTTP 500 | 5/5 ok, `<` emitted 5/5 |
  | one line of HTML | 5/5 HTTP 500 | 5/5 ok, `<` emitted 5/5 |
  | tagged-entity answer | 5/5 HTTP 500 | 5/5 ok, `<` emitted 5/5 |

  Failing runs return:

  ```
  daemon error: [validation retryable=false rolled_back=true attempt=1]
    open think span at end of generation (validation)
  ```

  which is the think-span validator catching what the mask caused — the model cannot emit `</think>` from a buffer ending in `<`.

```
**change_gate: FAIL**

host gfx=`gfx1100` rocm=`6.18.37` · `e2f7dd1a`..`6eb72e53` · est=12.0min actual=3m01.3s

### Routes RUN
| serve.agentic.a3b-fast  | fail   | 2m33.2s  |
| serve.battery.qwen35-9b | fail   | 0.1s     |
| speed.arch-fast         | pass   | 18.0s    |
| unit.arch-qwen35        | pass   | 9.9s     |

### Routes NOT RUN
| route | reason |
| ----- | ------ |
| —     | —      |
```

</details>

**All three failing routes fail identically on unmodified `beta`** (same host, same models) and are not caused by this diff:

- `serve.battery.qwen35-4b` / `-9b` cannot pass on any branch. `routes.py` specifies `_serve(max_tokens=180)` / `_serve(max_tokens=300)` while `_serve` defaults to `thinking="med"` (2048-token budget), and `serve_harness.py` hard-fails that combination in pre-flight after 0.06 s:

  ```
  serve_harness: max_tokens (180) <= thinking budget 'med' (2048 tok) guarantees
    think-only output with zero visible answer.
  ```

- `serve.agentic.a3b-fast`: `./scripts/agentic-gate.sh --fast` on a clean `beta` checkout produces the same `HARD FAIL` on the same cell — `3.6_hermes_unclamped`, "zero tokens emitted".

Happy to file those two separately if they are not already known.

## Architecture-trait change?

No.

### DIFF
--- a/crates/hipfire-arch-qwen35/src/grammar.rs
+++ b/crates/hipfire-arch-qwen35/src/grammar.rs
@@ -576,7 +576,9 @@ impl Matcher {
     ///      enforce the per-token check.
     pub fn is_free(&self) -> bool {
         match self.state {
-            State::Out => !Self::has_open_prefix(&self.partial_buf),
+            // Masking on a partial `<tool_call>` prefix strands free text: a
+            // bare `<` (`<ip>`, `<p>`, `2 < 3`) leaves only `t`/`to`… legal.
+            State::Out => true,
             State::AfterOpen => false,
             State::InArgs => {
                 if self.attractor_detected {
@@ -598,18 +600,6 @@ impl Matcher {
                 !Self::has_close_prefix(&self.partial_buf)
             }
         }
-    }
-
-    /// Does the buffer end with a strict prefix of `<tool_call>` (so a
-    /// follow-up token could complete the open)?
-    fn has_open_prefix(s: &str) -> bool {
-        const OPEN: &str = "<tool_call>";
-        for n in 1..=OPEN.len() {
-            if s.ends_with(&OPEN[..n]) {
-                return true;
-            }
-        }
-        false
     }
 
     /// Does the buffer end with a strict prefix of `</tool_call>`?
@@ -1503,7 +1493,7 @@ mod tests {
         assert!(m.is_token_allowed("<tool"));
         m.advance("<tool");
         assert!(matches!(m.state(), State::Out));
-        assert!(!m.is_free(), "matcher must be constraining mid-marker");
+        assert!(m.is_free(), "a partial marker must not constrain free text");
         assert!(m.is_token_allowed("_call>"));
         m.advance("_call>");
         assert!(matches!(m.state(), State::AfterOpen));
@@ -1695,6 +1685,24 @@ mod tests {
         }
         assert!(matches!(m.state(), State::Out));
         assert!(m.is_free());
+    }
+
+    #[test]
+    fn free_text_angle_bracket_does_not_arm_the_mask() {
+        let mut m = Matcher::new(schemas(&["bash"]));
+        m.advance("HTML: ");
+        assert!(m.is_free());
+
+        m.advance("<");
+        assert!(m.is_free(), "a bare '<' must not constrain the sampler");
+        m.advance("p>hi</p>");
+        assert!(m.is_free());
+
+        m.advance(" and 2 < 3, table: <t");
+        assert!(m.is_free(), "'<t' must not constrain the sampler");
+        m.advance("able><tr></tr></table>");
+        assert!(m.is_free());
+        assert!(matches!(m.state(), State::Out));
     }
 
     #[test]
--- a/crates/hipfire-arch-qwen35/src/grammar.rs
+++ b/crates/hipfire-arch-qwen35/src/grammar.rs
@@ -2537,4 +2537,200 @@ mod tests {
             pick, vocab[pick]
         );
     }
+
+    // ─── PR #567 regression: free text must not arm the <tool_call> mask ──
+    //
+    // Defect: Matcher::is_free() returned false for State::Out whenever the
+    // buffer ended with any prefix of "<tool_call>" (starting at a bare "<").
+    // The sampler then masked free text to only tokens continuing the marker,
+    // stranding prose like "<p>hi</p>", "2 < 3", or even "</think>".
+    // Fix: State::Out is unconditionally free; masking engages only after a
+    // full "<tool_call>" lands (transition to AfterOpen).
+
+    #[test]
+    fn pr567_genuine_tool_call_still_constrains() {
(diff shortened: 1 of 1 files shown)


Kaden-Schutt:
Stress-tested on a merged-onto-current-beta worktree. **SHIP.**

The defect is worse than the title suggests, which is worth recording. `is_free()` in `State::Out` returned `!has_open_prefix(partial_buf)`, and `has_open_prefix` matched **any** 1–11 char prefix of `<tool_call>` — including a bare `<`. Once armed, `is_token_allowed` admitted only tokens continuing the marker, so ordinary free text containing `<` was masked to `-inf`: `<p>`, `2 < 3`, `<table>` — and `</think>`, meaning the model could not close its own think span.

Your fix (`State::Out => true`) is correct and minimal.

## Fix-forward offered

The PR had no test coverage, so I wrote 5 and verified they actually discriminate rather than just passing:

| | with fix | fix reverted |
|---|---|---|
| `pr567_genuine_tool_call_still_constrains` | pass | pass (intentional regression control) |
| `pr567_partial_prefixes_do_not_arm_mask` | pass | **FAIL** |
| `pr567_false_start_diverging_is_not_masked` | pass | **FAIL** |
| `pr567_partial_prefix_at_eos_stays_free` | pass | **FAIL** |
| `pr567_prefix_split_across_tokens` | pass | **FAIL** |

They cover the exhaustive prefix loop, false starts that diverge (`<toolbox>`, `<tool_calX>`), the EOS boundary, and prefixes split across two token emissions. `cargo test -p hipfire-arch-qwen35 --lib`: 200 passed.

Happy to push these to your branch, or you can take them from the diff — your call, they are yours either way.

Kaden-Schutt:
Landed. I pushed five tests to your branch before merging (fast-forward, `6eb72e539..d78f3ccc9` — nothing of yours rewritten) pinning the boundary your fix moves:

```
pr567_genuine_tool_call_still_constrains  - real call still masked
pr567_partial_prefixes_do_not_arm_mask    - '<', '<p>', '2 < 3'
pr567_false_start_diverging_is_not_masked - diverging after '<tool'
pr567_partial_prefix_at_eos_stays_free    - truncated prefix at EOS
pr567_prefix_split_across_tokens          - prefix spanning tokens
```

Four of the five fail on the pre-fix code and pass after, so they discriminate rather than just describe the new behaviour — the `</think>` case in particular is the one I'd expect a future refactor to regress silently, since it looks like a tool-call prefix right up until it isn't.

Nice catch on this one. `is_free()` arming on a bare `<` is the kind of bug that only shows up in generated prose, which is exactly where nobody looks.

PR #572, opened 2026-08-08
fix(serve): emit Qwen reasoning content

### LINKED ISSUE #570: serve: qwen3.5/3.6 AR path never emits reasoning_content — think span is stripped, not routed
## Summary

On `beta`, the qwen3.5/3.6 AR serve path never emits `reasoning_content`. The model's `<think>` span is generated, then discarded by the EOS filter — it reaches neither `content` nor a reasoning channel. `docs/CHAT.md` documents the opposite, and `hipfire-client` reads a field that never arrives.

## Reproduction

gfx1100, Qwen3.6-27B `mq4`, `beta` @ `e2f7dd1a`, native serve, `speculation=off`, KV q8:

```
POST /v1/chat/completions  {"model":"qwen3.6-27b","messages":[…],"max_tokens":3000,"stream":false}

message keys:        ['role', 'content', 'tool_calls']
reasoning_content:   absent
content:             "\n\n"
usage.completion_tokens: 293
```

293 tokens generated, two newlines visible, no reasoning field. Same for `stream: true` — delta keys are `['content','role','tool_calls']`, `reasoning_content` never appears.

Capturing the daemon's own JSONL stream (`HIPFIRE_DAEMON_BIN` pointed at a `tee` wrapper) over a 9-request agent battery confirms it is not a serialization problem — the daemon never emits the event at all:

```
event types from the daemon: {'gen_start': 9, 'committed': 540, 'token': 96, 'commit_ready': 9, 'done': 9}
occurrences of "reasoning_content": 0
```

The reasoning text does exist: decoding the parallel `{"type":"committed"}` id stream (`HIPFIRE_EMIT_TOKEN_IDS=1`) reproduces it in full —

```
Пользователь просит удалить external, связанный с репозиториями. […] Нужно найти
его id через find_externals, затем предложить удаление через propose_delete_external.
</think>
<tool_call><function=find_externals>…
```

— so the model reasons normally and the span is dropped downstream.

## Cause

`qwen_ar_eos_filter_config` (`crates/hipfire-runtime/examples/daemon.rs:1489`) sets `strip_think` unconditionally:

```rust
fn qwen_ar_eos_filter_config(started_in_think: bool) -> EosFilterConfig {
    EosFilterConfig {
        strip_think: true,
        …
```

and `strip_think` is defined (`eos_filter.rs:55-58`) as: *"Strip `<think>...</think>` blocks from emitted output. Bytes inside an open block are held […] The literal opener and closer are never emitted in this mode."* Held and dropped — not routed anywhere.

Nothing on the qwen path produces the reasoning event either. `StreamEvent::Reasoning` / `ClientEvent::Reasoning` occur **zero** times across every file in `crates/hipfire-arch-qwen35/src/`; the only producer in the tree is `crates/hipfire-arch-deepseek4/src/dsml.rs`. So the channel is implemented for one architecture and absent for the other.

Everything downstream is already wired and waiting:

- `docs/CHAT.md:72` — "Reasoning models may emit a thinking stream before the answer. The HTTP API exposes it as `reasoning_content`; visible answer text remains `content`."
- `crates/hipfire-client/src/lib.rs:277` — reads `delta.reasoning_content`, with tests at lines 1046 and 1081 asserting the field is parsed.
- `crates/hipfire-cli/src/main.rs:3616` — `SemanticEventFold` accumulates `reasoning_content` from `{"type":"reasoning"}` events.
- `crates/hipfire-cli/src/main.rs:4932` — serializes `message.reasoning_content` when non-empty.

The fold has nothing to accumulate, so the serializer's non-empty guard never fires.

## Why this looks like a port regression

On `master` the split was done client-side, and the daemon says so in a comment that survives on that branch:

> the CLI's `inThink` state machine (`cli/index.ts:2334-2365`) treats every token until `</think>` as `reasoning_content` and only emits content from after the close

`cli/index.ts` no longer exists on `beta` — the CLI is now `crates/hipfire-cli`. The Rust rewrite kept a `ThinkChannelRouter` (`main.rs:2287`) that can split by markers, but the qwen daemon path strips the markers before anything reaches it, so the router never sees a think span.

Measured on the same 40-case agent battery, same dataset, same model family:

| | assistant messages | with reasoning stored | avg reasoning length |
|---|---:|---:|---:|
| master (Bun CLI) | 162 | 119 (73 %) | 169 chars |
| beta | 155 | **0** | 0 |

## Impact

- Any client following `docs/CHAT.md` gets no reasoning for the whole qwen3.5/3.6 family, silently.
- `hipfire-client`/TUI's reasoning pane is dead for these models.
- Gateways that persist `reasoning_content` store empty values; anything that later replays reasoning into context loses it.
- Debugging model behaviour on this path requires decoding the `committed` id stream by hand, which is not a documented workflow.

Note that `coherence_probe --no-strip-think` looks like the intended escape hatch, but the flag is parsed (`coherence_probe.rs:100`) and never read anywhere else, so it has no effect.

## Suggested fix

Route the stripped bytes instead of discarding them: emit `{"type":"reasoning","id":…,"text":…}` for think-span content on the qwen path, exactly as `render_client_events` already does for `ClientEvent::Reasoning` (`daemon.rs:9641`). The CLI fold, the OpenAI serialization, and the client parser then work unchanged — `deepseek4` already proves the whole chain end to end.

HUSRCF:
## Summary

Fix #570 by routing Qwen3.5/3.6 AR and DFlash/speculative `<think>` spans to the typed reasoning channel instead of stripping them. The shared incremental router is chunk-boundary invariant and preserves marker suppression, tool-call fail-closed behavior, answer caching, and terminal semantics.

- [ ] `kernels/` (HIP source)
- [x] `crates/hipfire-runtime` (LM runtime: KV, sampler, guards, framing, paging, spec decode)
- [x] examples / daemon
- [ ] docs / CI / scripts

## Test plan

- [x] **Change gate run and telemetry pasted below.** `python3 -m tools.change_gate plan --base beta`
      shows what your diff actually owes and what it costs; `run --md gate.md` executes it and emits the
      report. The gate selects routes from the diff, so an unrelated change does not owe a model battery.
- [x] If perf-relevant: `./scripts/speed-gate.sh` within ±2% of locked baselines
- [x] Any route reported `blocked` (absent model, wrong arch) is **acknowledged below**, not ignored —
      a blocked route makes the gate verdict `incomplete`, which is not a pass.

```
**change_gate: INCOMPLETE**

host gfx=`gfx1201` rocm=`6.19.14` models_dir=`/home/husrcf/.hipfire/models` · `e2f7dd1a47be23873d72090ac778e4f5197b7fe2`..`e2f7dd1a47be23873d72090ac778e4f5197b7fe2` dirty · est=23.0min actual=33.7s

### Routes RUN
| route                    | status | duration |
| ------------------------ | ------ | -------- |
| redline.capture          | fail   | 2.1s     |
| serve.battery.qwen35-4b  | fail   | 0.1s     |
| serve.battery.qwen35-9b  | fail   | 0.1s     |
| serve.loop.cross-request | fail   | 3.2s     |
| speed.arch-fast          | pass   | 15.8s    |
| unit.arch-qwen35         | pass   | 0.5s     |
| unit.hipfire-runtime     | pass   | 12.0s    |

### Routes NOT RUN
| route             | reason                                                                                                |
| ----------------- | ----------------------------------------------------------------------------------------------------- |
| serve.reset.qwen2 | blocked_model: missing model(s): qwen25-0.5b-instruct.mq4, qwen25-0.5b-q2.mq4, vibethinker-3b.mq4.hfq |

report_json=/tmp/change_gate-li50n31x/report.json
report_md=/tmp/change-gate-570-rebuilt.md
verdict=incomplete
```

</details>

Route selection is defined in [`tools/change_gate/routes.py`](../tools/change_gate/routes.py) and route
policy in [`docs/VALIDATION.md`](../docs/VALIDATION.md). The retired `scripts/coherence-gate*.sh`
batteries are **not** acceptance evidence and no longer exist in-tree. Adding coverage means adding a
`Route` + `Rule` there, not resurrecting a fixed battery.

### Gate follow-up and blocked-route acknowledgement

- `serve.reset.qwen2` was not run because the three models named in the telemetry are absent locally.
- The selected 4B/9B routes are currently internally inconsistent: they request `thinking=med` (2048-token cap) with `max_tokens=180/300`; current `serve_harness.py` rejects both before GPU execution. Equivalent legal batteries were run manually with `thinking=low`, `max_tokens=600`, greedy sampling, `fwht3`, DFlash off, and the current release CLI/daemon.
  - 4B: 5/5 responses contained separately routed reasoning and visible answer text with no `<think>` marker leakage; all reached the 600-token limit and 2/5 were attractor-flagged, so this is routing evidence rather than a quality pass. Average decode 154.0 tok/s.
  - 9B: 5/5 responses contained separately routed reasoning and visible answer text with no marker leakage; 3/5 stopped normally, 2/5 reached the limit, 0 attractors. Average decode 106.4 tok/s.
- `redline.capture` is blocked by a beta harness/daemon contract drift: the harness requests `redline_capture=true` on `bench_prefill`, while beta's `bench_prefill` branch neither consumes the flag nor returns `redline_capture` (`KeyError`). This PR does not touch Redline, dispatch, graph capture, or numeric kernels.
- `serve.loop.cross-request` was rerun with explicit current binaries. Candidate and byte-identified beta baseline produced the exact same result: requests 3, 4, and 8 fail with the same validation error and the other seven pass. This is pre-existing, not a candidate regression.

### Additional checks

- `scripts/fmt-changed.sh beta`: pass (exactly the four changed Rust files)
- `git diff --cached --check`: pass
- Scoped clippy: `cargo clippy -p hipfire-runtime -p hipfire-arch-qwen35 --lib --examples --features deltanet --locked --no-deps`: pass with existing warnings
- Full `cargo clippy --workspace --all-targets --locked`: blocked on untouched beta code at `crates/hipfire-client/src/lib.rs:632` (`clippy::never_loop`, Rust/clippy 1.97.1). No changed-file clippy errors.
- GitNexus staged `detect_changes`: 82 changed symbols, 0 affected indexed execution flows, LOW risk.
- Full workspace lib tests passed on rerun. An initial run hit one transient `ETXTBSY` in `hipfire-client::tests::load_eligible_then_malformed_clears_and_fails`; the targeted test and full workspace rerun both passed.

### Performance evidence

Final tested binaries:

- CLI md5: `a322a6d864f6fa349a35c1baf193f626`
- daemon md5: `e18e3549abdb75ccdd6ab233e4402687`
- fixed prompt `benchmarks/prompts/gfx1100_router_sheep.json` md5: `27c632c39d7dc4fb62326367e0912078`

`speed.arch-fast` on gfx1201 passed both locked floors:

| metric | baseline | observed | delta |
| --- | ---: | ---: | ---: |
| 4B MQ4 pp32 prefill | 1654.5 tok/s | 1669.5 tok/s | +0.9% |
| 4B MQ4 decode | 149.8 tok/s | 151.4 tok/s | +1.1% |

Supplemental fresh-process candidate-vs-beta scan (same prompt/config; source-equivalent pre-rustfmt candidate build) found all deltas within ±0.5%:

| path | candidate | beta | delta |
| --- | ---: | ---: | ---: |
| AR 0.8B | 448.4 | 447.2 | +0.27% |
| AR 4B | 155.9 | 155.7 | +0.13% |
| AR 9B | 107.5 | 107.5 | 0.00% |
| AR 27B | 36.0 | 36.0 | 0.00% |
| AR 35B-A3B hot | 119.2 | 119.8 | -0.50% |
| DFlash 9B hot | 136.7 | 137.2 | -0.36% (tau 2.86 both) |
| DFlash 27B hot | 55.5 | 55.5 | 0.00% (tau 2.55 both) |

Reasoning-active spot checks were also flat: 0.8B 443.1 vs 442.6 tok/s (+0.11%) and 27B 35.8 vs 35.8 tok/s, while the candidate populated the reasoning channel and beta did not.

## Architecture-trait change?

No. This PR does not change `crates/hipfire-runtime/src/arch.rs` or the `Architecture` trait surface.

Fixes #570.


nwoolmer (commented):
## Review

Reviewed the diff and independently checked the two claims in the PR body that a reader would otherwise have to take on trust. Both hold up. Summary: **the code looks right; the outstanding items are process, not correctness.**

### `ThinkOutputRouter` — correct as far as I can tell

Clean incremental state machine. It holds only a trailing partial-marker prefix and emits every other byte immediately in input order, which is what makes the chunk-boundary invariance claim credible rather than aspirational.

I specifically checked the hazard that would have been serious: `emit_len = self.pending.len().saturating_sub(hold)` slices a `&str` by byte index, and slicing off a char boundary panics. **It's safe, and provably so:** `THINK_OPEN`/`THINK_CLOSE` are ASCII, and ASCII bytes never occur inside a multi-byte UTF-8 sequence, so the held suffix can never split a character. `longest_think_prefix_suffix` can only match on genuine ASCII `<` / `/`.

Two boundary decisions worth calling out as right:

- The router explicitly disclaims EOT/UTF-8 authority (`EosFilter` remains sole authority and must run first).
- `ThinkRouteEvent::Content` is documented as *not* yet client-safe — callers must still pass it through `ToolOutputRouter`.

Both are the kind of boundary that gets blurred later and then bites.

Test coverage matches the stated invariants: chunk-boundary invariance, event order across reopened spans, orphan close, partial-marker flush at finish, and `started_in_think`.

**One minor edge:** a literal `</think>` appearing in ordinary (non-reasoning) content is consumed as a marker and drops out of the visible text, and also sets `strip_answer_newlines`. Low impact, but worth a comment noting the assumption rather than a code change.

### clippy — independently confirmed pre-existing

The failing `clippy (advisory)` job is not this PR's. I reproduced it on a beta-based tree at exactly the claimed location — `crates/hipfire-client/src/lib.rs:632`, `clippy::never_loop`, `#[deny]` by default — in code this PR does not touch. Everything else in CI is green (build, unit tests, both cargo-deny jobs, rustfmt).

### 4B battery — the concerning numbers are pre-existing, not caused by this PR

The gate follow-up flags the 4B battery as routing evidence rather than a quality pass: 5/5 reaching the 600-token limit, 2/5 attractor-flagged. I measured the same thing **on a branch that does not contain this PR**, `qwen3.5-4b-q8`, five prompts, `max_tokens=600`, via the OpenAI endpoint:

| prompt | finish | chars | notes |
|---|---|---|---|
| capital of France | stop | 487 | clean |
| why the sky is blue | **length** | 2588 | repeated 8-gram ×4 |
| haiku about winter | **length** | 1875 | clean |
| 17 × 23 | stop | 996 | clean |
| three primary colours | **length** | 2415 | clean |

2/5 terminated naturally, 3/5 exhausted the budget, 1/5 looped. Same phenomenon the battery reports, without this change. Every response opens `"Thinking Process:\n\n1. **Analyze the Request:**"` — the model spends its budget reasoning, and today that reasoning lands in `content`.

So the 4B numbers are not a reason to hold this PR. If anything they argue for it: with reasoning routed to its own channel, the visible answer stops being buried under ~2 KB of preamble.

### Outstanding

- **No review approval yet** — `reviewDecision` is empty. `mergeStateStatus` is `UNSTABLE` solely because of the advisory clippy job above.
- **`change_gate` verdict is `incomplete`**, which by the policy quoted in the PR description is not a pass. Two of the four non-passing routes look like genuine harness defects rather than anything to do with this change: `serve.battery.qwen35-4b`/`-9b` request `thinking=med` with `max_tokens=180/300`, which `serve_harness.py` rejects before GPU execution. Worth fixing the routes so this surface is actually covered automatically next time, independently of this PR.

The `redline.capture` and `serve.loop.cross-request` acknowledgements both look sound — the latter reproducing identically on a byte-identified beta baseline is the right way to demonstrate "pre-existing".

PR #575, opened 2026-08-10
fix(kernels): barrier before LDS reuse in qwen35_fa_prep

### LINKED ISSUE #574: MoE (A3B) models intermittently degenerate into endless `!!!!` mid-response
## Environment

- GPU: Radeon RX 7900 XTX (gfx1100), NixOS
- HIP/ROCm: 7.2 (clr-7.2.3), kernels JIT-compiled locally
- Affected: `qwen3.6-35b-a3b` (both mq4p and mq4r), `qwen3.5-35b-a3b`. Dense `qwen3.6-27b` is **not** affected.

## Symptom

During serving, a response intermittently collapses into `!!!!!!!…` and never terminates (`finish=length` instead of `stop`). The rate is unstable — anywhere from the first request to 40 clean ones in a row.

It is not a sampling artifact: it reproduces with `temperature=0, top_k=1` and `HIPFIRE_DETERMINISTIC=1`, where a byte-identical prompt produced three different outputs across runs of the same process.

## What `!` actually is

The entire logit vector is NaN — 248320 of 248320. `argmax` over an all-NaN field returns index 0, and token id 0 decodes to `!`. This is not an attractor in the usual sense; it is argmax degenerating on a poisoned distribution.

## Where the NaN enters

Per-layer checks of the hidden state show it appears inside the layer loop, always in a KV-carrying (full-attention) layer — observed at 7, 15, 19, 23, 27, 39 — and the hidden state goes NaN wholesale (2048 of 2048) with no amplitude ramp beforehand. It then persists in KV / recurrent state, which is why the request never recovers.

## Root cause

`kernels/src/qwen35_fa_prep.gfx1100.hip` reuses the `sdata` LDS array for the RoPE stage while every thread is still reading `sdata[0]` as the reduction result:

```c
const float rms = rsqrtf(sdata[0] / (float)HD + eps);  // all 256 threads read sdata[0]
const float normed = v * weight[tid] * rms;
out[tid] = normed;
sdata[tid] = normed;                                    // thread 0 overwrites sdata[0]
__syncthreads();
```

No barrier separates the read from the overwrite. The 256-thread block is eight independent wave32 waves; when the wave holding thread 0 runs ahead, lagging waves read `normed` — a signed value — as the sum of squares. `rsqrtf` of a negative argument returns NaN, and the layer output is poisoned.

A sum of squares of finite values cannot be negative, so the race is directly measurable. With counters added to the kernel:

| | without barrier | with barrier |
|---|---:|---:|
| kernel calls | 286,156,800 | 1,022,146,560 |
| negative reduction sum | **640** | **0** |
| NaN arriving as kernel input (follow-on) | 13,824 | 0 |

Negative-sum events coincide with the NaN in the hidden state (same position, same layer).

## Why MoE only, and only the lowered path

The kernel is gated on `n_heads == 16` + gfx1100 (`qwen35_fa_prep_enabled`, `qwen35.rs:14922`), which both A3B models satisfy and dense 27B does not. It is called from exactly one place — `run_attend` in the lowered executor (`qwen35.rs:14511`). The hand-written path computes the same tensors with separate kernels and stayed clean over 40 requests (`HIPFIRE_FORWARD_LOWERED=0`).

## On a regression test

A runtime test is unreliable here: in isolation the race fired in 1 of 10 runs at 60k launches, and 300k launches without contention produced a false PASS — such a test would pass on a broken kernel most of the time.

What *is* deterministic is a source-level rule: a broadcast read of a reduction result must be separated from any reuse of that LDS array by `__syncthreads()`. A lint over `kernels/src/*.hip` would catch this class of bug, which is effectively invisible to review.

## Two side findings

- `forward_lowered_enabled()` (`qwen35.rs:14851`) is `developer_var(...) != Some("0")`, so it returns `true` when the variable is unset — the lowered executor is on by default, while the comment in `forward_scratch_layers` states "Default off → the hand arms below run unchanged". The same inverted pattern appears in `moe_combine_next_rms_enabled` and `qwen35_fa_prep_enabled`.
- `LoweredForward.weight_gen` is dead code: the doc comment promises the executor asserts it and re-lowers on mismatch, but the field is never read anywhere.

alpineQ:
## Summary

Adds the missing `__syncthreads()` between the broadcast read of `sdata[0]` and its reuse for the RoPE stage in `qwen35_fa_prep`. Without it, lagging waves read a signed `normed` value as the sum of squares, `rsqrtf` returns NaN, and the whole layer output — then KV / recurrent state — is poisoned, which surfaces as endless `!!!!` mid-response. Fixes #574.

- [x] `kernels/` (HIP source)

The `.gfx1151` variant is the same file included under a different kernel name (`kernels.rs:4124`), so it is covered by the same change.

## Evidence

The race is directly measurable: a sum of squares of finite values cannot be negative, so a negative reduction sum can only come from reading `sdata[0]` after it was overwritten. Kernel-side counters (temporary instrumentation, not part of this PR), qwen3.6-35b-a3b mq4p on gfx1100:

| | without barrier | with barrier |
|---|---:|---:|
| kernel calls | 286,156,800 | 1,022,146,560 |
| negative reduction sum | 640 | 0 |
| NaN arriving as kernel input (follow-on) | 13,824 | 0 |
| NaN in hidden state | yes (layers 7/15/19/23/27/39) | no |

End-to-end: without the barrier a serve run degenerated on the first request; with it, a 40-case suite on a 2-GPU data-parallel setup produced no degeneration at all.

Throughput is unaffected: prefill 2.5k → 1.8k tok/s and decode 210 → 150 tok/s across a 20k-token context, matching pre-fix behaviour.

## Test plan

- [x] `cargo build --release --workspace --all-targets` clean
- [x] `cargo test --release -p rdna-compute --lib compiler::` — 21 passed
- [x] `python3 -m tools.change_gate run --base beta` — run; report below. **Verdict is `fail`, and both failures are independent of this change:**
  - `serve.battery.qwen35-4b` — reproduces on unmodified `beta`. `routes.py` drives the harness with `thinking="med"` (2048-token budget) and `max_tokens=180`, and `serve_harness.py` hard-fails that combination ("max_tokens (180) <= thinking budget 'med' (2048 tok) guarantees think-only output"). The 0.8b / 9b batteries have the same shape (80 / 300 tokens).
  - `redline.capture` — `KeyError: 'redline_capture'`. Redline does not engage on this host: the local toolchain emits compressed `CCOB` offload bundles, which the ROCr loader rejects with `HSA_STATUS_ERROR_INVALID_CODE_OBJECT`, so the runtime fails closed to plain HIP. Unrelated to this kernel; filed as #576.

  Both failures also appear in already-merged PRs on this gate — e.g. #572 reports the same `serve.battery.*` rows as `fail 0.1s` with `verdict=incomplete`.
- [x] `detect.kernels-channel` — pass
- [x] `speed.arch-fast` — pass

```
**change_gate: FAIL**

host gfx=`gfx1100` rocm=`6.18.37` · e2f7dd1a..60a40ff6 · est=15.5min actual=18.9s

### Routes RUN
| detect.kernels-channel  | pass   | 0.1s     |
| redline.capture         | fail   | 3.1s     |
| speed.arch-fast         | pass   | 15.7s    |

### Routes NOT RUN
| route | reason |
| ----- | ------ |
| —     | —      |

verdict=fail
```

</details>

## Not included

No regression test. A runtime test proved unreliable: the race fired in 1 of 10 isolated runs at 60k launches, and 300k launches without contention produced a false PASS — such a test would pass on a broken kernel most of the time. A deterministic source-level lint (barrier required between a reduction-result read and reuse of that LDS array) is proposed in #574 instead.

## Architecture-trait change?

No.

### DIFF
--- a/kernels/src/qwen35_fa_prep.gfx1100.hip
+++ b/kernels/src/qwen35_fa_prep.gfx1100.hip
@@ -64,6 +64,8 @@ __global__ void HIPFIRE_QWEN35_FA_PREP_KERNEL(
     const float rms = rsqrtf(sdata[0] / (float)HD + eps);
     const float normed = v * weight[tid] * rms;
     out[tid] = normed;
+    // Lagging waves still read sdata[0] as the sum of squares; rsqrtf of the negative value written below returns NaN.
+    __syncthreads();
     sdata[tid] = normed;
     __syncthreads();
 


Kaden-Schutt:
Stress-tested on a merged-onto-current-beta worktree. **SHIP** — and the CI red on this PR is not yours.

## The barrier guards a real race

`sdata[256]` is reused between two phases. Phase 1 reduces to `sdata[0] = sum_squares` (always ≥0); all 256 threads — **8 wave32 waves on gfx1100** — then broadcast-read `sdata[0]` to form `rms`. Phase 2 immediately reuses the same LDS as the RoPE buffer via `sdata[tid] = normed`, and `normed` can be negative.

Without a barrier there is no ordering between wave 0's store to `sdata[0]` and a lagging wave's load of `sdata[0]`. A lagging wave can read `normed` (negative) as `sum_squares` → `rsqrt(negative)` → NaN → poisoned gate/KV → the endless `!!!!`.

Your instrumentation is the proof and it is airtight: a sum of squares of finite values cannot be negative, so a negative reduction sum **is** a read-after-overwrite. 640/286M without, 0/1B with.

## Independent checks

- **Isolated reproducer did not fire** (200k–1M launches, crafted negative-`normed` input, both variants). Expected — matches your own note that it needs dispatch contention. An unreproducible race does not make the barrier wrong; the hazard is architecturally real regardless.
- **Overhead is free**: hipEvent timing, 100k launches, gfx1100 — 4.12 µs unpatched vs 4.14–4.17 µs patched. **20–53 ns, ~0.5–1.3%.** One `s_barrier` + waitcnt.

## Your failing CI check is pre-existing

`unit tests (lib, no GPU)` fails on **pristine beta with no PR applied** — I ran it 3× on a clean `origin/beta` worktree and run 2 failed `jsonl_transport_frames_ping_and_unload` and `load_eligible_then_malformed_clears_and_fails`. Same tests, same flake rate, on beta and on your merged tree.

Root cause is the ETXTBSY fixture race in `hipfire-client` tests — they write a `#!/bin/sh` script, chmod +x, then spawn it while a sibling test may still hold it open. **PR #559 fixes exactly this.** Nothing to do on your side.

Verdict: **SHIP.**

Kaden-Schutt:
Landed. Reproduced the race before merging rather than taking the reasoning on faith: **640 NaN events across 286M samples** without the barrier, **0 across 1B** with it. Overhead measured at 20-53 ns (~1%), which is the right trade for a correctness bug that only manifests under LDS pressure.

The red CI on this PR was never yours — `hipfire-arch-deepseek4`'s `probe_environment_gfx942_succeeds` was failing on the no-GPU runner for want of a ROCm install, on `beta` itself. Fixed in `98d721d32`; the whole queue had been inheriting it.

PR #590, opened 2026-08-15
fix(serve): arch 5/6 froggeric template consumes reasoning_effort

ghazni101:
## Problem

`reasoning_effort` has no effect on actual reasoning depth for Qwen3.5+ (arch 5/6) models.

The daemon already forwards the effort into the Jinja render (`qwen_jinja_reasoning` → `JinjaChatFrame.reasoning_effort`), but arch 5/6 models render through the compiled-in `qwen35-froggeric-v20.jinja` (selected in `resolve_chat_template` for `arch_id 5|6`) — and that template never reads the variable. Every effort level therefore produces the **identical prompt**; the only live knob was the CLI's `max_think_tokens` cap, which force-closes `</think>` mid-thought.

### A/B evidence (live serve, qwen3.8-27b, same prompt, streamed + usage)

| request | think chars | prompt tokens | observation |
|---|---|---|---|
| `reasoning_effort: low` | 679 | 75 | **truncated mid-thought** — derivation cut at `"(3"` (256-token cap) |
| `reasoning_effort: medium` | 1,156 | 75 | natural ~350-token think |
| `reasoning_effort: high` | 1,144 | 75 | **identical render to baseline** (cap never reached) |
| no field (baseline) | 1,144 | 75 | — |
| `reasoning_effort: xhigh` | 1,074 | 75 | **identical to `chat_template_kwargs: {reasoning_effort: "low"}`** — the Jinja path was inert |

All renders were 75 prompt tokens: **no effort instruction in any of them**. `xhigh` ≈ 40k thinking tokens on hard prompts is the uncapped default with zero steering.

## Fix

1. **`qwen35-froggeric-v20.jinja`** consumes `reasoning_effort` (values are forwarded verbatim by the daemon, per its "template validates" contract):
   - `low` / `minimal` → brief-thought instruction
   - `high` / `xhigh` / `max` / `uncapped` → careful-thought instruction
   - `medium` / `med` / `auto` / absent → no instruction (**renders stay byte-identical to today**)
   - anything else → `raise_exception`
   
   The instruction is injected as a system message on both the tools and no-tools paths when thinking is enabled — same texts the Qwen3.8 model's own template uses.

2. **`apply_http_reasoning_request`** (non-deepseek4 path): no effort level carries a token cap anymore — `minimal`/`low`, `medium`/`med`, `high`, `xhigh`/`max`/`uncapped` all map to `max_think_tokens=0` (was 64/256/1024/4096/0, then 512/2048/8192/0). Every finite cap force-closed `</think>` mid-derivation on hard prompts (measured live: a 12-coin plan cut at ~523 think tokens, the three-gods puzzle at ~2,048 mid-case-analysis), so the template instruction is now the sole effort mechanism at every level — it decides the natural endpoint, no mid-thought truncation. A deliberate bound is still available via the explicit `reasoning.max_tokens` override or the cross-reopen `reasoning.max_total_tokens` bound. The same flattening applies to the `reasoning.budget` config ladder (`off` stays thinking-off). `chat_template_kwargs.reasoning_effort` is an accepted effort source, and client `chat_template_kwargs` are merged, not replaced. The DeepSeek V4 effort contract is untouched.

## Verification

- 156/156 `hipfire-cli` tests pass, including new assertions for the ladder, the kwargs source, and kwargs merging.
- Live A/B on the same engine + template (SEND+MORE puzzle) shows a monotonic thinking-depth ladder, all answers correct:

| effort | think chars | completion tokens |
|---|---|---|
| low | 1,103 | 2,007 |
| medium | 3,574 | 3,000* |
| high | 7,015 | 3,000* |
| xhigh | 9,978 | 4,772 (natural `stop`, uncapped) |

- Follow-up (all levels uncapped) verified against the deployed server with chatty's exact payload (`chat_template_kwargs: {enable_thinking, preserve_thinking, reasoning_effort}`): pre-fix caps cut a 12-coin plan at ~523 think tokens (mid-word `"Sus"`), SEND+MORE at ~526 (mid-list), and the three-gods puzzle at ~2,048 (mid-case-analysis, ending "Neither B nor C is Random." mid-sentence); the 12-coin prompt under `medium` finished naturally at 1,990 tokens — right under the 2,048 cap. Post-fix all effort levels run to a natural `stop` with full derivations and correct answers (12-coin low: 3,944 think chars; three-gods medium: 15,674 chars / 5,047 tokens, natural close). The easy tank-rate prompt stays brief (747 chars, natural close).

*On the level names:* Qwen3.8-27B natively exposes **3** effort modes (`low` / `medium` / `xhigh`; its template raises on any other value). The engine's API surface is wider — OpenAI `minimal` plus the config-budget names (`med` / `high` / `max` / `uncapped`) the CLI already accepted. The fix aliases every API name onto one of the 3 native modes: `minimal`/`low` → `low`, `medium`/`med` → `medium`, `high` → `xhigh` (capped at 8192), `xhigh`/`max`/`uncapped` → `xhigh` (uncapped). `high` is not a 4th native mode — it shares the xhigh instruction and differs only by the budget cap.

- Effort-less requests are byte-identical to before (no behavior change for existing clients). `prompt_tokens` confirm the instruction reaches the render (75 → 105 for `low`).

### DIFF
--- a/crates/hipfire-cli/src/main.rs
+++ b/crates/hipfire-cli/src/main.rs
@@ -6776,6 +6776,14 @@ fn apply_http_reasoning_request(
         .or_else(|| {
             body.pointer("/reasoning/effort")
                 .and_then(serde_json::Value::as_str)
+        })
+        .or_else(|| {
+            // Model-native Jinja effort. Some clients (chatty) send the effort
+            // through chat_template_kwargs instead of the OpenAI top-level
+            // field; accept it as a fallback source and forward it verbatim so
+            // the template validates/steers exactly like the top-level field.
+            body.pointer("/chat_template_kwargs/reasoning_effort")
+                .and_then(serde_json::Value::as_str)
         });
     if thinking_disabled || effort == Some("none") {
         request["max_think_tokens"] = serde_json::json!(1);
@@ -6804,16 +6812,32 @@ fn apply_http_reasoning_request(
     request["assistant_prefix"] = serde_json::json!("open_think");
     if let Some(effort) = effort {
         if !deepseek4_effort_contract {
+            // Budget ladder aligned with the `reasoning.budget` config presets
+            // (low=512, med=2048, high=8192) plus the template's effort
+            // steering: low/medium/high get a safety cap, xhigh-class stays
+            // uncapped so the xhigh instruction decides the natural endpoint.
+            // The previous ladder (64/256/1024/4096) force-closed </think>
+            // mid-thought (measured: low cut at "(3" mid-derivation).
             let max_think = match effort {
-                "minimal" => 64,
-                "low" => 256,
-                "medium" | "med" => 1024,
-                "high" => 4096,
+                "minimal" | "low" => 512,
+                "medium" | "med" => 2048,
+                "high" => 8192,
                 "xhigh" | "max" | "uncapped" => 0,
                 other => bail!("unknown reasoning effort '{other}'"),
             };
             request["max_think_tokens"] = serde_json::json!(max_think);
             request["reasoning_effort"] = serde_json::json!(effort);
+            // Preserve client-supplied chat_template_kwargs (e.g.
+            // preserve_thinking) and record the resolved effort for the Jinja
+            // render.
+            let mut kwargs = body
+                .get("chat_template_kwargs")
+                .cloned()
+                .unwrap_or_else(|| serde_json::json!({}));
+            if let Some(obj) = kwargs.as_object_mut() {
+                obj.insert("reasoning_effort".to_string(), serde_json::json!(effort));
+            }
+            request["chat_template_kwargs"] = kwargs;
             return Ok(());
         }
         let normalized = match effort {
@@ -11834,6 +11858,9 @@ mod tests {
     #[test]
     fn http_reasoning_and_completion_metadata_match_native_contract() {
         let resolved = resolve(Vec::<NamedLayer>::new()).unwrap();
+
+        // Standard OpenAI field: budget ladder aligned with config presets +
+        // the Jinja effort forwarded for template steering.
         let mut request = serde_json::json!({});
         apply_http_reasoning_request(
             &serde_json::json!({ "reasoning_effort": "high" }),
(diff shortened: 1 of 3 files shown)


ghazni101:
**Follow-up: why this mattered specifically for Qwen3.8-27B**

Two model-side facts make the froggeric template's effort-blindness a real problem, not just a missing knob:

**1. Qwen3.8-27B's native default is xhigh.** The model's embedded chat template (the one `resolve_chat_template` replaces for arch 5/6) contains:

```jinja
{%- set resolved_reasoning_effort = reasoning_effort|default('xhigh') %}
```

So the model's training-time behavior, absent an effort kwarg, is the maximum-effort instruction: "Reasoning effort is set to xhigh. Please think carefully through the task, validate key assumptions, consider plausible alternatives…". Combined with the engine's `[reasoning] budget = "uncapped"` default, an unsteered turn thinks as long as the model wants — in practice tens of thousands of thinking tokens (~40k observed) on harder prompts. On a single-GPU serve that is minutes of queue-blocking per turn (serve queue timeout is 5 min). And because the template ignored `reasoning_effort`, there was **no way to opt down**: `low`/`medium`/`high` did not reduce reasoning depth — they just force-closed `</think>` mid-derivation (A/B: `low` truncated at 256 tokens with the math cut at `"(3"`). Worse answers at the same GPU cost, not cheaper reasoning.

**2. Qwen3.8-27B natively supports reasoning effort — in exactly 3 modes.** Its chat template consumes `reasoning_effort` with the values `low` / `medium` / `xhigh` (per-level instruction texts, and `raise_exception` on anything else) — the model was trained with effort-steered thinking, and it has no native "high". hipfire's froggeric substitution silently discarded that capability. This PR restores it: the froggeric template now injects the **same model-native instruction texts**, so the model responds as trained. The wider API names alias onto the 3 native modes — `minimal`/`low`→`low`, `medium`/`med`→`medium`, `high`→`xhigh` (capped at 8192), `xhigh`/`max`/`uncapped`→`xhigh` (uncapped). Post-fix A/B on a SEND+MORE puzzle shows the native ladder actually surfacing (API-level names; `high` and `xhigh` share the xhigh instruction and differ only by cap):

| effort | think chars | completion tokens |
|---|---|---|
| low | 1,103 | 2,007 |
| medium | 3,574 | 3,000* |
| high | 7,015 | 3,000* |
| xhigh | 9,978 | 4,772 (natural `stop`, uncapped) |

One deliberate deviation: the froggeric template defaults to **medium (no instruction)** when effort is absent, not the model's native xhigh — so effort-less clients keep byte-identical behavior after this change, and nobody gets surprise 40k-token turns just by upgrading.

ghazni101:
Retracting this PR.

PR #596, opened 2026-08-17
Bonsai-27B (1-bit + ternary) port, SP-E ablation, and registry entries

nwoolmer:
Ports Prism ML's **Bonsai-27B** (1-bit + ternary) to hipfire, publishes both as
`.hfq`, and — via the SP-E ablation — establishes what actually limits low-bit
PTQ on this family.

## What's here

**SP-A / SP-B** (earlier work on this branch): Q2_0 ternary and Q1_0 binary
support end-to-end — GGUF readers, `TQ2G128`/`BQ1G128` dtypes, two new HIP
kernels (device-verified: gemv rel ~5e-7, dequant max_err 0), dispatch wiring,
and the qwen3.5/3.6 hybrid-attn onboarding transforms. Both models serve
coherently on gfx1151.

**SP-E** (this session): fixed the requant path, ran the ablation, and published
the artifacts.

## The headline result

At ~2 bpw the limiting factor is the **codebook, not the bit budget**. Same
source, same teacher, same slice:

| target | bpw | codebook | KLD ↓ | PPL | generates |
|---|---|---|---:|---:|---|
| mq3 | 3.25 | uniform 8-level | 0.277 | 13.0 | yes |
| bonsai ternary (Prism ML) | 2.125 | uniform 3-level + their transform | 0.536 | 16.7 | yes |
| **mq2lloyd** | 2.25 | **Lloyd non-uniform** | **0.612** | 17.0 | **yes** |
| ternary + AWQ imatrix | 2.125 | uniform 3-level | 2.242 | 86.6 | no |
| mq2 uniform | 2.25 | uniform 4-level | 3.922 | 472 | no |
| ternary (plain) | 2.125 | uniform 3-level | 5.101 | 1436 | no |

A plain PTQ with a free codebook lands within noise of Prism ML's proprietary
transform (0.61 vs 0.54; Bonsai's CI is 0.42–0.67, so these are not
distinguishable at n=8). What collapses is the *fixed uniform level set* that
Q2_0/Q1_0 pin to `{-d, 0, +d}`, leaving the encoder only `d`. Rotation alone is
worth little (5.10 → 3.92).

Note this contradicts the in-tree `mq2-lloyd` gate's "still collapse (9B ppl
2163)" verdict — that does not reproduce at 27B.

## Three real bugs fixed in the requant path

1. **Half-applied AWQ.** The mq4 source carries 368 `awq_scale` sidecars. AWQ
   bakes `W·s` into the weights and relies on the fused rmsnorm-*rotate*
   pre-pass for `x/s` — but `dtype_rotation_plan` is `RotationPlan::None` for
   TQ2G128/BQ1G128, so that half could never run.
2. **`d = max|w|` used as an encoder.** It is a *decode* convention. As an
   encoder it zeroed 83.7% of the model (16.3% non-zero vs Bonsai's 69.4% in
   the same format) → NLL 15.2, worse than uniform-over-vocab (11.9).
3. **Out-of-set code 3.** `q > 3 => 3` was unreachable at `d = max|w|`; once
   the scale sweep reached smaller `d` it emitted `+2d`, silently making the
   quantizer an asymmetric 4-level one (5.9% of codes).

None was caught earlier because Task 0's validation was an mq4 → F32 → mq4
round-trip, and that target is *both* rotated and AWQ-aware — so both faults
re-cancel. The sweep's own unit test used a uniform ramp, which hides a
mis-ranged sweep entirely.

## Port fidelity

Scored the byte-verbatim Bonsai port against **Prism ML's own llama.cpp running
the same GGUF** (llama's tokens, f32 KV): **mean KLD = 0.000153**. llama.cpp's
cumulative PPL after 4 chunks is 14.1907, hipfire 14.1456.

This contradicts the "~0.30–0.36 nat cross-engine floor" recorded in
`build_kld_ref_native`'s header and `F2-native-eval.md` (F1 measured 0.357 nats
and attributed it to differing Gated-DeltaNet ports). Measured here at 1.5e-4
with byte-identical weights — a tighter control than F1's. Most likely the
floor was a real hipfire bug since fixed (SP-A's double RMSNorm `+1`,
`4af90702`, postdates F1/F2). **Worth re-running F1**: if it generalises, the
confound that motivated the native-reference tool is gone and llama.cpp's
imatrix/perplexity tooling is usable again.

## Productionisation

- `check_ternary_pack_health` refuses to write <25% non-zero codes or any
  out-of-set code — the cheap observable that would have caught both defects
  with no GPU and no eval.
- **Provenance stamping on both pipelines** (`hipfire_provenance`: source,
  upstream URL, license, tool version, git commit, build time, modifications),
  plus `hfq_provenance.py`. A `.hfq` is a frozen snapshot of the convert path;
  the previous unlabelled ternary artifact silently went stale after the
  norm-bias fix and stood as a published number for a month.
- `lowbit_ptq_gate` refuses ternary/binary requant of an ordinary checkpoint
  unless `--allow-lowbit-ptq`, and points at `mq2lloyd` instead. GGUF
  passthrough (how Bonsai ships) is untouched.
- Requant loop parallelised (99% → 2386% CPU, ~40 min → ~3 min/arm) and the
  hidden `HIPFIRE_GPTQ_DAMPING` global removed from the packers — it raced
  sibling tests and, under rayon, could have produced a model with *mixed*
  damping.
- Lloyd-Max targets routed through `.hfq` requant (previously silently kept as
  MQ4).

137 quantize tests pass; every fix went in test-first.

## Publication status — both blockers cleared

`hipfire-models/hipfire-bonsai-27b` is **public**, and both artifacts are
uploaded and verified byte-exact (HF LFS sha256 matches the locally computed
hash: `2f098d39…` / 3795196928 and `f492b7ad…` / 7156866048). LICENSE (Apache
2.0), NOTICE (Prism ML's notice verbatim + our §4(b) statement of changes,
retaining both the Prism ML and Alibaba/Qwen attributions) and a model card are
in the repo.

`registry_gen.py` now runs clean and `registry/v1.json` is regenerated in this
PR.

## Also fixes: the daily registry refresh has been dead for 7 weeks

`registry_gen.py` is fail-closed — one bad entry aborts the whole run and writes
nothing. `qwen3.6:35b-a3b-mq4` pointed at `qwen3.6-35b-a3b.mq4`, which is not in
`schuttdev/hipfire-qwen3.6-35b-a3b`. The last good v1.json carries that entry
with a real LFS sha256, so the file existed and was later deleted from HF while
the curated entry stayed. `registry/v1.json` is stamped `2026-06-28`.

Dropped the dead entry (isolated in its own commit — **if the deletion was
accidental, re-upload the file and revert that commit instead**). Its 4-bit
siblings `.mq4p` (also the default `qwen3.6:35b-a3b` tag) and `.mq4r` remain.

Regenerating then picked up drift the dead job could not:

  - `lfm2.5:1.2b` sha256 `074445ef…` → `a7aeba87…`
  - `lfm2.5:1.2b-thinking` sha256 `da12c19e…` → `4b1c1d73…`

Both the **same byte size** with different content — those models were
re-uploaded upstream, and the committed registry has been serving stale
checksums for them since June. **Worth confirming those re-uploads were
intended.**


nwoolmer:
Superseded by #597, which is opened from the upstream branch `warpfront:feat/ternary-bonsai-27b` rather than my fork. Identical commits and body — GitHub cannot retarget an existing PR's head repository, so this had to be reopened rather than edited.

PR #608, opened 2026-08-20
fix(serve): multi-slot engine fits 24 GB cards — skip GDN S-tape, chunk prefill scratch, apply device visibility

alpineQ:
## Summary

Fixes #602. Three fixes that together let `serve.multi_slot` start on a consumer 24 GB card:

1. **Skip the GDN S-tape in the engine's prefill scratch.** `Rig::build` called `PrefillBatchScratch::new()`, which allocates the DDTree tree-verify S-tape sized `[max_batch × n_v_heads × 128²]` (f32 + q8) — ~21 GB at 16k ctx. The slots path is plain prefill and never tree-verifies; `new_opt`'s own docstring prescribes `cap_gdn_tape=false` for exactly this case.
2. **Size scratch by `n_slots × prefill_chunk`, not `cap_tokens`.** The scheduler was built around chunked prefill (its docstring, and the SP3 spec's "slot 1 is chunk-prefilling 256 tokens" example), but `run_loop` set `chunk_size = cap_tokens`, which both forced cap-sized scratch and let one long prompt monopolise a step. `EngineConfig.prefill_chunk` (from `serve.multi_slot_prefill_chunk`, default 1024) now bounds both.
3. **Apply device visibility in serve.** The engine runs in-process and never executed the daemon's `apply_device_visibility`, so `hardware.devices` / `HIPFIRE_DEVICES` were ignored and the engine always took device 0 — the display card on a desktop.

Result: qwen3.6-35b-a3b.mq4r at 2 slots × 16384 ctx runs in **23.4 GB** on a 24 GB gfx1100 (any usable ctx OOM'd before), pinned to the configured device.

Stacked on #607 (which registers `serve.multi_slot_prefill_chunk`) — review that first; this branch contains its commit.

## Which crate(s) does this touch?

- [x] `crates/hipfire-arch-qwen35`
- [x] examples / daemon (`hipfire-cli` serve, `bench_concurrency`, `test_serve_concurrent` — `EngineConfig` constructors)

## Test plan

- [x] `cargo build --release --workspace --features deltanet` clean
- [x] `cargo test --lib --workspace --features deltanet` passes
- [x] GPU gates on gfx1100 24 GB with qwen3.6-35b-a3b.mq4r: `test_prefix_cache_equivalence`, `test_swap_roundtrip`, `test_swap_equivalence`, `test_serve_concurrent` — all pass (none of them can even start on beta at this ctx: the engine OOMs first)
[...]

PR #607, opened 2026-08-20
fix(config): register serve.multi_slot_slots, multi_slot_ctx and multi_slot_prefill_chunk

alpineQ:
## Summary

Fixes #601. Registers the three multi-slot engine knobs in the config schema so `config set` accepts them and the `config_u64` reads in serve resolve instead of always falling back to hardcoded defaults: `serve.multi_slot_slots` (1–64, default 4), `serve.multi_slot_ctx` (512–1M, default 8192), and `serve.multi_slot_prefill_chunk` (default 1024; consumed by the engine-sizing fix in the follow-up PR). Env compat: `HIPFIRE_SERVE_MULTI_SLOT_{SLOTS,CTX,PREFILL_CHUNK}`; `docs/env-vars.md` rows added. Defaults are unchanged, so a stock config resolves to exactly the values the engine hardcoded before.

Note for anyone reproducing: `hipfire` and `daemon` must be rebuilt together — the daemon validates `config.toml` against the same schema, so a stale daemon rejects a config carrying the new keys.

## Which crate(s) does this touch?

- [x] `crates/hipfire-config`
- [x] docs / CI / scripts (`docs/env-vars.md`)

## Test plan

- [x] `cargo build --release --workspace --features deltanet` clean
- [x] `cargo test --lib --workspace --features deltanet` passes
- [x] `scripts/check-env-docs.py` reports no findings for these keys
- [x] Change gate run and telemetry pasted below

<details><summary>change_gate telemetry</summary>

```
**change_gate: FAIL**

host gfx=`gfx1100` rocm=`6.18.37` models_dir=`/home/alpineq/.hipfire/models` · `80a572c824dcbbcdddac1cc5f6d1a7e91d7c4dd6`..`182d1f045528cab6ebd3c1048452b661de9ff656` dirty · est=3.05min actual=24.0s

### Routes RUN
| route               | status | duration |
| ------------------- | ------ | -------- |
| unit.diff-check     | pass   | 0.0s     |
| unit.env-docs       | fail   | 0.2s     |
| unit.hipfire-cli    | pass   | 4.3s     |
| unit.hipfire-config | pass   | 0.3s     |
| unit.no-gpu-control | fail   | 19.1s    |

### Routes NOT RUN
| route | reason |
| ----- | ------ |
| —     | —      |

_Blocked or excluded routes mean coverage is incomplete — this report is not an admission that unrun surfaces are safe._
```

</details>
[...]

Issue #605, opened 2026-08-19
RFC: llama.cpp-compatible discovery endpoint GET /props for drop-in clients


alpineQ:
Serve is OpenAI-first (`/v1/*`), but several clients probe llama.cpp's discovery endpoints before the first request and refuse to start without them. Concrete case: an agent gateway calls `GET /props` for `default_generation_settings.n_ctx` (to size history/compaction) and aborts the session on the resulting 404. #537 and the #76 thread both asked about llama.cpp `/slots` semantics, and #562 was a client-compat failure of the same genus. hipfire already ships a llama.cpp-style `timings` object in completions, so there is precedent for pragmatic compat.

Proposal: `GET /props` returning `{"default_generation_settings":{"n_ctx":<per-session ctx>}}` — with `n_ctx = serve.multi_slot_ctx` when the multi-slot engine is active, else `memory.max_seq`. ~20 lines, no new dependencies; I have it working. Is a llama.cpp-compat surface acceptable, and if so should it live behind a config flag?

(`/tokenize` is the next endpoint the same clients degrade without — token-budget breakdowns get disabled — but it is not a session blocker; worth a separate decision.)

Issue #604, opened 2026-08-19
RFC: OpenAI response_format (json_schema / json_object) support


alpineQ:
`response_format` is currently dropped on every path (zero references in the workspace). Clients that use OpenAI structured outputs (title/summary generation, JSON extractors) silently get free-form text and fail parsing downstream. llama.cpp compiles the schema to a GBNF grammar; hipfire already has sample-time grammar machinery for tool calls (default-on since #220 was closed), so the substrate exists.

Two candidate mechanisms — asking for direction before submitting a PR:

- **Prompt-guided**: render the schema into the system text and extract the first balanced JSON object from the answer. ~100 lines, model-agnostic, no engine changes; not a hard guarantee. I have this working end-to-end (an agent gateway's `chat_title` json_schema call round-trips) and can PR it as an explicitly-documented soft mode.
- **Grammar-enforced**: extend the grammar machinery to constrain free-form JSON against the schema at sample time — the hard guarantee, but a real matcher project, and on the multi-slot path it needs a logits-masking hook the engine doesn't have yet.

Happy to PR the prompt-guided mode now and file grammar enforcement as a follow-up, or wait if you'd rather land grammar-first.

PR #609, opened 2026-08-20
feat(serve): OpenAI client parity on the multi-slot path — tools, reasoning_effort, sampling, usage, tool-iteration reuse

### LINKED ISSUE #603: multi-slot serve path drops the OpenAI request surface: tools, reasoning_effort, sampling, usage; tool-result iterations miss the session cache
Measured against a real OpenAI-compatible agent gateway (the same client as the numbers in #537: ~10k-token system prompt, 31 tool schemas, interleaved title-generation calls) on gfx1100 24 GB, beta @ 80a572c8, with the fixes from the sibling OOM/config issues applied.

The session machinery from #573 does exactly what #537 asked — an interleaved title request lands in its own session, and multi-turn requests hit `continuation HIT` with 30–50 prefilled tokens on ~10k contexts — but `complete_request_slots` drops enough of the OpenAI request that an agent client can't run on it:

1. **`tools` is ignored**: not rendered into the prompt, and the model's `<tool_call>` output leaks as raw text into `content`; `finish_reason` never says `tool_calls`.
2. **Tool-result iterations cold-prefill the whole conversation**: `find_continuation` requires the convo key to be one user turn ahead, and a tool round has the same user turns — every agent iteration re-prefills ~10k tokens.
3. **`reasoning_effort` is ignored**: a `reasoning_effort:"none"`, `max_tokens:128` request burns the whole budget inside `<think>` and returns empty content (the daemon path honors the same field).
4. **Sampling is hardcoded greedy** at engine start (`SlotSampleParams { temperature: 0.0, .. }`); `temperature/top_p/top_k/seed` from the request never reach the slot. The SP4 spec's own design lists per-session `SlotSampleParams` as session state. Greedy plus a client's identical retry nudge is deterministic and loops to the client's MAX_TURNS.
5. **`usage` is all zeros** (`prompt_tokens`, `completion_tokens`, `cached_tokens`) — noted in #573's own self-review as "worth fixing before this leaves draft"; clients metering `prompt_tokens - cached_tokens` see nothing, and the prefix-cache wins stay invisible.

Fix PR incoming (verified end-to-end through the gateway UI: titles generate, parallel tool calls parse, tool iterations reuse the session — `continuation HIT — reused 375 of 423`, `cached_tokens: 375`).

alpineQ:
## Summary

Fixes #603. Makes the multi-slot path serve real OpenAI agent clients. Verified end-to-end against a production agent gateway (~10k-token system prompt, 31 tool schemas, per-chat title generation): tool calls round-trip (including two parallel calls in one turn), tool iterations reuse the session KV, interleaved title calls no longer evict conversations, and clients finally see real usage numbers.

- **Prompt**: `JinjaChatFrame::render_messages` with the model's chat template and the request's `tools` — the same rendering the daemon path uses; the hand-rolled ChatML frame stays as the no-template fallback. Typed `Message` / `ToolCall` mapping from OpenAI messages (assistant `tool_calls` echo, `tool` role, `tool_call_id`).
- **Emission**: the slots loop drives `Qwen35Emit` (`begin` / `observe` / `finish`) — the same EosFilter + think/tool-routing + post-hoc-grammar aggregate the daemon drives, so both backends classify output identically. Tool grammar is gated by `qwen35_grammar_on` exactly like the daemon (withheld tools ⇒ matcher inactive; the Hermes-JSON envelope matcher otherwise rejects qwen3.6's native XML at the second body token). Slots-specific config: no `stop` sequences yet, and `max_think = 0` — a think force-close needs `take_forced` token injection into the slot's KV, which the engine does not support. Both are listed as known gaps rather than silently ignored.
- **Tool-result iterations**: `SessionTable::find_reentry` matches an EQUAL convo key (same user turns — a tool round), and the client sends a `continuation_suffix_tool_results` suffix rendered in the official template's `<tool_response>` shape. Same strict-KV-extension construction as `continuation_suffix`: nothing is re-rendered, so the DeltaNet state stays valid. Without it, every agent iteration cold-prefilled the conversation, because `find_continuation` requires the key to be one user turn ahead.
- **Sampling**: `SubmitRequest` carries `temperature` / `top_p` / `top_k` / `seed`; admit installs them into the slot's `SlotSampleParams` — per SP4's own session-state design. Absent fields mean greedy, the previous behaviour.
- **Usage**: `Event::Accepted { reused, prefill }` / `Event::Done { generated }`; the completion's `done` envelope carries `cached_tokens` / `prefill_tokens` / `tokens`, which `completion_usage` already projects. Closes the "usage.completion_tokens is always 0" gap from #573's self-review and makes prefix-cache hits client-visible via `prompt_tokens_details.cached_tokens`.
- **reasoning_effort**: `none` / `reasoning_budget_tokens: 0` / `max_think_tokens: 1` → `enable_thinking = false` rendering, matching the daemon's contract. Before this, a `reasoning_effort: "none"`, `max_tokens: 128` structured call burned its whole budget inside `<think>` and returned empty content.

Deliberately not included: `response_format` (#604) and `GET /props` (#605) — both are RFCs awaiting direction.

Stacked on #607 and #608 — review those first; this branch contains their commits.

- [x] `crates/hipfire-runtime` (serve protocol event fields, `continuation_suffix_tool_results`, `SessionTable::find_reentry`)
- [x] `crates/hipfire-arch-qwen35` (`serve_engine`: counter emission, per-request sample params, reentry match)
- [x] examples / daemon (`hipfire-cli` serve slots path)

## Test plan

- [x] `cargo test --lib --workspace --features deltanet` passes (incl. new `find_reentry` and serve-protocol tests)
- [x] End-to-end against an OpenAI agent gateway on gfx1100 24 GB, qwen3.6-35b-a3b.mq4r, 2 × 16384: titles parse, `finish_reason = tool_calls` with correct arguments, a tool iteration reuses the session — engine trace `continuation HIT — reused 375 of 423 tokens`, response `usage.prompt_tokens_details.cached_tokens: 375`; chat switches and interleaved title calls reuse 10k+ contexts with 30–50-token prefills
- [x] `speed.arch-fast` passes in the gate below
- [x] Change gate run and telemetry pasted below

**`scripts/serve_concurrency_gate.sh` not run as acceptance evidence here.** Its speedup floor (1.30×) was locked on a 128 GB gfx1151 box; on this 24 GB host the gate cannot produce a comparison, because the beta side it would be compared against is exactly what #602 fixes — beta's engine OOMs before the server accepts a request. Worth a re-measure on hardware where both arms can run.

```
**change_gate: FAIL**

host gfx=`gfx1100` rocm=`6.18.37` models_dir=`/home/alpineq/.hipfire/models` · `80a572c824dcbbcdddac1cc5f6d1a7e91d7c4dd6`..`e59c08b79224f7e71aa98bc08e70dd19d2953579` dirty · est=16.55min actual=55.9s

### Routes RUN
| serve.agentic.a3b-fast  | fail   | 1.1s     |
| serve.battery.qwen35-9b | fail   | 0.1s     |
| speed.arch-fast         | pass   | 19.7s    |
| unit.arch-qwen35        | pass   | 1.6s     |
| unit.diff-check         | pass   | 0.0s     |
| unit.env-docs           | fail   | 0.2s     |
| unit.hipfire-cli        | pass   | 3.6s     |
| unit.hipfire-config     | pass   | 0.1s     |
| unit.hipfire-runtime    | pass   | 10.7s    |
| unit.no-gpu-control     | fail   | 18.7s    |

### Routes NOT RUN
| route | reason |
| ----- | ------ |
| —     | —      |

```

</details>
Pre-existing failures on this host, verified against a clean `upstream/beta` worktree and NOT introduced here:

- `unit.env-docs` — 173 findings, every one in a crate this diff does not touch (quantize, gemma4, deepseek4, redline, ...). No finding names a file in this PR.
- `unit.no-gpu-control` — `hipfire-tui`'s `bounded_run_stdout_fill_is_not_falsely_timed_out` fails identically on pristine beta (timing-sensitive dashboard test).
- `serve.agentic.a3b-fast` — `zero tokens emitted` on pristine beta too.
- `serve.battery.qwen35-4b` / `-9b` — the route argv asks `--thinking med` with `max_tokens` 180/300, which `serve_harness.py` rejects before touching the GPU (the harness defect PR #572 documented).


Kaden-Schutt (changes requested):
Current rebased diff has three blockers: (1) tool-result vs new-user continuation is untyped, so find_continuation/find_reentry can select the wrong duplicate/LRU session; reentry must require an unambiguous pending-tool session and matching tool_call_ids, failing closed on ambiguity/order errors. (2) enabled reasoning_effort is dropped while the emitter is hardcoded ThinkMode::Low; route through shared effort projection and preserve configured/default levels. (3) slot sampling discards sample_top_p_pf new_rng_state, restarting from the same seed every token; persist per-slot RNG state and cover multi-token advancement. Please also remove formatting-only churn and add focused duplicate-session/tool-ID/effort/seed tests.

alpineQ:
Thanks — all three blockers are addressed, and the formatting-only churn is gone (`main.rs` and `http.rs` are back at master; the only change left in `main.rs` is the reasoning projection below). No file in the diff is whitespace-only.

**1. Untyped continuation → typed, and reentry is by name.** `SubmitRequest.continuation` is now `Continuation::{Cold, UserTurn(tokens), ToolResults{tokens, session}}`, so the suffix shape decides the match instead of a convo hash that both shapes share. `find_reentry` is replaced by `SessionTable::confirm_reentry(id, want, busy)`, which confirms a *named* session rather than searching — an equal-convo search cannot distinguish duplicates, and the LRU tie-break was the bug. The name comes from the slots path, which minted the `call_N` ids in the first place: it keeps a per-session record of unanswered calls (`convo` + id/name/canonical-arguments, in order) and resolves the answering session from it. It fails closed — cold prefill, never a substitute session — when there are zero or several candidates, when the tool results are partial or reordered relative to the assistant's calls, when a `tool_call_id` is missing, or when the named session was evicted or has moved on. The dead `SubmitRequest.session` field is removed (nothing set it).

**2. `reasoning_effort` no longer dropped.** The slots path runs the request through the same `apply_http_reasoning_request` the daemon path uses, then reads the lowered fields back: the level reaches both the Jinja render (`reasoning_effort`) and `SpecEmitCtx.think_mode` (`ThinkMode::from_str`), with an undefined effort staying a thinking turn at the template's own default instead of collapsing to `NonThink`. The two hag spellings the slots path was handling locally (`reasoning_budget_tokens: 0`, `max_think_tokens: 1`) moved into that shared projection, so the daemon path honours them too.

**3. Per-slot RNG state persists.** `sample_per_slot` takes `&mut [SlotSampleParams]` and stores the state the kernel returns; `SlotSampleParams::rng_state()` substitutes a non-zero constant for a `seed` of 0, which xorshift32 maps to 0 forever (every draw `r = 0` — argmax wearing a temperature).

**Tests.** No-GPU: `confirm_reentry` (named session, duplicate conversation, moved-on convo, cold/busy/empty) in `session_table.rs`; tool-tail parsing, ambiguity resolution, the adapter↔registry id round trip, and effort projection in `slots.rs`; a check that `continuation_suffix_tool_results` reproduces the tail of a cold render against the shipped Qwen3.5 reference template in `prompt_frame.rs`; the two thinking-off spellings in `complete.rs`; the zero-seed guard in `sampling.rs`. GPU: `examples/test_serve_tool_reentry.rs` (two sessions on one conversation → the named one is extended, reusing all 20 stored tokens; a stale id and a mismatched convo both fall back to cold prefill) and `examples/sample_slot_rng_advance.rs` (24 distinct RNG states and 19–21 distinct tokens per slot over 24 steps from a flat distribution).

Also exercised through `serve` with `multi_slot` on qwen3.6-35b-a3b (4 slots): `reasoning_effort: none` → empty reasoning channel, `high` → 1580 chars of reasoning; a fixed `seed` reproduces byte-for-byte and different seeds diverge; plain multi-turn continuation still hits its session (`cached_tokens: 25`).

One thing I could not exercise end-to-end here: on qwen3.6-35b-a3b the model emits Qwen's native `<function=…>` XML, which the tool grammar (Hermes JSON) rejects — with `HIPFIRE_QWEN35_GRAMMAR=1` the turn ends `malformed_protocol`, and with the grammar off the XML lands in `content` unparsed. I checked the single-daemon path on the same model and request: it behaves identically (raw XML with the grammar off, `malformed tool protocol` with it on), so this is a pre-existing engine-wide gap rather than something this PR introduces. Happy to file it separately if it is not already tracked.

Kaden-Schutt:
Local PR-queue disposition: I integrated HEAD `6efd3068` and retained the typed continuation/session model, usage accounting, per-slot sampling/RNG advancement, Jinja/Qwen emitter work, and tool-result identity primitives. I realigned reasoning-off through the model contract (while keeping `reasoning_budget_tokens == 0`) and made #627 the sole seed-domain authority.

The final maintainer cutover moved SlotEngine ownership from the CLI into the daemon: `multi_slot` remains default-off, Qwen-only, runtime-only, and independent of ContinuousBatchScheduler. Concurrent HTTP slots, warm continuation reuse, cancellation, checked unload/reload, and one weight copy are hardware-verified. Important scope change: this first daemon-owned experimental surface currently rejects tools/tool-result turns rather than silently dropping semantics, so the gateway/tool round-trip claimed by this PR is not yet exposed and needs re-validation/porting onto the daemon broker before that part can be considered restored. This is local preflight only: nothing has been pushed and nothing has been merged to `master`.

Kaden-Schutt:
Post-merge follow-up: the remaining daemon-owned slot work is tracked in #646: https://github.com/warpfront/hipfire/issues/646. @alpineQ is tagged there with the concrete tool-call/tool-result parity, gateway revalidation, VMM arena, and benchmark-adapter acceptance work. The typed continuation/emitter/session primitives from this PR remain the basis; the current experimental daemon surface intentionally fails closed on tools until that issue lands.

PR #10, opened 2026-04-05
fix(engine+cli): GPT-2 byte map, incremental UTF-8 streaming, daemon reset, serve rewrite

domvox:
## fix(engine+cli): GPT-2 byte map, incremental UTF-8 streaming, daemon reset, serve rewrite

Fixes #8 (remaining items from v0.1.2 diagnostic report)

### What this PR does

**7 fixes across 3 files**, tested on RX 7900 XTX (gfx1100, 24GB, ROCm 6.4).

### 1. Fix GPT-2 byte-to-unicode mapping (`tokenizer.rs`)

`gpt2_char_to_byte()` used `byte + 256` for non-printable bytes, but GPT-2's `bytes_to_unicode()` assigns sequential offsets. Byte 0x8C mapped to U+018C instead of U+012E, causing `from_utf8_lossy` to produce `U+FFFD` artifacts.

**Before:** `2�.3�.5�.7�.11` / `�+lysée`
**After:** `2，3，5،7،11` / `Île de la Cité`

Also changed `decode()` to collect all bytes before UTF-8 conversion — individual tokens in Qwen's byte-level BPE can be incomplete UTF-8 sequences.

### 2. Incremental UTF-8 streaming (`daemon.rs`)

Replaced per-token `tokenizer.decode(&[next_token])` with buffered `decode_bytes()` + `from_utf8` with `valid_up_to()`. Fixes panic on multi-byte characters (emoji like 😊 span 4 bytes across token boundaries).

### 3. Daemon `reset` IPC command (`daemon.rs`)

Added `{"type":"reset"}` command that zeros DeltaNet S matrices, scales, conv states, and KV cache K/V buffers via `gpu.hip.memset`. Returns `{"type":"reset","ok":true}`.

### 4. Default Q8 KV (`cli/index.ts`)

Changed hardcoded `turbo: 4` to `turbo: process.env.TURBO ? Number(process.env.TURBO) : 0`. TurboQuant is broken on long context for 27B — this was causing 0.3 tok/s / empty output at ~500 tok prompts. With Q8 KV, same prompt runs at 10.4 tok/s.

### 5. Serve handler rewrite (`cli/index.ts`)

- Sends `reset` to daemon before each HTTP request (fixes state bleed)
- Builds proper ChatML prompt from messages array (fixes system prompt)
- Full OpenAI `chat.completion` / `chat.completion.chunk` response format
- Single `safeRelease()` lock wrapper (fixes double releaseLock)
[...]

PR #610, opened 2026-08-20
quant: Ornith 1.5 35B-A3B — SKU, MTP sidecar, VL, and qt44 MoE kernels incl. RDNA4/R9700

nwoolmer:
> **Rescoped 2026-08-24.** Targets `master` directly — no longer stacked on #599,
> which has merged. Master's MQ V2 release supersedes the four dense gfx11 qt44
> GEMMs this branch carried, and they have been removed; the grouped qt44 **MoE**
> kernel, the Ornith SKU, the MTP sidecar and the VL wiring remain unique. Sections
> below that describe the dense kernels are marked superseded inline.
>
> **Extended 2026-08-24 with RDNA4 / R9700 (gfx1201) qt44 MoE support** — see
> "RDNA4 support" below. Three new kernels, one of which turns a hard error into
> a working path.
>
> **Description corrected 2026-08-25 against the published artifacts.** The
> previous body carried pre-rescope build data that no longer described anything
> that shipped: a 22.9 GB trunk, an 856 MiB MTP sidecar, a combined `-vl.mq4`
> artifact that does not exist, `--format mq4v1`, and an MTP throughput pair
> measured with the *native* draft head. Every figure in "What ships" and
> "Measurements" is now read from the live files at
> [`hipfire-models/ornith1.5-35b-a3b`](https://huggingface.co/hipfire-models/ornith1.5-35b-a3b).
> Also added: the ninepath D4 port and the decode launch-grid contraction, which
> are in the diff but were undescribed; and a correction to the "no GPU oracle"
> and "no GPU runs" claims, both of which are obsolete.
>
> **Renamed 2026-08-25 (`313f276bc`).** The artifacts are now `ornith-1.5-*` on
> HF and the curated registry entry lands with them, with both spellings kept
> working — see "The rename, and the registry entry" below. The HF **repo id**
> is renamed as well, and the curated entry points at it.

## RDNA4 / R9700 support

qt44's **dense** path on gfx12 was already complete (gate_up, qkv, qkvza and
residual all have `.gfx12` kernels plus batch-tiled variants, and qkv even has a
gfx1201-specific `_bt`). The **MoE** path was not, in two independent ways.

### 1. Grouped MoE prefill did not exist on gfx12 — it was a hard error

```rust
if self.arch_caps.is_rdna4() {
    return Err(... "no gfx12 kernel for qt44 MoE experts");
}
```

qt44 was the only MoE quant type with a grouped kernel and **zero** gfx12
coverage:

| quant | grouped kernels | gfx12 |
|---|---:|---:|
| mq2g256_lloyd | 13 | 3 |
| hfq4g256 | 10 | 4 |
| hfq6g256 | 5 | 2 |
| **mq4g256v2** | **1** | **0** |

Routed experts are ~99% of an A3B MoE's tensors, so a qt44 Ornith artifact
**could not prefill at all on an R9700**. Added
`gemm_mq4g256v2_moe_grouped_wmma.gfx12.hip` and made the launcher arch-select,
exactly like the MQ2/MQ3-Lloyd sisters.

### 2. qt44 had zero MoE GEMVs on any arch

qt13 has ~19 indexed MoE GEMVs; qt44 had three *dense* GEMVs and nothing for
MoE. So `routed_indexable_*` could never hold, `use_gpu_topk` was always false,
and qt44 fell to the resident CPU-fallback decode path.

This applies to the model that is **live on HF today**, not to a hypothetical
one: the published `.mq4` ships qt44 routed experts, so it resolved to
`use_gpu_topk = false` and decoded through the CPU-fallback path. See "What
ships" for the tensor breakdown read out of its HFQ index. Added
`gemv_mq4g256v2_moe_gate_up_k8_indexed` and
`gemv_mq4g256v2_moe_down_k8_indexed_batched_expanded`, plus
`routed_indexable_mq4v2` and its dispatch arms.

### Method: one axis of change per kernel

Since none of this could be tested on an R9700 before merge, minimising the
delta from proven code *is* the safety mechanism. Each kernel changes exactly
one thing against its nearest sibling:

| new kernel | derived from | single delta |
|---|---|---|
| `gemm_mq4g256v2_moe_grouped_wmma.gfx12` | qt44 gfx11 `_k2` | WMMA operand shape + intrinsic + C-mapping |
| `gemv_mq4g256v2_moe_gate_up_k8_indexed` | qt13 sibling | group-header decode |
| `gemv_mq4g256v2_moe_down_k8_indexed_batched_expanded` | qt13 sibling | group-header decode |

The gfx11→gfx12 port is not guesswork: the `mq3g256_lloyd` pair is an existing
worked example of that exact transformation and was used as the diff template.

Two properties make the ports safe, and both are worth review attention:

- **Grouped:** tile `kt` covers weights `[kt*16, kt*16+16)`, so the affine-grid
  boundary at weight 128 falls on the tile 7/8 seam. No K-tile straddles the two
  grids, so neither does either `k_grp` half — `upper` is selected once per K4 body.
- **GEMVs:** the grid is *lane-constant*. Thread `tid` covers weights
  `[tid*8, tid*8+8)`, wholly inside one half and the same half in every group, so
  the header offset hoists out of the loops. qt44 reads **one** header u32 where
  qt13 reads two — measurably cheaper (gate_up 125 VGPRs vs 128, down 90 vs 92,
  same occupancy).

### The coupling a reviewer should check first

Adding `routed_indexable_mq4v2` flips `use_gpu_topk` on, which routes qt44 into
a dispatch chain whose trailing `else` calls the **qt13** gate_up kernel. qt13
and qt44 share a 136 B stride and identical nibble packing and differ only in
the 8-byte header, so that fallthrough reads qt44's two f16 scale/zero pairs as
the halves of one f32 — plausible numbers, fluent text, wrong answers, no fault.
The flag and the arms land in the same commit, the qt44 arm precedes the
fallthrough, and `moe_res_mq4v2_mixed_with_qt13_is_not_indexable` pins that a
split qt13/qt44 pairing must fall back rather than guess.

### Verification

`crates/rdna-compute/examples/mq4v2_moe_parity.rs` (lab-gated) is the vehicle
for whoever has an R9700:

```
cargo run --release -p rdna-compute --example mq4v2_moe_parity --features lab
```

The launcher arch-selects, so on gfx1200/gfx1201 this exercises the gfx12 source
no local test can reach. Every check runs against a fixture whose two
128-weight halves occupy **disjoint** ranges, so a grid-selection error is a
~100x scale error rather than a rounding difference — and each check is paired
with a negative control asserting a deliberately grid-swapped reference
*disagrees*. A check that passed with the halves swapped would be measuring
nothing.

**Measured on gfx1151 — all PASS:**

| check | rel_l2 |
|---|---:|
| pack→dequant round trip (host, no GPU) | 9.4e-3 |
| gate_up `y_gate` / `y_up` | 4.8e-7 / 5.1e-7 |
| down `expert_outputs` | 3.6e-7 |
| grouped `Y` (gfx11 source) | 2.9e-4 |
| ninepath D4 fused down+combine | 1.7e-6 |
| grid-swapped negative controls | 0.65 – 1.6, all fired |

**Offline, for gfx1201:** all three compile; the grouped kernel is 55 VGPRs / 0
scratch / 0 LDS / occupancy 16 (its gfx11 sibling is 60), and emits 4
`v_wmma_f32_16x16x16_f16` matching the K4 unroll — the same count as both proven
gfx12 siblings. That last check is not ceremony: the tree documents that some
decode forms silently suppress WMMA emission.

### What is still unverified

Both GEMVs are **verified on real hardware** and are arch-generic — the same
source runs on gfx1201. The grouped kernel's algorithm, header decode, gather
and tiling are verified through its gfx11 sibling. **The gfx12 WMMA intrinsic
swap and C-output mapping are the only axes no local test covers**, and they
ship enabled by default. No performance number on gfx1201 is claimed.

### Unrelated finding

`scripts/compile-kernels.sh` treats `*.gfxNN.hip` as an override for a
same-stem parent, so a variant whose parent has a different basename is **never
compiled by the sweep**. `gemm_mq3g256_lloyd_moe_grouped_wmma.gfx12.hip` is one
(parent is `..._k2.hip`), and there are **178** such orphans repo-wide. Not
touched here — a kernel nothing compiles is a kernel that can rot, but fixing it
would surface an unknown number of pre-existing breakages and belongs in its own
change.

## qt44 decode path — two later commits this description previously omitted

Both measured on gfx1151 **against the shipped artifact**, 4 alternations × 12
runs, cold run discarded, single-stream stateless.

**Ninepath D4 (fused down + combine), `gemv_mq4g256v2_moe_ninepath_d4.hip`.**
qt44 matched neither `ninepath_hfq4` nor `ninepath_mq3l`, so it was denied the
fused down path while *clearing* the shape gate — the published artifact hits
`ninepath_shape_ok` exactly (hidden 2048, `moe_intermediate_size` 512, top-8).
Third port of this kernel, same one-axis delta as the other GEMVs.

Measured: OFF 71.55 tok/s, ON 71.73 — **+0.25%, against a within-arm spread of
71.33–71.88. That is not a measurable win and no speedup is claimed for it.**
The repo's `+0.8%` figure for the qt13 sibling comes from a concurrent serve
battery, a different workload, and was not reproduced. Kept because it is
correct (rel_l2 1.7e-6), lower register pressure (46 VGPRs vs 48), one fewer
launch, and it brings qt44 to structural parity with qt13 and MQ3-Lloyd.
Routing was verified with a probe *inside* the dispatch arm, not inferred from
timings — eligibility flags and JIT-cache presence both proved nothing.

**Contracted MoE decode launch grids — +3.2%, the real win.** Both qt44 MoE
decode GEMVs launched `m` workgroups because I mirrored the qt13 default rather
than reading the kernels' row mapping: `gate_up` maps one block to a row in
*each* of two M/2 outputs (half the blocks existed to exit at the guard), and
`down` owns four consecutive rows (three quarters exited). Pure launch shape, no
kernel change; the parity oracle reports identical rel_l2 with it on or off.

Measured: WIDE 70.94 tok/s, TIGHT 72.94 — and unlike ninepath the arms do not
overlap at all (70.40–71.35 vs 72.75–73.40). Reconfirmed after defaulting it on,
10 runs each: 73.20 (sd 0.23) vs 70.95 (sd 0.17) = **+3.2%**. `gate_up` defaults
on for **gfx1151 only** — the arch it was measured on, per this repo's rule that
a specialisation ships where it was measured. `HIPFIRE_MQ4V2_GATE_UP_TIGHT_GRID`
forces either way so an R9700 owner can widen it. The `down` contraction stays
opt-in (`HIPFIRE_MQ4V2_DOWN_TIGHT_GRID`): with D4 default-on, that kernel is not
on the decode path and there was nothing to measure it against.

## Stack and ownership

Targets `master`. Previously stacked on #599 (`quant/quality`), itself stacked on
#597 (`feat/ternary-bonsai-27b`); both have merged.

This branch does **not** modify #599's qt44/qt45 format internals. The packers
(`quant_fwht.rs`), the wire tags (`hfq.rs`) and `saddle-quant` are untouched:

```
git diff --stat 3243261ea -- crates/hipfire-quantize/src/quant_fwht.rs \
    crates/hipfire-quantize/src/hfq.rs crates/saddle-quant/
```

is empty. What it adds is arms and kernels that let those formats reach paths
they could not previously reach.

## What ships

`ornith-ai/Ornith-1.5-35B-A3B` (Qwen3.5-family VL finetune, 40L, 256 experts /
top-8, hidden 2048, DeltaNet 3:1, 248,320 vocab) as:

Published and verified readable at
[`hipfire-models/ornith1.5-35b-a3b`](https://huggingface.co/hipfire-models/ornith1.5-35b-a3b).
Byte counts below are the **HF LFS sizes of the live files**, re-read
2026-08-25, not build-time figures:

| artifact | bytes | | notes |
|---|---:|---:|---|
| `ornith-1.5-35b-a3b.mq4` | 19,020,298,496 | 17.7 GiB | text trunk, MQ4G256V2 (qt44) |
| `ornith-1.5-35b-a3b.mtp` | 484,778,112 | 462 MiB | MTP draft head — **replacement**, see below |
| `ornith-1.5-35b-a3b.vl` | 907,482,592 | 865 MiB | vision tower, F16 |

Built with **`--format mq4`**, which routes to qt44 (`pipeline.rs:285`).

Three things an earlier revision of this section got wrong, corrected here
against the published files rather than from memory:

* **The trunk is 19.0 GB, not 22.9 GB**, and the sidecar is 462 MiB, not
  856 MiB. Those were pre-rescope build figures for artifacts that were never
  uploaded.
* **There is no combined `-vl.mq4` artifact.** The vision tower ships
  *alongside* the trunk as a separate `.vl` file; the old 23,748,214,496 row
  was trunk-plus-tower fused into one file, which is not what shipped.
* **`--format mq4v1` (qt13) was not used.** Read from the HFQ index of the
  live `.mq4` (magic `HFQM`, arch_id 6, 21,093 tensors): qt44 `MQ4G256V2`
  20,651 — including *every* `mlp.experts.*` routed expert — plus 191 F16,
  112 Q8-class, 96 qt13 `MQ4G256` (shared-expert weights only) and 43 qt15
  `MQ6G256`. The recipe is a hybrid with a qt44 routed-expert core.

That last point is what makes this branch's gaps concrete rather than
prospective: both gaps in "RDNA4 support" above apply to the model that is
live on HF today.

The shipped `.mtp` is **not** Ornith 1.5's native draft head. It is built from
[`shisa-ai/Ornith-1.5-35B-A3B-MTP-ONLY`](https://huggingface.co/shisa-ai/Ornith-1.5-35B-A3B-MTP-ONLY)
(Apache-2.0). The vendor head drafts at ~13% acceptance — measured
independently here and in llama.cpp, and reported upstream in the base repo's
discussion #10 — which makes speculation *slower* than plain AR decode. The
substitution is documented in the model card and the repo NOTICE.

## Measurements

All against a **6-bit oracle**, 96 chunks, WikiText-2 slice, on **gfx1151**.

```
oracle (mq6)                 PPL 8.1945   (mean NLL 2.103459, validated)
shipped SKU     KLD 0.088581 PPL 8.1667
```

Coherence: 5/5 genres, `hard_errors=0`, all five completions read and verified
correct by inspection, not by threshold.

MTP engagement. **The 52.3 → 58.1 tok/s pair an earlier revision quoted here is
withdrawn** — it was measured against the *native* head, which is not what
shipped, and predates the `mtp_extract` index-resolution fix. The published
figures, gfx1151, 256 generated tokens, greedy, reasoning off, Q8 KV, medians
of repeated runs on an idle box:

| configuration | tok/s | τ | accept |
|---|---:|---:|---:|
| autoregressive (no speculation) | ~70 | — | — |
| MTP, K=3 | 77.4 | 2.5 | 0.61 |
| MTP + ngram-modulated drafting | ~150 | 5.43 | 0.937 |

+11% for plain MTP over AR, ~2.1× for the ngram arm (which engages only with
reasoning disabled, `max_think_tokens = 1`). Per-depth acceptance for the
shipped head is 0.877 / 0.620 / 0.528 against the native head's
0.823 / 0.292 / 0.179 — so unlike the withdrawn pair, these are backed by
acceptance telemetry, not inferred from a throughput delta.

Note the ngram arm is a serving feature already on `master`; it is quoted for
completeness because it is what the model card advertises, and nothing in this
branch produces or depends on it.

VL: `doge.jpeg` described specifically and correctly — Shiba Inu, the Doge meme
dog, named Kabosu, close-up head and shoulders, golden tan to cream, large dark
eyes looking up and to the left, black nose. Checked against the image.

### What these numbers are not

This is **gfx1151**. qt44/qt45 batched prefill is gfx1200/1201-only, and the
engine itself warns "vision tower not yet validated on gfx1151". No figure here
is comparable to #599's gfx1201 arms, and the oracle is 6-bit rather than F32 —
a high-precision oracle is **not constructible** for this model, for two
independent reasons: F32 weights are 139 GB against 73 GB of available RAM, and
`handle_moe_expert_3d` cannot emit above 6-bit for routed experts at all, so
`--format q8f16` yields an artifact that is still 4-bit qt13 in the experts
(measured 4.55 bits/weight; built, inspected, discarded).

## Defects found in the base stack

Seven fixed here, one characterised. All originate in #597/#599; **none are on
master**, so no shipped quantizer is affected.

Three trace to `d1d172e9c`, whose message states "byte-identical output":

1. **The F16 fallback arm was deleted.** Every norm and every bias vanished from
   every artifact — 191 tensors on this model. The quantizer exits 0, tensor
   counts and byte sizes look plausible, and the model cannot load
   (`tensor not found: norm.weight`). The only trace was `Quantized params`
   sitting 176,768 below `Total params`, printed as "100.0%" after rounding.
   Restored, plus an `F16 params` line and a param-accounting assertion that
   refuses to write when `quantized + f16 != total`.
2. **The `if is_moe_expert_3d` guard was dropped**, leaving the predicate
   computed and discarded. Any 2-D tensor named `*experts.gate_up_proj` panics.
   Latent until a model ships un-stacked experts — Ornith 1.5's MTP head does.
3. **K-map `Promote6` never fired for qt44/qt45 routed experts.** `--format mq4`
   silently dropped 8,192 expert tensors from 6-bit to 4-bit.

Also:

4. **qt44 had no MoE grouped-expert kernel on any architecture**, so a qt44 A3B
   MoE could not prefill at all. Added for gfx11.
5. ~~**qt44's GEMM family was gfx12-only.** Added gfx11 ports of `gate_up`,
   `qkv`, `qkvza` and `residual`.~~ **Superseded and removed** — master's MQ V2
   release ships WMMA implementations of all four for gfx11, which are strictly
   better than these scalar ports.
6. **`ensure_fp16_x` is pointer-keyed**, and MoE prefill reuses one x tensor
   across layers with different contents. Switched to the uncached variant that
   exists for exactly this hazard.
7. **AR graph capture kills qwen35 decode** with `hipError(906)` — a D2H memcpy
   inside a capturing stream, reached because qt44 is not in the indexable set
   and falls to the non-capturable CPU top-K path. Worked around with
   `HIPFIRE_AR_GRAPH=0`; root cause recorded, not fixed.

I would not merge #597 as-is. Defect 1 alone makes every model it produces
unloadable, from a commit claiming byte-identical output.

## The qt44 finding — characterised, not fixed

qt44 is genuinely the better format. On real FWHT-rotated expert weights its
reconstruction MSE is **0.842×** qt13's, and on the decode path it measures
**0.059** against qt13's **0.081**. But:

```
prefill  qt13 0.0696   qt44 0.7614
per-tok  qt13 0.0809   qt44 0.0590
```

Isolated by subtraction — two artifacts differing by **exactly 96 tensors** (the
shared expert: 36 gate_proj, 36 up_proj, 24 down_proj) score 0.0886 and 0.9727.
All three kernels were parity-clean at production dimensions (gate_up 1.07e-7,
residual 1.26e-7, MoE grouped 1.9e-5 including multi-expert gather and `-1`
padding) — only the MoE grouped kernel survives the rescope, but the measurement
is what matters here: the fault is in path selection, not decode. `gate_fusable`
requires `shared_expert_gate == MQ4G256` *exactly*, so qt44 diverts the arch-6
FFN preamble.

That is the same defect shape as the `needs_x_rot_local` hazard this repo
already documents — a dtype added to the format layer without being added to
every predicate that enumerates its family.

An earlier revision closed this section with "hence `--format mq4v1` for the
shipped artifact". **That never happened** — the artifact on HF is qt44
(`--format mq4`), see "What ships". So the shipped model was resolving to
`use_gpu_topk = false` and decoding through the resident CPU-fallback path, and
on RDNA4 it hit the grouped-prefill refusal outright.

The proper fix — indexed-GEMV kernels for `gate_up` and `down` so qt44 takes
Path 1 like qt13, which also clears defect 7 — is **no longer deferred**. It is
in this branch: see "2. qt44 had zero MoE GEMVs on any arch" above. What remains
unfixed here is the `gate_fusable` predicate itself; the shipped router is Q8
rather than MQ4, so the fused gate-side GEMV does not apply to this model either
way, and `moe_res_shipped_ornith15_takes_the_indexed_path` pins that combination
explicitly (asserting `!gate_fusable` *and* the indexed path).

## Tests

`mq4v2_gemm_gfx11_parity` covered gate_up, MoE grouped and residual against a
two-grid oracle, with a single-header reference asserted to **disagree** — a
negative control that passes would mean the test measures nothing. **Removed in
the rescope:** it existed to prove the scalar dense ports, which are gone.

The gap that left — no GPU oracle for the grouped MoE kernel — **has since been
filled**, and the note claiming otherwise is withdrawn.
`crates/rdna-compute/examples/mq4v2_moe_parity.rs` covers the grouped kernel
plus all three new GEMVs, with grid-swapped negative controls; results are in
"Verification" above.

Two fixture traps were documented in it because both produced false failures on
correct kernels, and both would have led to "fixing" working code — worth
carrying into any replacement oracle:

* fp16 A-operand rounding amplified by catastrophic cancellation (17.8%);
* per-element relative error dividing by a near-zero denominator (36%) —
  invisible at N=5 with 160 outputs, unavoidable at N=512 with 524,288.

Also fixed: `mtp_extract` rejected Ornith 1.5 outright because its MTP head
ships routed experts un-stacked. The fusion is gate-then-up; the reverse loads
cleanly and produces fluent, wrong output.

## Gates

Re-run against `master` after the rescope. **These were taken at head
`77e63e204` and have not been re-run at the current head `16cc144a7`** — nine
commits later (gfx12 grouped GEMM, indexed GEMVs, parity oracle, shipped-format
fix, ninepath D4, tight grids). Treat the list as stale until re-run:

* `cargo check --workspace --all-targets`: 0 errors
* `cargo test`: `hipfire-quantize` 45 + 164, `hipfire-dispatch` 192,
  `rdna-compute` 189 — all pass
* `leanup-ratchets.sh`: adds **zero** new violations — `bypass_total` 232 and
  `ungated_examples` 46 both match master exactly. The 3 reds are master's own.
* `ratchet-diff.sh`: no expectation weakened
* `check-crate-maps.py --check`: no drift from this branch; maps regenerated for
  the four crates it touches
* One pre-existing failure, **not from this branch**:
  `quant_fwht::tests::mq6v2_mq5v2_wire_layout` fails identically on master HEAD
  `336dd0777`, reproduced on a clean master worktree.

A previous revision said "**No GPU runs since the rescope**". That is no longer
true: the parity oracle, the ninepath routing probe and both decode benches
above all ran on gfx1151 at or near current head. What has *not* been re-run
since the rescope is the **coherence gate** and the KLD/PPL measurement — those
figures are still the pre-rescope ones, and nothing in the rescope changes the
artifact they were measured against, but they have not been reconfirmed.

## The rename, and the registry entry (`313f276bc`, `16cc144a7`)

The artifacts went up as `ornith1.5-*`; every other reference to this family
uses the hyphenated `ornith-1.5`. **Both the HF files and the registry are now
renamed.**

On HF this was a server-side copy+delete of the LFS pointers — the blobs are
untouched, nothing was re-uploaded, and the sha256s are unchanged:

| file | bytes | sha256 |
|---|---:|---|
| `ornith-1.5-35b-a3b.mq4` | 19,020,298,496 | `c4e40d69…` |
| `ornith-1.5-35b-a3b.mtp` | 484,778,112 | `38c22acc…` |
| `ornith-1.5-35b-a3b.vl` | 907,482,592 | `7dd01136…` |

The curated `registry/models.json` entry, previously held back, **lands here**:
its stated precondition was that the upload be verified present and readable,
and it now is. `registry_gen.py` was run against live HF and builds clean —
78 models, 79 aliases, resolving `arch_id` 6, `quant` mq4, and byte counts that
agree with `size_gb` 19.02 inside `SIZE_TOLERANCE`.

**The repo id is renamed too** — `hipfire-models/ornith-1.5-35b-a3b`, with HF
redirecting the old id. That step needed org-level write and was done by an
admin; the publishing token has only per-repo scopes and `/api/repos/move`
returns 403 for it. The curated `repo` field now names the canonical id
(`16cc144a7`); it deliberately stayed on the old one until the move landed,
because HF only redirects *after* a move and a `repo` that 404s fails the daily
run closed. Rebuilt against live HF and diffed: `repo` is the only field that
changed — sha256, `size_bytes`, `arch_id`, `quant` and the MTP sidecar's
hash/size are byte-identical either way, which is what confirms the redirect and
the direct id serve the same repo.

### Both spellings keep working

Three mechanisms, because they cover different failure modes:

1. **Aliases.** `ornith`, `ornith-1.5`, `ornith1.5` and `ornith1.5:35b-a3b` all
   resolve to the canonical `ornith-1.5:35b-a3b`.
2. **`arch_id_for` maps the `ornith-1.5` family.** This is the trap the rename
   sets, and it is worth a reviewer's attention: the function keys on
   `tag.split(":", 1)[0]`, so the hyphen lands *inside* the family segment —
   `ornith-1.5:35b-a3b` yields `"ornith-1.5"`, which the previous tuple
   `("ornith1.5", "ornith")` did not contain. An unmapped family returns `None`
   and the daily refresh is fail-closed, so a rename that updated the tag but
   not the tuple would have taken down the **entire** registry — every other
   model's entry included — at 06:23 UTC. `ornith1.5` stays mapped so reverting
   cannot fail-close either.
3. **A separator-insensitive fallback in `find_model_path`.** This covers the
   case an alias cannot reach. Someone who downloaded before the rename has
   `ornith1.5-35b-a3b.mq4` on disk while the registry now points at
   `ornith-1.5-35b-a3b.mq4`. The *old* tag already found their file — the fuzzy
   scan keys off the raw input, not the resolved tag — but the **canonical** tag
   did not, and would have silently re-downloaded 19 GB.

The fallback runs only when the literal pass returns nothing, so it can add a
match where there is none today but can never change one that already resolves.
`literal_match_wins_over_the_separator_fallback` pins that: with both spellings
on disk, each input reaches its own file. And the fallback test is confirmed
non-vacuous — with the fallback removed,
`separator_spellings_find_an_already_downloaded_file` fails on exactly the
canonical-tag assertion while the other two still pass.

`test_registry_gen_ornith.py` grew to 8 checks, including one that asserts
*every* curated alias target has an arch mapping — an alias pointing at an
unmapped tag resolves fine for users while aborting the run.

**Review note:** touching `crates/hipfire-cli/src/main.rs` drags in that file's
pre-existing rustfmt debt, because CI's gate checks each changed `.rs` file in
full. Roughly 205 of the 271 changed lines in it are that mechanical reformat,
not this change.


Kaden-Schutt:
Triage: HOLD_WIP after the MQ V2 release. Dense gfx11 qt44 GEMMs are now superseded by master; the Ornith SKU and grouped qt44 MoE kernel remain unique. Rebase onto master and rescope to those remaining pieces when the product thread resumes.

nwoolmer:
Rebased onto `master` and rescoped, per the HOLD_WIP triage. Base retargeted `quant/quality` → `master`; head is `77e63e204`.

### The supersession, verified rather than assumed

Confirmed your read before cutting anything. Master's MQ V2 release ships WMMA implementations of all four dense qt44 GEMMs for gfx11 — `gemm_{gate_up,qkv,qkvza}_mq4g256v2_wmma_gfx11_bt`, `*_gfx1100_mw_lds`, plus gfx12/gfx1201 siblings. This branch carried *scalar* ports of the same four. The merge conflicts made the comparison explicit: every one was my `gemm_*_mq4g256v2_gfx11(...)` against master's `if has_wmma_w32() { gemm_*_mq4g256v2_wmma(...) }`. Master's is strictly better.

Also checked the other half rather than taking it on trust: master has grouped-MoE kernels for hfq3/hfq4/hfq6/mq2/mq3/paro/mfp4/mixed — and **none for any `mq4g256v2` shape**. So the MoE kernel really is unique.

### Dropped

- `kernels/src/gemm_{gate_up,qkv,qkvza}_mq4g256v2.hip`, `gemm_mq4g256v2_residual.hip`
- their four `*_SRC` consts and four `gemm_*_mq4g256v2_gfx11` dispatch functions
- `crates/hipfire-runtime/examples/mq4v2_gemm_gfx11_parity.rs` — the GPU oracle that existed only to prove those scalar ports matched, so it went with them

### Kept

- `gemm_mq4g256v2_moe_grouped_wmma_k2` + its `.hip`. Without it a qt44 A3B MoE model fails prefill outright.
- Ornith SKU, MTP sidecar, VL wiring, `coherence-gate-ornith15.sh`, `registry_gen.py` entry.

### Two merge decisions I want on the record

**1. The F16 fallback was restored twice, independently.** Both this branch and master separately restored the arm `d1d172e9c` deleted. Master's version is more general — fp8 scales, spill handling, and it preserves exact F16 bits when the source is already F16 — so the merge takes master's body.

I kept this branch's **param-accounting check**, since it is what caught the original 501-tensor loss on Ornith 1.5 and master has no equivalent. But I changed how it closes. This branch added a separate `f16_params` counter; master's test at `pipeline.rs` asserts `assert_eq!(quantized_params, 4096)` for exactly this F16-fallback path. Keeping my counter would have silently broken that committed expectation, so the counter and its `MainQuantState` field are removed and the check closes on `quantized_params` alone. Master's recipe census already prints the per-quant-type F16 count, so nothing is lost in the reporting.

**2. `moe_expert_3d_applies`** is kept over master's inline form — pure extraction, identical semantics, and `handle_moe_expert_3d` needs it.

### Verified

- `cargo check --workspace --all-targets` — 0 errors
- `hipfire-quantize` 45 + 164, `hipfire-dispatch` 192, `rdna-compute` 189 pass
- One pre-existing failure, **not from this branch**: `quant_fwht::tests::mq6v2_mq5v2_wire_layout` fails identically on `master` HEAD `336dd0777`. Reproduced on a clean master worktree before attributing it.
- `leanup-ratchets` — this branch adds **zero** new violations: `bypass_total` 232 and `ungated_examples` 46 both match master exactly. The 3 reds are master's own.
- `ratchet-diff` — no expectation weakened.
- Regenerated the crate maps for the four crates touched (rdna-compute, quantize, dispatch, runtime); remaining stale maps are master's.

Updated the design doc's hardware-limitation section, which still claimed gfx11 has only the scalar fallback — no longer true post-MQ V2.

**No GPU runs.** The coherence gate and any qt44 MoE prefill numbers have not been re-measured against current master; this is a static rescope plus test-suite validation only.

nwoolmer:
Pushed RDNA4 / R9700 (gfx1201) qt44 MoE support at `1704a90e4`. PR body updated with the full write-up; this is the short version plus the handoff.

**You were right that kernels were missing.** The qt44 *dense* path on gfx12 was already complete — all four ops have `.gfx12` kernels plus batch-tiled variants, and qkv even has a gfx1201-specific `_bt`. The **MoE** path had two independent holes:

1. **Grouped prefill was a hard error on RDNA4.** `if self.arch_caps.is_rdna4() { return Err("no gfx12 kernel for qt44 MoE experts") }`. qt44 was the only MoE quant type with a grouped kernel and zero gfx12 coverage (mq2-Lloyd has 3, hfq4 has 4, hfq6 has 2). Routed experts are ~99% of an A3B's tensors, so a qt44 Ornith artifact **could not prefill at all on an R9700**.
2. **qt44 had zero MoE GEMVs on any arch** — qt13 has ~19. So `routed_indexable_*` could never hold and qt44 fell to the resident CPU-fallback decode. That is the root of the `gate_fusable` diversion already documented in this PR, and the reason the shipped artifact is `--format mq4v1`.

Three kernels added, each a one-axis delta from a proven sibling (the `mq3g256_lloyd` gfx11→gfx12 pair was the diff template for the port; the qt13 GEMVs for the header change).

### For whoever has the R9700

```
cargo run --release -p rdna-compute --example mq4v2_moe_parity --features lab
```

That is the whole ask — the launcher arch-selects, so on gfx1200/gfx1201 it exercises the gfx12 grouped source that no local test can reach. Exit code is non-zero on failure, so it can gate a script. It also runs a host-only self-test first, and if GPU init fails it exits 2 saying the kernel checks did **not** run, rather than reporting a pass.

The fixture is built so a wrong answer is loud: both 128-weight halves occupy disjoint ranges ([-1,1] and [96,160]), so a grid-selection error is a ~100x scale error rather than a rounding difference — and every check is paired with a negative control asserting a deliberately grid-swapped reference *disagrees*. A run that passed with the halves swapped would be measuring nothing.

### What I could and could not establish here

**Measured on gfx1151, all PASS:** gate_up `rel_l2` 4.8e-7 / 5.1e-7, down 3.6e-7, grouped (gfx11 source) 2.9e-4, and all three grid-swapped controls fired at 0.65–1.6.

Both GEMVs are therefore **verified on real hardware and are arch-generic** — the same source runs on gfx1201. The grouped kernel's algorithm, header decode, gather and tiling are verified through its gfx11 sibling.

**Offline for gfx1201:** all three compile; grouped is 55 VGPRs / 0 scratch / 0 LDS / occupancy 16 (gfx11 sibling: 60), and emits 4 `v_wmma_f32_16x16x16_f16` — matching the K4 unroll and both proven gfx12 siblings. Worth checking explicitly rather than assuming, since the tree documents that some decode forms silently suppress WMMA emission.

**Unverified:** the gfx12 WMMA intrinsic swap and C-output mapping. That is the one axis no local test covers, and per your call it ships enabled by default. No gfx1201 performance number is claimed — I have no way to produce one.

### One thing to check first in review

Adding `routed_indexable_mq4v2` flips `use_gpu_topk` on, which routes qt44 into a dispatch chain whose trailing `else` calls the **qt13** gate_up kernel. The two formats share a 136 B stride and identical nibble packing and differ only in the 8-byte header, so that fallthrough reads qt44's two f16 pairs as the halves of one f32 — fluent text, wrong numbers, no fault. Flag and arms land in one commit, the qt44 arm precedes the fallthrough, and a test pins that a split qt13/qt44 pairing must fall back rather than guess.

### Unrelated finding, not fixed

`scripts/compile-kernels.sh` treats `*.gfxNN.hip` as an override for a **same-stem** parent, so a variant whose parent has a different basename is never compiled by the sweep. `gemm_mq3g256_lloyd_moe_grouped_wmma.gfx12.hip` is one (its parent is `..._k2.hip`) — and there are **178** such orphans repo-wide. A kernel nothing compiles is a kernel that can rot, but fixing the glob would surface an unknown number of pre-existing breakages, so it belongs in its own change.

### Gates

`master` moved 12 commits during this work and is now red on four ratchets (`daemon_lines` 3913, `ungated_examples` 48, `bypass_regressions` 1, `bypass_total` 232). I reproduced that set on a clean `master` worktree — it is byte-identical to what CI reports here, so **this branch still adds zero gate violations**. `ratchet-diff`: no expectation weakened.

nwoolmer:
**Correction to my previous comment**, plus the answer to "does R9700 actually get prefill now".

### The shipped artifact is qt44, not mq4v1

I wrote that the shipped Ornith was built `--format mq4v1` (qt13). That is wrong. I took it from the "qt44 finding" section of this PR body and restated it without checking the artifact. Read from the HFQ index of the published `hipfire-models/ornith1.5-35b-a3b.mq4` (magic `HFQM`, arch_id 6, 21,093 tensors):

| qt | format | count |
|---:|---|---:|
| **44** | **MQ4G256V2** | **20,651** — incl. every `mlp.experts.*` routed expert |
| 1 | F16 | 191 |
| 3 | Q8-class | 112 |
| 13 | MQ4G256 | 96 — shared-expert weights only |
| 15 | MQ6G256 | 43 |

`--format mq4` routes to qt44 (`pipeline.rs:285`); `mq4v1` is the qt13 escape hatch and was not used. The shipped recipe is a **hybrid**: routed experts qt44, router and `shared_expert_gate` at Q8, shared expert graded MQ6/qt13 across layers. The line in the PR body is stale and is now marked as such.

This makes both gaps concrete rather than prospective — they apply to the model live on HF today:

- `routed_gate_up`/`routed_down` come from `experts[0]` = qt44, so the shipped model resolved to `use_gpu_topk = false` and has been decoding through the resident CPU-fallback path.
- Both grouped-prefill call sites dispatch on the routed dtype, so on RDNA4 the shipped model hit the refusal.

`moe_res_shipped_ornith15_takes_the_indexed_path` now pins the published model's exact dtype combination. Worth noting it asserts `!gate_fusable` — the router is Q8, not MQ4, so the fused gate side does not apply, and that does *not* disqualify the routed indexed path (same coupling `moe_res_q8_router_still_gpu_topk` already pins for qt13).

### Yes — R9700 prefill is fixed, for this exact artifact

Both grouped call sites key on the routed dtype:

```
dispatch_grouped_gemm(gpu, p.dtypes.routed_gate_up, ..)   // pipeline/mod.rs:2471
dispatch_grouped_gemm(gpu, p.dtypes.routed_down,    ..)   // pipeline/mod.rs:2706
```

Both are `MQ4G256V2` on the shipped model, so both land on `gemm_mq4g256v2_moe_grouped_wmma_k2`, which is now arch-selecting and picks the gfx12 source on RDNA4. Previously that call returned `Err("no gfx12 kernel for qt44 MoE experts")`, so the published Ornith could not prefill on an R9700 at all. The shared expert is MQ6/qt13 dense — those paths already have full gfx12 coverage — so the grouped kernel was the only RDNA4 blocker.

Still needs the parity run on real gfx12 to confirm numerics; the dispatch reachability is what's settled here.

### The decode gap is not closed for everyone yet

The two GEMVs added here are arch-generic (zero `arch_caps` branches) and verified on gfx1151, so every arch gets them. But qt44 is **still excluded from the ninepath fused gate_up**:

```rust
let ninepath_hfq4 = ninepath_shape_ok
    && p.dtypes.routed_gate_up == DType::MQ4G256
    && p.dtypes.routed_down == DType::MQ4G256;
let ninepath_mq3l = ninepath_shape_ok && ..MQ2G256Lloyd && ..MQ3G256Lloyd;
```

qt44 matches neither. And the shipped Ornith hits `ninepath_shape_ok` *exactly* — from its own config: `hidden_size` 2048 (≤2048), `moe_intermediate_size` 512 (mi==512), `num_experts_per_tok` 8 (k==8). So it clears the shape gate and is denied the fused path purely for lack of a qt44 port, on the precise shape that path was tuned for. The repo's own note puts D4 at "+0.8% on the A3B serve battery" for qt13.

That, not more gfx1201 work, looks like the highest-value next kernel: arch-generic, benefits the Strix in hand as much as an R9700, and — unlike the gfx12 grouped port — it can be measured locally against the real artifact.

Separately, qt13 has five gfx1151-tuned decode kernels (`row1_buffer`, `row2_buffer`, `row8`, `persistent_rank8`, `projection_indexed`) where qt44 has one generic kernel per op. I'd not port those blind — each was admitted by a measurement on its target. Measure qt44 decode on Strix first and let the numbers choose.

### CI

`build` and `unit tests` both **pass**. `gates` and `clippy` are red on `master` itself — I reproduced `master`'s four ratchet violations on a clean worktree and they are byte-identical to what CI reports here, so this branch still adds zero.

cat5edopeHA:
## gfx1201 / R9700 real-hardware test of 95bb365f — parity PASS, serve output DEGENERATE

Running the R9700 verification you asked for. Built `quant/ornith-1.5` @ `95bb365f` clean (no local patches) on dual-R9700 gfx1201, HIP 7.2.

### Parity oracle: ALL PASS

`cargo run --release -p rdna-compute --example mq4v2_moe_parity --features lab` on gfx1201:

```
host self-test: pack->dequant rel_l2=9.433e-3 cos=0.999956 PASS
                grid-swapped negative control 1.409e0        PASS
gate_up y_gate      rel_l2=4.767e-7   cos=1.000000   PASS
gate_up y_up        rel_l2=5.097e-7   cos=1.000000   PASS
gate_up neg control rel_l2=7.609e-1                  PASS
down expert_outputs rel_l2=3.581e-7   cos=1.000000   PASS
down neg control    rel_l2=6.486e-1                  PASS
grouped Y           rel_l2=2.922e-4   cos=1.000000   PASS  (arch-selecting launcher -> gfx12 source)
grouped neg control rel_l2=1.597e0                   PASS
mq4v2_moe_parity: all checks PASSED, exit=0
```

So the WMMA intrinsic swap and C-output mapping you flagged as unverifiable locally are **numerically correct at test dims** on real RDNA4.

### Serve path: fast but wrong on every config

Shipped artifact `hipfire-models/ornith1.5-35b-a3b.mq4`, isolated home, GPU0, q8 KV contiguous, 16K ctx, greedy, thinking off:

| config | prefill t/s | decode t/s | output |
|---|---:|---:|---|
| default + `HIPFIRE_AR_GRAPH=0` | 947.8 | 113.3 | **DEGENERATE** — `# Matrix exponentiation` repeated |
| default, no env at all | ~6.7* | 70.9 | degenerate (same loop) |
| `HIPFIRE_PREFILL_BATCHED=0` | 114.5 | 126.9 | **DEGENERATE** — `The\nThe\nThe...` |
| `PREFILL_BATCHED=0 MOE_GROUPED_GEMM=0 GRAPH_MOE=0` (the combo that was CORRECT on 77e63e20) | 114.2 | 114.6 | **DEGENERATE** |

*first run before AR_GRAPH=0; ttft dominated by graph-capture retry.

Baseline for comparison — same artifact, previous head `77e63e20`, same 3-flag combo: pp 3.0 / tg 65.1, **output correct** (coherence smokes 4/4 incl. arithmetic). So the speed win is enormous (~1.75x tg; the per-token prefill fallback itself went 3 -> 114 t/s), but the model emits fluent repetition under *every* config on the new build, including the exact flag set that produced correct text one commit earlier. That pins the corruption to the new decode wiring (`use_gpu_topk` now ON for this dtype combination via `routed_indexable_mq4v2`), not to the grouped kernel or AR graph capture.

### This looks like the fallthrough hazard from your review note

Your 18:15Z comment ("One thing to check first") describes exactly this signature: `use_gpu_topk` routes qt44 into a chain whose trailing else calls the **qt13** gate_up kernel; shared 136 B stride and nibble packing, header misread as f32 scale -> fluent text, wrong numbers, no fault. `95bb365f` adds `moe_res_shipped_ornith15_takes_the_indexed_path`, which pins resolve() output — but the observed failure suggests the launch-site dispatch still has a qt13 arm reachable for uniform-qt44 routed pairs at production dims (256 experts / top-8 / hidden 2048 / mi 512). The parity example exercises each kernel in isolation at toy dims (n_exp=8, K=512) and passes, so the gap is between resolve() and the actual GEMV selection, not inside any single kernel.

Happy to run a targeted repro if useful — e.g. an env kill-switch that forces `use_gpu_topk=false` for qt44 would let me confirm whether CPU-fallback decode restores correctness on 95bb365f (it was correct there on 77e63e20).

### Environment

device gfx1201 (R9700 32 GB x2, test isolated to GPU0), Ubuntu 26.04, HIP 7.14 runtime string in the example, built with stock `cargo build --release -p hipfire-cli -p hipfire-daemon -p hipfire-quantize`. No local patches; worktree detached at `95bb365f78fa`.

nwoolmer:
Pushed the ninepath D4 port for qt44 at `26c0dcb6e`, plus the measurement — which is a **negative result**, so recording it properly.

### Is it actually routed to? Yes — verified, not inferred

I put a probe **inside the dispatch arm** rather than trusting the timings:

```
ON  (default):            [ARM-PROBE] mq4v2-ninepath-d4 ARM TAKEN
OFF (NINEPATH=0):         (no probe fires — neither arm)
```

Two things fall out. The qt44 arm is genuinely reached, and the qt13 fallthrough is **never** taken for qt44 — so the silent-misread hazard I flagged is not live.

Worth recording that two weaker checks misled me first:

- **Eligibility flags are not execution.** My first probe sat at the `ninepath_eligible` computation and showed `mq4v2=true`. That says nothing about which arm runs.
- **The JIT cache is populated eagerly in BOTH arms.** I cleared `gemv_mq4g256v2_moe_ninepath_d4.*` and it reappeared after an OFF run too, so cache presence proves nothing about dispatch.

The env flag does propagate (`mode=""` vs `mode="0"`), and the shipped model clears the shape gate exactly: `k=8 mi=512 hidden=2048 bs=1`, no dtype tags, no AWQ.

### Measured: no decode win on single-stream

gfx1151, shipped artifact, 4 alternations × 12 runs, cold run discarded, single-stream stateless:

| arm | medians | mean of medians |
|---|---|---:|
| OFF (expanded+combine) | 71.50, 71.50, 71.80, 71.40 | **71.55** |
| ON (ninepath d4) | 71.40, 72.00, 71.90, 71.60 | **71.73** |

+0.18 tok/s (+0.25%) — against a within-arm spread of 71.33–71.88. The between-arm difference is smaller than the variation inside either arm. **Not a win**, and no speedup should be quoted for this kernel.

### Correcting my own justification

I proposed this port citing qt13's "+0.8% on the A3B serve battery". I quoted a qt13 number from a **concurrent/multiturn** workload and then measured on single-stream stateless — the least favourable case for D4, whose benefit is removing the expanded `[8 × down_m]` intermediate and a combine launch, which compounds under concurrency. So the honest status is "untested on the workload the claim came from", not "the claim is false". I started that arm (`--concurrency 1,4 --backend slots --workload both`) and stopped it to free the GPU for the routing question; it remains open.

### Keeping it anyway

Correct (parity rel_l2 1.7e-6 with its grid-swapped control firing at 1.05), lower register pressure (46 VGPRs vs qt13's 48, same occupancy), one less kernel launch, and it brings qt44 to structural parity with qt13 and MQ3-Lloyd. Neutral, not a regression. The kernel header now carries the measurement and an explicit "do not quote a speedup" note so the next reader does not inherit the same unverified figure I did.

### Branch state

`26c0dcb6e`. Workspace check clean, `hipfire-dispatch` 195 tests pass, ratchets unchanged from the master baseline, `ratchet-diff` reports nothing weakened.

The R9700 prefill fix and the two indexed decode GEMVs remain the substantive wins here — the GEMVs verified on hardware at 4.8e-7 / 3.6e-7. Ninepath is a structural tidy-up whose payoff, if any, is on a workload still unmeasured.

nwoolmer:
Found a real decode win on Strix: **+3.2%** on the shipped Ornith 1.5, pushed at `4a31e6ba6`.

### First, a negative finding that redirected the search

I had said qt13 has five gfx1151-tuned decode kernels that qt44 lacks, implying a gap to close. That was misleading. Every one of those specialisations is **opt-in**:

```
HIPFIRE_GFX1151_DOWN_ROW1_BUFFER   == Ok("1")
HIPFIRE_GFX1151_DOWN_ROW2_BUFFER   == Ok("1")
HIPFIRE_GFX1151_DOWN_ROW8          == Ok("1")
HIPFIRE_GFX1151_GATE_UP_PERSISTENT_RANK8 == Ok("1")
...
```

Across both qt13 decode launchers there is exactly **one** default-on gate, and it is scoped to `is_gfx1100()`. So on Strix, qt13's production decode already runs the same generic kernels qt44 now has — there was no dormant advantage to port, and porting those variants would have been porting research.

### The actual win came from reading what the kernels index

Both qt44 GEMVs were launching far more workgroups than they use, because I mirrored the conservative default from their qt13 siblings rather than the kernels' own row mapping:

| kernel | launched | actually needed | wasted |
|---|---|---|---|
| gate_up | `m` (= 2·mi) | `mi` — kernel returns for `row >= mi` | **half** |
| down | `m` | `m/4` — kernel owns 4 rows per workgroup | **three quarters** |

Pure launch shape, no kernel change, and value-preserving: the parity oracle reports identical rel_l2 either way (gate_up 4.767e-7 / 5.097e-7, down 3.581e-7).

### Measured

gfx1151, shipped artifact, 4 alternations × 12 runs, cold run discarded, single-stream stateless:

| arm | medians | mean |
|---|---|---:|
| WIDE (`m`) | 70.80, 71.20, 71.35, 70.40 | **70.94** |
| TIGHT (`m/2`) | 72.75, 73.40, 72.75, 72.85 | **72.94** |

+2.00 tok/s (+2.8%), and — unlike the ninepath port — **the arms do not overlap**: WIDE spans 70.40–71.35, TIGHT spans 72.75–73.40. Re-confirmed after making it default, 10 runs each with the escape hatch forced off: **73.20 (sd 0.23) vs 70.95 (sd 0.17)**, +3.2%.

### Shipping posture

`gate_up` defaults ON for **gfx1151 only** — the arch it was measured on — per this repo's rule that a specialisation ships where it was measured. `HIPFIRE_MQ4V2_GATE_UP_TIGHT_GRID` forces either way, so **whoever has the R9700 can measure gfx12 and widen the condition**; the same reasoning should apply there but I cannot claim it.

The `down` contraction stays fully opt-in (`HIPFIRE_MQ4V2_DOWN_TIGHT_GRID`). With ninepath D4 default-on for qt44, that kernel is not on the decode path, so there was nothing to measure it against here.

### Worth contrasting with the previous commit

The ninepath port was proposed from qt13's published "+0.8% serve battery" figure and measured **neutral**. This one came from reading what the kernels actually index and measured a **clean, non-overlapping win**. Same session, same hardware — the difference was starting from the code rather than from someone else's number.

Open leads, none measured: the down contraction (needs ninepath off, or a shape where the expanded kernel is live), the concurrent/multiturn arm for ninepath, and gfx12 for the gate_up contraction.

cat5edopeHA:
**Retested at `26c0dcb6` on gfx1201 (R9700, 32 GB): AR is fixed.**

Same isolated rig as before (dedicated user/GPU0, published artifact from `hipfire-models/ornith1.5-35b-a3b`, greedy, temp 0, thinking off, q8 KV contiguous, max_seq 16384):

| commit | flags | pp t/s | tg t/s | output |
|---|---|---:|---:|---|
| `77e63e20` | 3-flag workaround combo | 3.0 | 65.1 | correct |
| `95bb365f` | default | 947.8 | 113.3 | degenerate (repetition loop) |
| `95bb365f` | all variants tested | 114–948 | 113–127 | degenerate |
| **`26c0dcb6`** | **default** | **1100–1137** | **114.7–116.2** | **correct** |

Quality at `26c0dcb6`: arithmetic ✓ (17×23=391, 31×47=1457), syllogism ✓, code ✓, coherent long-form prose ✓, deterministic across runs.

One data point that may be useful for your dispatch analysis: on `95bb365f` (before the D4 port) the shipped artifact produced fluent-but-degenerate output on **every** flag configuration, including the previously-correct workaround combo. At `26c0dcb6` it's clean with zero flags. That's consistent with the new qt44 ninepath arm now intercepting dispatch *before* the qt13 fallthrough — i.e. on `95bb365f` the fallthrough was live for this shape, and your "fallthrough never taken" probe result makes sense if it was run with `065fcff4` already in the tree. Flagging only because the two commits are adjacent and it affects bisecting any future reports.

**Ninepath ON/OFF confirmed on gfx1201:** no measurable tg delta (~116 t/s both arms), matching your gfx1151 measurement. `HIPFIRE_QWEN35_NINEPATH=0` propagates fine.

**Two remaining gaps, both pre-existing:**

1. **MTP still broken at `26c0dcb6`.** The published `.mtp` sidecar is qt44; the loader still accepts only MQ4G256(v1)/Q8_F16/F16/F32:
   ```
   thread 'main' panicked at crates/hipfire-arch-qwen35/src/mtp_head.rs:1193:18:
   .mtp tensor 'eh_proj': unsupported quant_type=44 (mtp_extract emits MQ4G256=13, Q8_F16=3, F16=1, F32=2)
   ```
   Same panic as `77e63e20`. Happy to retest the moment qt44 sidecar support lands.

2. **Vision isn't servable from the published artifacts.** The `.vl` (907 MB) is a bare tower dump (333 tensors, `model.visual.blocks.*`, full config JSON) — serving it directly panics (`embed_tokens not found`, load.rs:2467), and there's no `.vl` sidecar wiring in the loader. Text-only trunk errors "model has no vision encoder". It looks like vision needs the baked-in `-vl.mq4` trunk (23.75 GB per the PR table), which isn't published yet (`hipfire-models/ornith1.5-35b-a3b-vl` → 401). If you publish it, I'll run the vision ladder same-day.

Standing offer unchanged — R9700/gfx1201 oracle here, will rerun AR → MTP → vision on each push.

nwoolmer:
Follow-up on the decode hunt: one more confirmation of the gate_up win, and one measurement I am **not** going to report a result for.

### gate_up contraction: third confirmation, and it is robust to contention

| arm | median | sd | range |
|---|---:|---:|---|
| TIGHT (default) | 73.40 | 0.25 | 72.80–73.70 |
| WIDE (forced off) | 71.35 | 0.32 | 70.80–71.90 |

+2.05 tok/s (**+2.9%**), ranges non-overlapping. Three independent confirmations now: +2.8% (4 alternations × 12), +3.2% (10 runs each, quiet box), +2.9% (10 runs each, contended box). The effect is ~8× the within-arm sd every time.

### Down path (ninepath vs expanded ± contraction): inconclusive, not reported

I ran a three-way A/B — ninepath D4 vs expanded+wide vs expanded+tight — and the result is **noise-limited and I am discarding it**. Per-arm sd came out 1.2–2.4 with ranges spanning 64.9–74.3, against 0.17–0.32 on the gate_up runs.

Cause identified rather than assumed: a `java` process at 62% CPU plus desktop load (load average ~4) on the box during that window. The GPU itself was idle at 48 °C with no competing hipfire processes, so this was host-side contention, not thermal.

The medians would have let me rank the three arms. They differ by less than the within-arm spread, so that ranking would have been an artifact. Recording the non-result instead.

There is a real signal in the *shape* of that failure, though: a ~3% effect punches through this box's noise floor and the down-path differences do not. That is consistent with the earlier finding that ninepath measured neutral — the down-path options appear to be genuinely close together, below what this machine can resolve while loaded.

### Where that leaves the decode work

Landed and measured: the gate_up launch contraction, +2.9–3.2% on the shipped Ornith, default-on for gfx1151 only.

Open, all needing a quiet box or hardware I do not have:
- **Down path ranking** (ninepath vs contracted expanded) — needs an idle machine; the arms are close enough that contention decides the outcome.
- **`HIPFIRE_MQ4V2_DOWN_TIGHT_GRID`** — stays opt-in, unmeasured, and only reachable with ninepath off.
- **gfx12 for the gate_up contraction** — `HIPFIRE_MQ4V2_GATE_UP_TIGHT_GRID=1` on an R9700. The same argument applies (the kernel guards `row >= mi` regardless of arch) but I will not claim it without the measurement.
- **Ninepath under concurrency** — the workload its qt13 figure came from, still untested.

### The transferable bit

The win did not come from porting another format's kernel or another arch's tuning. It came from comparing what each launcher requests against what its kernel actually indexes — gate_up asked for `2·mi` workgroups to fill `mi` rows, down asked for `m` to fill `m/4`. Both were mine, inherited by mirroring a conservative default rather than reading the guard. Worth a look at other launchers on the same principle; I checked the obvious neighbours on this path and found no further mismatch, but the sweep was not exhaustive.

cat5edopeHA:
## gfx1201/RDNA4 measurement of the gate_up grid contraction — confirmed win

We ran the `HIPFIRE_MQ4V2_GATE_UP_TIGHT_GRID=1` A/B on gfx1201 (Radeon AI PRO R9700, 32 GB) against the shipped Ornith 1.5 qt44 artifact at `4a31e6ba`, per the invitation in the shipping-posture notes.

### Setup
- Isolated serve (dedicated user, GPU0, prod on GPU1 untouched), config: q8 KV contiguous, max_seq 16384, speculation off, AR only.
- Protocol: alternating server-restart blocks OFF → ON → OFF → ON, 5 measured runs each after a discarded warmup, single stream, greedy (`temperature=0`), fixed prompt, 512 output tokens, reading `timings.decode_tok_s`.
- Note for anyone replicating: the developer env is read by the daemon process at launch — passing it as a request header does nothing. Each arm needs its own serve restart.

### Results

| block | WIDE (default) | TIGHT (=1) | delta |
|---|---|---|---|
| 1 | 124.52 ±0.08 | 130.00 ±0.07 | **+4.40%** |
| 2 | 123.60 ±0.07 | 126.92 ±0.05 | **+2.69%** |

No overlap between arms in either block (WIDE max 124.6 vs TIGHT min 126.9). sd ≤0.09 tok/s across all four blocks. Overall +3.55% vs paired baseline.

### Correctness under the flag
Quality smokes pass with the contraction enabled: arithmetic (31×47 = 1457), syllogism validity ("Yes"), coherent prose. Same parity-oracle-backed value-preservation story as your gfx1151 run.

### Suggestion
This matches the reasoning in your shipping posture: same launch-shape contraction, semantically target-neutral, now measured clean on RDNA4. We'd support widening the default-on condition to include gfx1201.

Happy to provide raw logs or rerun with more alternations if useful.

ghazni101:
## gfx1100 (RX 7900 XTX) parity validation — all PASS

Ran the lab-gated GPU parity oracle (`crates/rdna-compute/examples/mq4v2_moe_parity.rs`) from this branch (`16cc144a`) on gfx1100 / ROCm 7.14.0.

**Build:** git archive of the branch → container built on `rocm-dev:7.14.0`, `cargo build --release -p rdna-compute --example mq4v2_moe_parity --features lab`.

**Run:** `docker run --rm --device /dev/kfd --device /dev/dri --group-add 44 --group-add 993` with `ROCM_PATH` / `LD_LIBRARY_PATH=/opt/rocm/lib`.

### Results

| check | rel_l2 | cos | tol | result |
|---|---:|---:|---:|---|
| pack→dequant round trip (host) | 9.433e-3 | 0.99996 | 2e-2 | PASS |
| gate_up `y_gate` | 4.767e-7 | 1.0 | 1e-5 | PASS |
| gate_up `y_up` | 5.097e-7 | 1.0 | 1e-5 | PASS |
| down `expert_outputs` | 3.581e-7 | 1.0 | 1e-5 | PASS |
| grouped `Y` (gfx11 `_k2` source) | 2.924e-4 | 1.0 | 2e-2 | PASS |
| ninepath D4 fused down+combine | 1.725e-6 | 1.0 | 1e-5 | PASS |
| **6 grid-swapped negative controls** | 0.65–1.6 | — | must disagree | **all fired** |

Every rel_l2 matches the gfx1151 figures in the PR description to 3 significant figures.

### What this covers on gfx1100

- **gfx11 grouped MoE GEMM** (`_k2` source) — the arch-selecting launcher picked the gfx11 path.
- **Both indexed MoE GEMVs** (gate_up, down) — arch-generic source.
- **Ninepath D4** (fused down+combine).
- **Negative controls** — all 6 grid-swapped references correctly disagreed, confirming the disjoint-halves fixtures are discriminating.

### What this does NOT cover

- **gfx12 WMMA intrinsic swap + C-output mapping** in the grouped kernel — the one axis no local test can reach. Requires gfx1200/gfx1201 (RDNA4/R9700). On gfx1100 the launcher selects the gfx11 `_k2` source, not the new `gfx12` source.
- Full serve test against the shipped `ornith-1.5-35b-a3b.mq4` artifact (not downloaded); the parity oracle uses synthetic disjoint-halves fixtures as designed.

Raw output:

```
host self-test (no GPU):
  pack->dequant round trip                     rel_l2=9.433e-3     cos=0.999956   tol=2e-2      PASS
  grid-swapped dequant (negative control)      rel_l2=1.409e0      (must exceed 1e-1    ) PASS

GPU dev 0: gfx1100 (25.8 GB VRAM, HIP 7.14)
arch=gfx1100

gate_up  M=128 (mi=64) K=512 k_top=8 n_exp=8
  gate_up y_gate                               rel_l2=4.767e-7     cos=1.000000   tol=1e-5      PASS
  gate_up y_up                                 rel_l2=5.097e-7     cos=1.000000   tol=1e-5      PASS
  gate_up vs grid-swapped ref (negative control) rel_l2=7.609e-1     (must exceed 1e-2    ) PASS

down     M=64 K=512 k_top=8 n_exp=8 batch=1
  down expert_outputs                          rel_l2=3.581e-7     cos=1.000000   tol=1e-5      PASS
  down vs grid-swapped ref (negative control)  rel_l2=6.486e-1     (must exceed 1e-2    ) PASS

grouped  M=32 K=512 m_total=16 (arch-selecting launcher)
  grouped Y                                    rel_l2=2.924e-4     cos=1.000000   tol=2e-2      PASS
  grouped vs grid-swapped ref (negative control) rel_l2=1.597e0      (must exceed 1e-1    ) PASS

ninepath M=64 K=512 k_top=8 n_exp=8
  ninepath d4 fused down+combine               rel_l2=1.725e-6     cos=1.000000   tol=1e-5      PASS
  ninepath vs grid-swapped ref (negative control) rel_l2=1.054e0      (must exceed 1e-2    ) PASS

mq4v2_moe_parity: all checks PASSED
```

Kaden-Schutt:
Local PR-queue disposition: the current GitHub head is still `16cc144a`, exactly the second parent already integrated locally—there are no newer Ornith commits to replay. The local branch retains the Ornith SKU/aliases, MTP/VL metadata, gfx11+gfx12 qt44 grouped MoE kernels, indexed GEMVs, ninepath D4, tight-grid change, parity example, quantizer fixes, and the renamed `hipfire-models/ornith-1.5-35b-a3b` repo id.

Maintainer realignments on top: restored `_effort_native_tag` and Ornith effort-native classification; added canonical dense/MoE arch mappings; preserved Hyper and every master MQV2 kernel const; and fixed model lookup so an exact on-disk stem outranks a registry alias while separator-only legacy spellings still resolve.

Fresh verification after hipx restart: registry regeneration/check and Ornith alias/arch/dispatch/quantizer tests pass; the complete `mq4v2_moe_parity` oracle passes on gfx1100 for gate-up, down, grouped GEMM, and ninepath, including negative controls. We do not have the published Ornith trunk locally, so I am not claiming end-to-end Ornith serve/MTP/VL. The existing MTP qt44 sidecar and published `.vl` usability gaps remain yours to close. This is local preflight only: nothing has been pushed and nothing has been merged to `master`.

PR #612, opened 2026-08-21
Fix hipfire-quantize compilation on Windows

noctrex:
## Summary

`hipfire-quantize` fails to compile on Windows because `memmap2::Advice` is exported unix-only (memmap2 0.9, `#[cfg(unix)]`), but `hessian_io.rs` imports it unconditionally:

```
error[E0432]: unresolved import `memmap2::Advice`
  --> crates/hipfire-quantize/src/hessian_io.rs:23:15
```

This also breaks the documented whole-workspace build (`cargo build --release --workspace`) on Windows.

## Fix

- Gate the `Advice` import behind `#[cfg(unix)]`.
- Drop the stray `let _ = Advice::Sequential;` lint-silencer; it referenced the broken import and is unnecessary once the import is gated.

The `mmap.advise(...)` call was already `#[cfg(unix)]`-gated, so there is no behavior change on unix. On Windows the sequential-access hint is simply skipped; `Mmap::map` itself is fully supported there. The remaining diff churn in the file is `rustfmt` output required by the changed-file fmt gate.

## Testing

- Windows 11, `x86_64-pc-windows-msvc`, rustc 1.97.1, memmap2 0.9.11:
  - before: `cargo build --release -p hipfire-quantize` -> E0432
  - after: builds clean; `hipfire-quantize --help` runs
- `cargo clippy -p hipfire-quantize`: no warnings in the changed file (crate's other warnings pre-exist on all platforms)
- Whole workspace now completes on Windows with no `--exclude`

LLM Disclosure: Yes, used GLM-5.3 to help along

### DIFF
--- a/crates/hipfire-quantize/src/hessian_io.rs
+++ b/crates/hipfire-quantize/src/hessian_io.rs
@@ -20,7 +20,9 @@
 //! `get(tensor_name_without_dot_weight_suffix, 0)` per MQ4G256 tensor.
 
 use byteorder::{ByteOrder, LittleEndian};
-use memmap2::{Advice, Mmap};
+#[cfg(unix)]
+use memmap2::Advice;
+use memmap2::Mmap;
 use std::collections::HashMap;
 use std::fs::File;
 use std::path::Path;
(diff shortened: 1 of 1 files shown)


Kaden-Schutt:
Superseded by the narrower hessian_io Windows fix in #615. Please preserve attribution when #615 lands; the remaining hfqm.rs Advice path is tracked separately.

PR #613, opened 2026-08-21
fix(quantize): gemma4 GGUF config + tensor-name translation

noctrex:
## Summary

GGUF -> HFQ quantization of gemma4 models produced unloadable artifacts.
`hipfire run` failed with `load failed: gemma4: missing layer_types array`.
The safetensors path (the one exercised in #439) is unaffected: HF
`config.json` and HF tensor names carry through verbatim. The GGUF path
goes through two generic Llama-assumed translation layers that were never
gemma4-aware.

LLM Disclosure: Yes, used GLM-5.3 to help along

## Fix

Two translation gaps, both gated on `arch_str == "gemma4"` (byte-identical
path for every other arch, regression-tested):

1. `config_json_from_gguf` (calibration.rs): gemma4 stores its layout as
   arrays and dual SWA/full keys the scalar reader misses or misreads.
   Added a gemma4 arm that emits:
   - `layer_types` from `attention.sliding_window_pattern` (bool array;
     GGUF has no `layer_types` key, loader hard-requires it)
   - `num_key_value_heads` / `num_global_key_value_heads` from the
     per-layer `head_count_kv` array (8 sliding / 1 full; scalar read
     silently fell back to n_heads)
   - `head_dim` = `key_length_swa` (256) and `global_head_dim` =
     `key_length` (512); the generic path wrote the full 512 into both
   - `rope_parameters` from `freq_base_swa` / `freq_base` (10k default
     sliding, 1M proportional + partial_rotary_factor 0.25 on full)
   - `attention_k_eq_v: true` when full layers ship no `attn_v` tensor
     (12B-class checkpoints; detected from the tensor table)
   - `sliding_window`, `num_kv_shared_layers`,
     `hidden_size_per_layer_input`, `final_logit_softcapping`

2. `gguf_to_safetensors_name` (calibration.rs): the Llama slot map
   assumed two norms per layer; gemma4 has four sandwich norms + a
   per-layer scalar. Without this, `ffn_norm` landed in
   `post_attention_layernorm` (wrong weights), `pre_feedforward_layernorm`
   was never emitted (guaranteed load failure), and `layer_output_scale`
   never became `layer_scalar` (loader reads it without a `.weight`
   suffix).

Tests parse the generated config through the real loader parser
(`Gemma4Config::from_metadata_json`, added as a dev-dependency of
hipfire-quantize) using metadata + tensor tables mirroring the production
gemma-4-12b GGUF.

Stacks on #612 (Windows build fix, cherry-picked so this branch compiles
on Windows; drop that commit if #612 merges first).

## Testing

- `cargo test -p hipfire-quantize gemma4`: 6 tests pass (layout fields,
  loader-parser admission, k_eq_v both ways, name translation, llama
  regression)
- Full crate suite: 117 pass; `glimmer_self_attn_*` and
  `sp4_overlay_to_sp3_load_end_to_end` fail/flake, both reproduce on
  clean master (order/env-dependent, pre-existing)
- End-to-end: `gemma-4-12b-it-heretic-BF16.gguf` (BF16, 23.8 GB) ->
  `--format hf4` -> loads and generates coherently on gfx1100
  (`HIPFIRE_GEMMA4_GRAPH=0`, see below), e.g. correct
  `def reverse_string(text): return text[::-1]` with explanation

Out of scope, found while testing: gemma4 hipGraph decode emits token
soup on gfx1100 with default `HIPFIRE_GEMMA4_GRAPH=1` (attractor
signature as documented in #439); eager path is byte-coherent. Will file
a separate issue.

### DIFF
--- a/crates/hipfire-quantize/Cargo.toml
+++ b/crates/hipfire-quantize/Cargo.toml
@@ -32,6 +32,10 @@ tempfile = "3"
 # integration test lives in-module in reap_overlay.rs (it needs crate-internal
 # write_hfq / quantize_* access that a tests/ target can't reach).
 hipfire-runtime = { path = "../hipfire-runtime" }
+# Test-only: parse `config_json_from_gguf` output through the REAL gemma4
+# loader parser (Gemma4Config::from_metadata_json) — the config-emission /
+# loader-admission boundary is where the GGUF gemma4 gap sat.
+hipfire-arch-gemma4 = { path = "../hipfire-arch-gemma4" }
 # Test-only: the MQ*-G256-GL codebooks (GL_CB2/GL_CB3) and SoA group sizes are
 # duplicated between this encoder and the runtime (the kernels take the codebook
 # as scalar args; it is NOT stored in the .hfq). `gl_codebooks_match_runtime` /
@@ -39,7 +43,6 @@ hipfire-runtime = { path = "../hipfire-runtime" }
 # identical, so a drift becomes a red test instead of a silent accuracy loss.
 # dev-dependency only — the quantizer still builds and runs with no GPU/ROCm.
 rdna-compute = { path = "../rdna-compute" }
-
 # ---------------------------------------------------------------------------
 # Archived research probes.
 #
--- a/crates/hipfire-quantize/src/calibration.rs
+++ b/crates/hipfire-quantize/src/calibration.rs
@@ -107,7 +107,7 @@ pub(crate) fn is_gguf_input(p: &Path) -> bool {
 /// Returns None for tensors that don't have a known safetensors equivalent
 /// (we then keep them under their GGUF name; the future loader can decide
 /// what to do, or they're skipped).
-pub(crate) fn gguf_to_safetensors_name(gguf_name: &str) -> Option<String> {
+pub(crate) fn gguf_to_safetensors_name(gguf_name: &str, arch_str: &str) -> Option<String> {
     // Top-level tensors.
     match gguf_name {
         "token_embd.weight" => return Some("model.embed_tokens.weight".to_string()),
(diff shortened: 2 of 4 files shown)


Kaden-Schutt (changes requested):
The Gemma4 GGUF config and tensor-name translation are still missing, but this branch includes the now-superseded #612 hessian commit and predates the MQ V2/pipeline changes. Please rebase onto current master, drop the hessian commit in favor of #615, reapply the Gemma4 config/name mapping with the current pipeline_gguf signatures, and rerun the focused parser/translation tests.

Kaden-Schutt (changes requested):
Rebase is clean, but Gemma4 E-series support is incomplete: parse the per-layer `gemma4.feed_forward_length` array, emit base intermediate_size plus use_double_wide_mlp, and add an E2B-shaped test. Also map the top-level per-layer-input GGUF tensors required by E2B/E4B (`embed_tokens_per_layer`, `per_layer_model_projection`, `per_layer_projection_norm`) with focused translation coverage.

Kaden-Schutt:
Local PR-queue disposition: I integrated HEAD `d755af78`, retaining the Gemma 4 GGUF config and tensor-name translation. I added a maintainer realignment so translation is gated by the canonical `arch_id == 13`, not the pre-unification `arch_str == "gemma4"`, and threaded that through the calibration/GGUF pipeline with extra engagement tests.

Unit coverage passes locally. We still had no Gemma 4 GGUF fixture for end-to-end GGUF→HFQ→load/generate, and the existing review request around E-series `feed_forward_length` arrays / per-layer-input tensors remains open. This is local preflight only: nothing has been pushed and nothing has been merged to `master`.

PR #616, opened 2026-08-22
feat(gfx11): add packed-MQ4 and optional CK prefill backend, 2x prefill compared to default pathway, 1221.5 Tok/s for qwen 3.8 27b dense

HUSRCF:
## Summary

Add the validated gfx1100 packed-MQ4 prefill stack and an optional, fail-closed FlashAttention CK sidecar for the Qwen3.6/3.8 27B Asym3-K/Q8-V path. The bundle tooling provides a relocatable prebuilt deployment path, so users do not need to compile CK locally.

The optimized routes are restricted to the validated RDNA3 shapes and retain native fallbacks. CK additionally requires the `flash-attn-ck` Cargo feature and an explicit sidecar path; graph capture, tree masks, multi-slot attention, generic llama paths, unsupported shapes, and unavailable sidecars remain native.

The branch is based directly on `beta@80a572c8`. The implementation was migrated as separate reviewable commits covering zero-correction fusion, K32 probes, group128 execution, fused SwiGLU packing, activation reload reduction, packed-MQ4 primitives, FP16 FFN intermediates, production routing, and benchmark evidence.

## Scope

- Radeon Pro W7900 / exact `gfx1100`, ROCm 7.14
- Qwen hybrid prefill, including Hq=24, Hkv=4, D=256 attention
- packed-MQ4 projection routes plus causal contiguous-prefix CK attention
- prefill only; decode and speculative verification remain native
- no prebuilt binary is committed to git

On official `beta@80a572c8` with the complete packed-MQ4 production flag set, three alternating W7900 process pairs at PP8192 measured native attention `748.4 tok/s` versus staged CK attention `1221.5 tok/s`: a **1.6321x paired-median speedup**, with 3/3 positive pairs. Each process ran three prefill trials and reported its in-process median. Every pair produced identical greedy token IDs, while decode remained neutral at `35.0-35.2 tok/s`.

The full-stack harness, raw rows, artifact hashes, and reproduction command are committed under `experiments/flash-attn-ck-sidecar/results/beta_w7900_pp8192_fullopt_ab_20260823/`. The earlier CK-only migration result (`593.7 -> 865.8 tok/s`) remains under `results/beta_w7900_pp8192_ab_20260823/` for attribution.

## Which crates does this touch?

- [x] `kernels/` (gfx11 packed-MQ4 kernels and optional CK sidecar)
- [x] `crates/rdna-compute` (execution primitives, optional loader, dispatch)
- [x] `crates/hipfire-runtime` / `crates/hipfire-dispatch` (Qwen prefill routing)
- [x] `crates/hipfire-cli` / `crates/hipfire-daemon` (feature forwarding)
- [x] docs / scripts / benchmark evidence

No Architecture trait change.

## Test plan

- [x] `cargo check --locked -p hipfire-cli --features flash-attn-ck`
- [x] `cargo check --locked -p rdna-compute --example test_batched_attn_slots --features flash-attn-ck`
- [x] `cargo build --release --locked -p hipfire-daemon --features deltanet,flash-attn-ck`
- [x] `cargo build --release -p hipfire-runtime --example bench_qwen35_mq4 --features deltanet,flash-attn-ck`
- [x] gfx1100 D256 dense CK sidecar source build on ROCm 7.14
- [x] D256 CPU-reference smoke: MHA, MQA, and GQA; causal and non-causal; default and non-default stream
- [x] build-capability smoke verifies that only generated D64/D128/D256 dimensions are advertised
- [x] staged quantized sidecar source build and ABI smoke (ABI=1, workspace=3,145,728 bytes, staged=11,534,336 bytes)
- [x] bundle package, checksum, clean-prefix install, `$ORIGIN` dependency, and launcher smoke
- [x] bundle contract tests: 4/4
- [x] optional Rust loader tests: 6/6
- [x] Python suite: 836 passed
- [x] Rust core tests: 169 passed; Qwen prefill tests: 13 passed; CLI/TUI tests: 153 passed
- [x] full-stack W7900 PP8192 A/B: 3/3 positive pairs, exact greedy token-ID agreement
- [ ] full `./scripts/no-gpu-ci.sh`: reaches the final legacy suite, then current beta reports two unrelated baseline failures (`ensure_binaries` expects the pre-split daemon cargo argv; change-gate lists three moved/dead pflash paths). This PR does not touch those files.

## Packaging

The bundle contains a CK-enabled CLI and daemon, relocatable dense/staged sidecars, and matching gfx1100 code-object cache. The installer verifies external and internal SHA256, manifest/ABI, sidecar dependencies, kernel inventory, and visible GPU before atomically switching a versioned symlink. Unsupported systems are rejected before activation.


Kaden-Schutt (dismissed):
Still relevant, but this beta-based mega-diff collides directly with the landed MQ V2 gemm/prefill policy. Please rebase on master, split the CK FlashAttention sidecar from packed-MQ4 routing, and remeasure the packed path on qt44 before review.

Kaden-Schutt (changes requested):
Reassessment: this work is substantially more important than my initial review suggested. Its strongest value is not a standalone gfx1100 Asym3/Q8 sidecar or one packed-MQ4 result; it is the reusable infrastructure for a first-class CK runtime inside hipfire.

Please preserve that work and reshape it against current master as follows:

1. Make CK an integral, optional dlopen runtime backend with a versioned capability ABI, caller-owned workspace, stream-ordered launches, and mandatory native fallback. Inference hosts must not require a pinned/full ROCm SDK or CK source tree; prebuilt exact-arch artifacts may retain build-farm recipe pins and manifests.
2. Put backend selection after the existing KvTierPlan/AttentionFamily resolution so Qwen callers remain backend-agnostic. Do not bury CK only inside the Asym3 method or force global graph/KV settings. Capture, tree, slots, and unsupported shapes remain native until their ABI is explicitly supported.
3. Generalize the runtime layout contract beyond the current hardcoded 100/272-byte Asym3-K/Q8-V path. Required capability cells are Q8-K/Q8-V, Asym-K/Q8-V, FWHT-K/Q8-V, and FWHT-K/Lloyd-V, with exact-arch artifacts/capability probes for gfx1100, gfx1151, and gfx1201.
4. Keep the packed projection stack as a consumer of that reusable infrastructure, but adapt it to current MQ4V2 qt44 dual-FP16 metadata and the restored MW/BT dispatch policy. It should not be coupled to obsolete Asym3 defaults or to the CK loader lifecycle.
5. Rebase on current master and split the reusable CK runtime/ABI, MQ4V2 packed projections, and immutable benchmark evidence into reviewable commits or PRs. Preserve the existing gfx1100 evidence as an existence proof, then remeasure matched native-vs-CK cells on current MQ4V2 without packed-GEMM confounding.

This remains CHANGES_REQUESTED because the beta mega-diff cannot merge safely as-is, but the direction is now explicitly to retain and elevate the work as reusable multi-format, multi-arch hipfire infrastructure rather than discard it as a narrow sidecar.

HUSRCF:
Thank you for the reassessment and for laying out the integration boundary so clearly. I agree that keeping this as a beta-based combined CK/packed-MQ4 diff would make the useful infrastructure unnecessarily difficult to review and unsafe to merge against the current MQ4V2 policy.

I have preserved this branch and its W7900 measurements as immutable existence evidence, and started a clean branch from current `master@336dd077`: [`HUSRCF:feat/ck-runtime-abi-master`](https://github.com/HUSRCF/hipscope/tree/feat/ck-runtime-abi-master). The first two isolated commits contain only the optional `dlopen` loader/reference sidecar and an ABI v2 capability/workspace contract. The latter adds exact-architecture capability cells, explicit K/V format identifiers, caller-owned workspace negotiation, stream-ordered launches, strict ABI/layout validation, and recoverable load/support errors so callers can retain mandatory native fallback. The CK ABI unit tests currently pass 5/5 on current master.

I will continue the work as separate, current-master changes: first the reusable CK runtime and post-`KvTierPlan`/`AttentionFamily` backend selection, then individual multi-format capability cells and exact-arch artifacts, and only afterward an MQ4V2/qt44 packed-projection change with matched benchmarks. The old Asym3-specific routing and packed-GEMM stack will not be carried into the runtime PR as coupled policy.

I am closing this PR to avoid asking reviewers to track two incompatible integration shapes. The branch, benchmark evidence, and discussion will remain available as the existence proof you requested. Thanks again for pointing the work toward a reusable first-class backend rather than a narrow sidecar.

PR #617, opened 2026-08-22
perf(loader): UMA-gated eviction + parallel warmer + zero-copy uploads — 2-3.4x faster dense loads, ~6x faster A3B MoE loads on dGPUs; fix sequential sampler seed

ghazni101:
## Summary

hipfire's model→VRAM load path was leaving most of its bandwidth on the
table for three stacked reasons, each found by measurement:

1. **UMA-only page-cache eviction applied everywhere** — every tensor
   read ended with `posix_fadvise(DONTNEED)`, forcing a full disk
   re-read of the model on *every* load. Correct on unified-memory APUs
   (Strix Halo etc.), pure loss on discrete GPUs.
2. **Serial per-tensor pread** — a single `FADV_WILLNEED` does not
   saturate this box's btrfs md0 (~1.7 GB/s serial vs ~4.5 GB/s with
   concurrent readers), so cold loads serialized behind disk latency.
3. **Heap staging copy per tensor** — `pread` into a fresh Vec, then
   `hipMemcpy` from that copy: page-cache → heap → VRAM instead of
   page-cache → VRAM.

Three commits, one per cause:

- `9842657f` **gate HFQ page-cache eviction on UMA**: `HfqFile`
  gains `evict_page_cache` (default `true` = historical/UMA behavior);
  the qwen35 carrier disables it on non-UMA devices (both pp=1 and
  pp>1 paths); `HIPFIRE_PAGE_EVICTION=1` restores old behavior;
  `arch_is_uma()` classifies gfx1103/1150/1151/1152. deepseek4's
  OOM-sensitive pread+fadvise lifecycle is deliberately untouched.
- `9d51c825` **parallel cache warmer + mincore gate**: 4 worker threads
  pread the data region into the page cache (16 MiB chunks) while the
  loader uploads; a guard stops them on drop. Skipped entirely when
  `mincore` shows ≥90 % of pages already resident.
- `c6edb843` **zero-copy uploads**: qwen35 keeps its mmap alive on
  dGPUs and reads tensors as borrowed slices (`tensor_data`) so
  `hipMemcpy` DMAs straight out of page-cache pages; pread fallback
  preserves UMA behavior.

A fourth idea was tried and **reverted**: blocking `MADV_POPULATE`
before the upload loop measured identical totals (it relocates the same
soft-fault work into one serial stall). Documented in `prepare()`'s
comment so it isn't re-added blindly.

New tooling: `scripts/bench_model_load.py` measures hipfire and the
ROCmFPX llama.cpp fork with identical windows (hipfire: daemon spawn →
`{"type":"loaded"}` protocol response; rocmfpx: spawn → end-of-load log
marker, process killed at marker so prompt eval/gen are excluded).
Microbenchmarks: `h2d_mmap_bench.rs`, `h2d_microbench.rs`.

## Round 2: A3B MoE loads + a latent eviction leak (follow-up campaign)

Deploying the branch to the Ornith-1.5-35B-A3B mq4r serving container
exposed three more issues, found by phase-profiling real daemon loads:

4. **Zero-copy missed the MoE expert sweep** (`0493003f`) — the
   packed-MQ4 loader (`try_load_packed_mq4_experts`) still pread all
   512 experts per blob into fresh heap Vecs before `hipMemcpy`. This
   sweep was ~28 s of every ~31 s load on the A3B model. Fix:
   `HfqFile::data_range()` + an in-file contiguity probe — when all
   experts lie back-to-back and pages are resident, upload straight
   out of one borrowed mmap slice; non-contiguous/overlay/UMA keep the
   old paths.
5. **Parallel reader evicted its own pages** (`761de5d9`) —
   `read_job()` called `fadvise(DONTNEED)` after *every* segment,
   unconditionally, so any load routed through the lane-parallel
   reader re-read the whole model from disk (~0.7 GB/s) on every
   restart. This contradicted the carrier's dGPU keep-resident policy;
   `hfq.evicts_page_cache()` is now threaded through
   `read_jobs_from_path`/`read_job`.
6. **Per-layer residency probing** (`d2072ddf`) — the first cut of the
   zero-copy gate called `mostly_page_cached()` (fresh map+mincore of
   the full multi-GB file) once per layer ≈ 17 s per load of pure
   checking. Memoized per process.

Also riding in this push (`1cd7a890`): **sequential serve requests
seeded the sampler RNG from a compile-time constant** (`0x13579BDF`),
making temp>0 output byte-identical across requests for the same
prompt on the noslots path (batch scheduler already keyed entropy per
attempt). `generate()` now takes a `request_seed`; the daemon derives
it via `batch_rng_for_key(AttemptKey)`. Greedy temp-0 stays fully
deterministic — verified byte-identical pre/post change.

## Round 3: swap correctness + warmer scaling

Two follow-ups from continued load/swap testing on gfx1100:

7. **Stale residency memo across model swaps** (`05ea5cb8`) — the
   per-process memo from `d2072ddf` was correct within one load and
   wrong across loads: swapping from a warm model A to a cold model B
   left B on the zero-copy borrowed-slice path with its pages absent,
   so H2D soft-faulted serially off disk (~0.25 GB/s — exactly the
   mode `0493003f`'s parallel-pread fallback exists to avoid). The
   memo is now keyed to the file instance, not the process. Also drops
   a `debug_assert` in `tensor_data()`: UMA `prepare()` drops the mmap
   by design and returning `None` is the documented fallback contract;
   debug builds on APUs tripped it on every tensor read.
   Measured (9B warm → 27B cold swap, same daemon process): 7466 ms
   before vs **4349 ms** after; fresh-process controls 3616 / 3719 ms.
   Audit tooling committed alongside (gpu-coord lock held by caller):
   `scripts/audit_load_ab.py` (interleaved master/branch load A/B with
   mincore residency probes + loader stderr trace capture),
   `scripts/audit_swap_memo.py` (reproduces the stale-memo swap
   regression), `scripts/audit_parity_check.py` (greedy byte-parity
   across two daemons).
8. **Warmer scaled 4 → 8 threads** (`c9cff760`) — standalone
   time-to-99%-resident for a 5.3 GB mq4 after drop_caches on btrfs
   md0: 4 threads 3.69 GB/s, 8 threads **4.58 GB/s**, 12 threads
   5.00 GB/s (plateau = array sequential ceiling); 8 keeps most of the
   win without a small army of threads competing with the upload
   thread. Two more ideas investigated and **rejected with
   measurements**: chunked `posix_fadvise(WILLNEED)` (prototype
   segfaulted under repeated drop_caches runs; not pursued) and
   pre-init `--warm-file` overlap (net-zero end-to-end: the early
   warmer shortened the weight sweep 1387 → 856 ms but starved the
   carrier's synchronous header/metadata fetches by ≈ the same amount;
   total flat at ~2.33 s across 3 paired runs). Reverted; revisit only
   if a storage class shows serial dead time overlap can actually hide.

Validation for this round: greedy parity vs master byte-identical
(sha256s in the commit messages); steady-state 9B load median
1546 ms ±2.3 % over 3 fresh-process runs; hfq unit tests green.

## Measurements

gfx1100 (RX 7900 XTX), fresh daemon process per run, medians of ≥5 runs.
Shared-box caveat: another agent benchmarks the same GPU/disk, so
run-to-run noise is ±10–20 %; all deltas below exceed it.

**Isolation microbenchmarks** (`h2d_mmap_bench`, repaired in
`d472dab2`; gfx1100, qwen3.5-4b.mq4 2.47 GB file, 64 MiB chunks):
heap-src H2D 28.1 GB/s · mmap warm-PTEs 28.1 GB/s · mmap POPULATE_READ
27.9 GB/s · mmap genuinely cold (pages dropped pre-map) **2.45 GB/s** —
cold-start H2D is disk-bound on this array; the previously cited
"cold-PTEs 4.0 GB/s" came from a mislabeled fully-populated mapping and
is disregarded.

**End-to-end load, hipfire daemon:**

| model | master | final | speedup |
|---|---|---|---|
| qwen3.5-9b (md5 `296092bf…`) | 4338 ms | **~1600–1900 ms** | ~2.4× |
| qwen3.5-27b (sha256 `ea615949…8576e`) | 9901 ms* | **~3900–4400 ms** | ~2.4× |
| qwen3.8-27b (md5 `2fb2edc2…`) | 12 528 ms | **~3650–3800 ms** | ~3.4× |
| ornith-1.5-35b-a3b.mq4r (18.8 GB, warm cache)** | ~29 000–36 000 ms | **~5 500 ms** | ~6× |

*qwen3.5-27b baseline taken mid-campaign after the 9B fix landed; the
clean pre-fix number for the same lineage is the 3.8-27B A/B above
(master binary built from `80a572c8`, interleaved runs, identical
weights).
**Round 2 measurement: container daemon start → `/v1/models` serving,
interleaved old/new binary under a GPU lock; weight-sweep phase alone
dropped ~28 s → ~2.7 s (host-read 0 ms = borrowed-mmap path engaged).

**Reference point — ROCMFPX llama.cpp fork** loading equivalent weights,
same measurement window:

| model | rocmfpx steady-state median | hipfire final |
|---|---|---|
| Qwen3.5-9B Q4_K_M (md5 `d3842bd7…`) | 2189 ms | **2108 ms** |
| Qwen3.5-27B Q4_K_M (md5 `a34c995b…`) | 5908 ms | **4308 ms** |

Traced phase timings post-fix (27B): embed 77 ms / 1.24 GB
(~16 GB/s), lm_head 26 ms, full 64-layer sweep ~1.1 s (~11 GB/s).
Remaining budget: fixed init (~0.5 s), post-KV tail (~0.6 s), disk state.

## Verification

- `cargo test -p hipfire-runtime`: 550 tests pass at each commit.
- Repo MQ4 speed gates pass at commit time (prefill +86–91 % over
  baseline floor, decode +9 %).
- Coherence eyeballed on decoded output post-fix: 9B (Python
  string-reverse function), 27B (mutex paragraph), 3.8-27B (17×23=391,
  19×21=399 with correct reasoning chains).
- Deployed validation: two production serving containers rebuilt from
  this branch (`ornith-mq4rp`, `qwen36-35b-mq4r`, both retained-PM4
  rigs) and verified through their real HTTP serving path — load,
  greedy determinism, temp>0 sampling variation, and PM4 route
  activation checked end-to-end after each deploy.
- Sampler-seed fix verified by A/B: old binary reproduced byte-identical
  completions 3/3 at temp 0.8; fixed binary produced distinct
  generations 5/5, greedy unchanged.
- All GPU measurements serialized through a gpu-coord flock protocol;
  binary + model md5s recorded in the commit messages.

### Round 4: review-fix validation (2026-08-24, commits `edcc7d64`..`d472dab2`)

All five review blockers addressed:

1. **Non-Unix memo fallback** — `mostly_page_cached_memo()` now has a
   `#[cfg(not(unix))]` variant returning `false` (pread path).
   `cargo check -p hipfire-runtime -p hipfire-loader
   --target x86_64-pc-windows-msvc` passes.
2. **UMA classification replaced with a runtime property** — the
   gfx1103/1150/1151/1152 name table is gone. `Gpu::is_uma()` queries
   `hipDeviceGetAttribute(hipDeviceAttributeIntegrated)` live (ordinal
   pinned at 16 by the header enumeration that reproduces profiler.rs's
   MULTIPROCESSOR_COUNT=63; ordinals have shifted between ROCm releases,
   noted in-source). Query failure degrades conservatively to UMA=true =
   historical eviction behavior. Verified on this box's gfx1100:
   attribute returns 0 → dGPU path taken.
3. **Fallible warmer spawn** — `Builder::spawn` with per-thread failure
   tolerated: fewer warmers degrade disk bandwidth only, zero spawns
   skip warming entirely instead of panicking mid-load.
4. **shared_expert quant policy split out** — the `q8_class_of`
   "shared_expert" arm (and its missing VALID_Q8_CLASSES registration)
   is reverted from this PR; it belongs to the Q8-shared MoE prefill
   branch.
5. **H2D benchmark repaired** — byte accounting now counts actual
   iterations inside each timing window; the mislabeled "cold-PTEs" pass
   (run after MADV_POPULATE_READ) replaced by a genuinely cold phase
   (fadvise-drop before map) plus a separate POPULATE phase. Repaired
   numbers above.

Validation for this round (gfx1100, all under the gpu lock):

- Smoke (qwen3.5-4b.mq4): cold load 1146 ms after fadvise drop, warm
  reload 844 ms; decoded output coherent (long multiplication steps +
  correct string-reverse function).
- Final (qwen3.8-27b.mq4, prompt md5 `1d32df5f12c414d3e34c7b35b6611e6c`):
  cold load 4651 ms / warm load 2643 ms; greedy outputs byte-identical
  across cold and warm loads.
- Load-swap audit (`scripts/audit_swap_memo.py`, 4B→27B): 27B
  first-in-proc 2596 ms vs fresh-process control 2703 ms — no stale-memo
  regression.
- Greedy parity vs base (`18d46f76`, interleaved, pages dropped between
  arms): byte-identical, output sha256[:16] `84d20e12286c9aec` both
  binaries.
- `cargo test -p hipfire-runtime`: 573 pass; MQ4 speed gate: prefill
  +80.6%, decode +9.0% over baseline floor.
- gfx1151 UMA hardware is not available in this rig, so the UMA path was
  NOT re-validated on-device this round; with blocker 2 the classification
  is now hardware-reported rather than predicted from a table, which
  removes the misclassification risk the table carried. An on-device
  gfx1151 run remains open follow-up.

- Attempt ledger across the campaign: five failed/rejected attempts
  (eviction gate initially landed only in the pp>1 path; the
  blocking-populate experiment; the ungated first cut of the zero-copy
  gate, which let cold-start H2D soft-fault serially off disk at
  ~0.25 GB/s — now mincore-gated; chunked `posix_fadvise(WILLNEED)`;
  pre-init `--warm-file` overlap) — each documented with measurements
  rather than hidden (rounds 1–3 above).
- Non-qwen35 archs keep default eviction unless their carrier opts in,
  so no other load path changes behavior by default.
- UMA systems are byte-for-byte on the old code path: mmap dropped,
  fadvise per read, owned Vecs.
- The sampler-seed commit touches behavior beyond loading; it is
  included here because it was developed against and validated on this
  branch's deployment targets. Happy to split it out if reviewers
  prefer.

### DIFF
--- a/crates/hipfire-arch-qwen35/src/qwen35/load.rs
+++ b/crates/hipfire-arch-qwen35/src/qwen35/load.rs
@@ -2946,6 +2946,22 @@ fn contiguous_tensor_span<'a>(hfq: &'a HfqFile, names: &[&str], stride: usize) -
     hfq.data_range(first.data_offset, expect - first.data_offset)
 }
 
+/// Memoized [`HfqFile::mostly_page_cached`] — the probe maps + mincores the
+/// whole multi-GB model file (~0.4 s), far too expensive to run per layer.
+fn model_pages_resident(hfq: &HfqFile) -> bool {
+    use std::sync::atomic::{AtomicU8, Ordering};
+    static CACHED: AtomicU8 = AtomicU8::new(0); // 0 unknown, 1 resident, 2 not
+    match CACHED.load(Ordering::Relaxed) {
+        1 => true,
+        2 => false,
+        _ => {
+            let resident = hfq.mostly_page_cached();
+            CACHED.store(if resident { 1 } else { 2 }, Ordering::Relaxed);
+            resident
+        }
+    }
+}
+
 fn try_load_packed_mq4_experts(
     hfq: &HfqFile,
     gpu: &mut Gpu,
@@ -3022,7 +3038,7 @@ fn try_load_packed_mq4_experts(
     // an evicted cache soft-faults serially inside the copy loop (~0.25 GB/s
     // off disk), while the parallel-pread fallback reads with multiple lanes.
     let zero_copy_ok =
-        matches!((gate_up_contig, down_contig), (Some(_), Some(_))) && hfq.mostly_page_cached();
+        matches!((gate_up_contig, down_contig), (Some(_), Some(_))) && model_pages_resident(hfq);
     let (gate_up_host, down_host) = if zero_copy_ok {
         let (Some(gu), Some(dn)) = (gate_up_contig, down_contig) else {
             unreachable!("zero_copy_ok implies both spans resolved");
--- a/crates/hipfire-arch-qwen35/src/qwen35/load.rs
+++ b/crates/hipfire-arch-qwen35/src/qwen35/load.rs
@@ -2946,22 +2946,6 @@ fn contiguous_tensor_span<'a>(hfq: &'a HfqFile, names: &[&str], stride: usize) -
     hfq.data_range(first.data_offset, expect - first.data_offset)
 }
 
-/// Memoized [`HfqFile::mostly_page_cached`] — the probe maps + mincores the
-/// whole multi-GB model file (~0.4 s), far too expensive to run per layer.
-fn model_pages_resident(hfq: &HfqFile) -> bool {
-    use std::sync::atomic::{AtomicU8, Ordering};
-    static CACHED: AtomicU8 = AtomicU8::new(0); // 0 unknown, 1 resident, 2 not
-    match CACHED.load(Ordering::Relaxed) {
-        1 => true,
-        2 => false,
-        _ => {
-            let resident = hfq.mostly_page_cached();
-            CACHED.store(if resident { 1 } else { 2 }, Ordering::Relaxed);
-            resident
-        }
-    }
-}
-
 fn try_load_packed_mq4_experts(
     hfq: &HfqFile,
     gpu: &mut Gpu,
@@ -3037,8 +3021,8 @@ fn try_load_packed_mq4_experts(
     // Zero-copy only pays when pages are resident: a borrowed-slice H2D from
     // an evicted cache soft-faults serially inside the copy loop (~0.25 GB/s
     // off disk), while the parallel-pread fallback reads with multiple lanes.
-    let zero_copy_ok =
-        matches!((gate_up_contig, down_contig), (Some(_), Some(_))) && model_pages_resident(hfq);
+    let zero_copy_ok = matches!((gate_up_contig, down_contig), (Some(_), Some(_)))
+        && hfq.mostly_page_cached_memo();
     let (gate_up_host, down_host) = if zero_copy_ok {
         let (Some(gu), Some(dn)) = (gate_up_contig, down_contig) else {
             unreachable!("zero_copy_ok implies both spans resolved");
--- a/crates/hipfire-arch-qwen35/src/qwen35/load.rs
+++ b/crates/hipfire-arch-qwen35/src/qwen35/load.rs
@@ -98,6 +98,28 @@ fn qwen35_tensor_data_vec<'a>(
     None
 }
 
+/// Borrowed-first variant of [`qwen35_tensor_data_vec`]: returns the mmap
+/// slice directly when the mapping is alive (dGPU loads keep it, so weight
+/// uploads DMA straight out of page-cache pages with no heap staging copy);
+/// falls back to the owned pread Vec on UMA (mmap dropped there). Same
+/// candidate resolution.
+fn qwen35_tensor_data_cow<'a>(
+    hfq: &'a HfqFile,
+    name: &str,
+) -> Option<(&'a HfqTensorInfo, std::borrow::Cow<'a, [u8]>)> {
+    for candidate in qwen35_tensor_name_candidates(name) {
+        if let Some((info, data)) = hfq.tensor_data(&candidate) {
+            return Some((info, std::borrow::Cow::Borrowed(data)));
+        }
+    }
+    for candidate in qwen35_tensor_name_candidates(name) {
+        if let Some((info, data)) = hfq.tensor_data_vec(&candidate) {
+            return Some((info, std::borrow::Cow::Owned(data)));
+        }
+    }
+    None
+}
+
 fn load_norm_weight(
     hfq: &HfqFile,
     gpu: &mut Gpu,
@@ -813,12 +835,20 @@ pub(crate) fn load_weight_tensor(
     k: usize,
     candidates: fn(&str) -> Vec<String>,
 ) -> HipResult<WeightTensor> {
-    // Use pread path to avoid page cache buildup on unified-memory APUs.
+    // Zero-copy first: when the mmap is alive (dGPU loads keep it), DMA
+    // straight from the page-cache-backed slice — no heap staging copy.
+    // Pread fallback preserves UMA behavior (mmap dropped there).
     #[cfg(unix)]
     {
         let mut wt: Option<WeightTensor> = None;
         let mut matched: Option<String> = None;
         for candidate in candidates(name) {
+            if let Some((info, data)) = hfq.tensor_data(&candidate) {
+                let qt = info.quant_type;
+                wt = Some(load_weight_tensor_raw(gpu, qt, data, m, k)?);
+                matched = Some(candidate);
+                break;
+            }
             if let Some((info, buf)) = hfq.tensor_data_pread(&candidate) {
                 let qt = info.quant_type;
                 wt = Some(load_weight_tensor_raw(gpu, qt, &buf, m, k)?);
@@ -2441,8 +2471,18 @@ impl WeightSource for HfqSource<'_> {
     }
 
     fn prepare(&mut self, n_devices: usize) -> HipResult<()> {
+        // Keep the mmap alive on discrete GPUs (the carrier cleared
+        // `evict_page_cache` there): weight uploads DMA straight out of
+        // page-cache pages with no heap staging copy — measured 11–16 GB/s
+        // end-to-end vs ~4 GB/s through a pread staging buffer.
+        // UMA keeps the drop — evict=true is exactly the carrier's UMA
+        // signal — so cached model pages can't starve hipMalloc of RAM.
+        //
+        // NOTE: do NOT madvise-populate the mapping here. A blocking
+        // MAP_POPULATE measured identical totals (it relocates the same
+        // soft-fault work out of the upload loop into one serial stall).
         #[cfg(unix)]
-        if n_devices == 1 {
+        if n_devices == 1 && self.hfq.evicts_page_cache() {
             self.hfq.drop_mmap();
         }
         let _ = n_devices;
@@ -2463,7 +2503,7 @@ impl WeightSource for HfqSource<'_> {
                 c.mrope_interleaved, c.mrope_section
             );
         }
-        let (embd_meta, embd_data) = qwen35_tensor_data_vec(self.hfq, "embed_tokens.weight")
+        let (embd_meta, embd_data) = qwen35_tensor_data_cow(self.hfq, "embed_tokens.weight")
             .expect("embed_tokens not found");
         let out = load_embedding(gpu, embd_meta.quant_type, &embd_data, c.vocab_size, c.dim)?;
         drop(embd_data);
@@ -2497,11 +2537,11 @@ impl WeightSource for HfqSource<'_> {
             c.dim,
             |gpu| {
                 let (lm_info, lm_data) =
-                    qwen35_tensor_data_vec(hfq, "lm_head.weight").expect("lm_head present");
+                    qwen35_tensor_data_cow(hfq, "lm_head.weight").expect("lm_head present");
                 load_weight_tensor_raw(gpu, lm_info.quant_type, &lm_data, c.vocab_size, c.dim)
             },
             |gpu| {
-                let (embd_meta, embd_data) = qwen35_tensor_data_vec(hfq, "embed_tokens.weight")
+                let (embd_meta, embd_data) = qwen35_tensor_data_cow(hfq, "embed_tokens.weight")
                     .expect("embed_tokens not found");
                 dequant_weight_raw(gpu, embd_meta.quant_type, &embd_data, c.vocab_size, c.dim)
             },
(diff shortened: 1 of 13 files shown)


ghazni101:
Round 2 pushed to this branch (4 new commits, `1cd7a890`..`d2072ddf`) — description updated accordingly.

**New:**
- `perf(qwen35)`: zero-copy packed-MQ4 expert uploads (`HfqFile::data_range` + contiguity probe + mincore-gated borrowed-slice H2D). Weight sweep on ornith-1.5-35b-a3b.mq4r: ~28 s → ~2.7 s; daemon start→ready ~31 s → ~5.5 s warm.
- `fix(runtime)`: the parallel HFQ reader called fadvise(DONTNEED) per segment unconditionally — it evicted its own pages and forced a full disk re-read every restart on dGPUs. Now honors `evicts_page_cache`.
- `fix(generate)`: sequential serve seeded the sampler from constant `0x13579BDF`; same prompt was byte-identical across requests at temp>0. Now keyed per attempt like the batch scheduler; greedy unchanged (verified byte-identical).

All gates pass (fmt, MQ4 speed floor); both PM4 serving containers rebuilt and verified end-to-end.

Kaden-Schutt (changes requested):
Rebase is clean and seed scope is removed, but blockers remain: add a non-Unix mostly_page_cached_memo fallback; replace the incomplete gfx1103/gfx115x UMA classification with a trustworthy runtime property or complete supported-APU table; make warmer thread creation fallible/degrade gracefully; split the unrelated shared_expert quant policy (and its incomplete VALID_Q8_CLASSES support); and repair the H2D benchmark byte accounting/cold-PTE setup before citing it. Required follow-up validation is Windows compile plus gfx1151 UMA and gfx1100 cold/warm/load-swap parity.

ghazni101 (commented):
Thanks for the review — all five blockers are addressed in `edcc7d64`..`d472dab2`, with the requested validation.

**1. Non-Unix `mostly_page_cached_memo` fallback** — added (`#[cfg(not(unix))]`, returns `false`, callers take the pread path). `cargo check -p hipfire-runtime -p hipfire-loader --target x86_64-pc-windows-msvc` passes clean.

**2. UMA classification → runtime property.** The gfx1103/115x table is deleted. `Gpu::is_uma()` (rdna-compute) queries `hipDeviceGetAttribute(hipDeviceAttributeIntegrated)` live; ordinal pinned at 16 by the same header enumeration that reproduces profiler.rs's `MULTIPROCESSOR_COUNT = 63`. Since ordinals have shifted between ROCm releases before (`ReservedSharedMemPerBlock` insertion moved `MaxSharedMemoryPerBlock` 74→75), the re-verify-on-bump caveat is documented in-source next to the pin. Query failure degrades to UMA=true — the conservative direction: a misclassified dGPU only loses load speed, while a misclassified APU can OOM mid-load. Verified here: gfx1100 reports attribute value 0 → dGPU path taken, matching the box.

**3. Fallible warmer spawn.** `Builder::spawn` per worker; on error we keep the workers that started and degrade gracefully (fewer warmers = less disk bandwidth; zero = unwarmed upload), never the old mid-load panic under RLIMIT_NPROC / cgroup pids.max.

**4. shared_expert quant policy split out.** The `q8_class_of("mlp.shared_expert.")` arm is reverted from this PR (it also shipped without a VALID_Q8_CLASSES registration — any HIPFIRE_Q8_CLASSES/`--fixed-tier` parse naming it would have errored). It rides with its own feature branch.

**5. H2D benchmark repaired** (`d472dab2`) — you were right on both counts. Byte accounting: loops transferred `size/chunk` chunks but reported `min(size/chunk,128)` — fixed by counting actual iterations inside each timing window. Cold-PTE setup: the "cold" pass ran *after* `MADV_POPULATE_READ` had populated every page and PTE, so it measured a fully-warm mapping. Now: `posix_fadvise(DONTNEED)` on the inode *before* mapping (only honored unmapped), then a genuinely cold sweep; POPULATE becomes its own fourth phase. Repaired numbers (gfx1100, qwen3.5-4b.mq4): HEAP_SRC 28.1 GB/s · MMAP_COLD **2.45 GB/s** · MMAP_WARM_PTE 28.1 GB/s · MMAP_POPULATE 27.9 GB/s. The old "cold-PTEs 4.0 GB/s" line is struck from the PR description — real cold-start H2D is disk-bound at ~2.5 GB/s on this array, which is exactly the regime the parallel warmer exists to hide.

**Validation (gfx1100, serialized through gpu-coord):**

| check | result |
|---|---|
| Windows cross-check | `cargo check --target x86_64-pc-windows-msvc` (runtime + loader) ✅ |
| Smoke: qwen3.5-4b cold/warm | 1146 ms after fadvise drop / 844 ms reload; output coherent |
| Final: qwen3.8-27b cold/warm | 4651 ms cold / 2643 ms warm; greedy outputs byte-identical across both loads |
| Load-swap memo audit (4B→27B) | 27B first-in-proc 2596 ms vs fresh-process 2703 ms — no stale-memo regression |
| Greedy parity vs base `18d46f76` | byte-identical, sha256[:16] `84d20e12286c9aec` both binaries (prompt md5 `1d32df5f12c414d3e34c7b35b6611e6c`) |
| Tests/gates | `cargo test -p hipfire-runtime` 573 pass; MQ4 speed gate prefill +80.6% / decode +9.0% over floor |

One honest gap: **gfx1151 UMA was not re-validated on-device** — no Strix Halo in this rig. With blocker 2 fixed the classification no longer predicts from a name table (the device reports it), so the specific misclassification risk is structurally gone, but I haven't exercised the UMA load path on real hardware this round. Flagged as open follow-up in the PR description.

Unrelated note: `cargo test -p hipfire-quantize` has one failing test on this branch independent of this PR — `quant_fwht::tests::mq6v2_mq5v2_wire_layout` fails identically with this PR's quantize change reverted (the crate diff vs master is otherwise empty). Worth a look in whatever branch owns the MQ6v2 wire layout.

Kaden-Schutt:
Local PR-queue disposition: I integrated HEAD `d472dab2`, retaining UMA-gated eviction, parallel warming, zero-copy uploads, MoE expert sweeping, and the swap-memo fix. I realigned the pipeline-parallel UMA decision from `devices[0]` to `devices.iter().any(is_uma)`, so a mixed topology cannot miss UMA because rank 0 is discrete. The PR is now closed; this is an FYI for any future cherry-pick. Nothing was pushed and `master` was not modified.

Kaden-Schutt:
Post-publish confirmation: this PR's exact head `d472dab2cfbc9c82b2036316e3d8c1dad3db9829` is now contained in `origin/master` through preserved merge commit `8732fd841`. Commit authorship is intact and the PR is included in the regenerated `CREDITS.md` inventory.

GitHub kept the API state as `CLOSED` because the PR had already been manually closed before master was published. Both reopen APIs refuse with `These commits are already merged`, so I applied the visible `merged-via-master` label rather than adding a fake empty merge commit to master.

ghazni101:
Ohh I closed all the PRs thinking I will redo them with the newer mq4v2 weight validations. No worries. I will be opening new PRs after testing their validation on the newer mq4v2 weights

PR #626, opened 2026-08-22
Multi-slot VL support; quantizer F16-fallback fix; slot engine VRAM + context-cap fixes

ghazni101:
> **Superseded** by #636 (multi-slot VL delta) + #637 (residual engine fixes) — rebased on master and split per review triage ([comment](https://github.com/warpfront/hipfire/pull/626#issuecomment-5389792387)).

---

## Multi-slot VL support + quantizer/slot-engine fixes

One PR, five independent fixes. All validated end-to-end on gfx1101 (RX 7800 XT) serving `ornith-1.5-9b` (multimodal qwen3_5 checkpoint) through the multi-slot engine.

### 1. Multi-slot engine: vision support

The slot engine was text-only; image requests either failed or silently lost the image.

- ViT tower loads alongside trunk weights in `SlotEngine` (before the trunk loader drops the mmap — order matters)
- Image requests are preprocessed engine-side (CPU smart-resize + patch extraction), run through `vision_forward`, and their embeddings are spliced into the batched prefill via `SlotStepOpts.embed_override`
- FA layers dispatch to `rope_mrope_halfsplit_f32_batched` with per-row (t,h,w) positions on visual steps; text-only steps keep the 1D kernel and hipGraph decode path
- Scheduler tracks per-slot VL state through chunked prefill (`SlotBatch.vl_rows`)
- CLI slot path extracts OpenAI-style `image_url` parts and renders the exact pad span (CPU preprocess computes `n_vis`; GPU re-preprocesses for embeddings)

### 2. Quantizer: silent tensor drop + MoE panic on multimodal checkpoints

Quantizing a multimodal qwen3_5 safetensors checkpoint produced an HFQ that failed at load with `tensor not found: norm.weight`, and earlier builds panicked with an index-out-of-bounds.

- `handle_moe_expert_3d`: guard rank/name before indexing `inner_shape[1]` (dense `lm_head.weight` reached it and panicked); non-matching tensors fall through
- `handle_main_quant`: restore the missing F16 fallback branch — norms, biases, and sub-32-element weights were silently dropped from every safetensors quant (stock artifacts carried 79 norm tensors; affected quants carried zero)

### 3. Slot engine: VRAM + context-cap

- Drop the spec-decode DeltaNet S-tape from `PrefillBatchScratch` in `SlotEngine::build` — never consumed by the slot path, ~4 GiB wasted at 2048-token context
- Context-cap guard in the decode loop: stop cleanly instead of overflowing `set_seq_len` (long sessions previously killed the entire batched step)

### 4. Slot engine: decouple prefill chunk width from KV capacity

`chunk_size` was set to `cap_tokens`, so a 50k-token context attempted a single
50k-row prefill step, sizing `PrefillBatchScratch` at ~22 GiB. Prefill now runs
at a fixed 2048-row chunk through the existing scheduler chunking; scratch stays
constant (~0.9 GiB) and only the KV arenas scale with context.

Validated on gfx1101: **2 slots x 51200 ctx = 102k total context** at 9.1 GiB
resident; a 34k-token prompt prefills in chunks and answers coherently; vision
and 2-way concurrency unaffected.

### 5. Slot engine context scaling

- Fixed a boot-fatal OOM: prefill scratch was sized by total KV capacity
  (~22 GiB of f32 at 50k ctx); prefill now runs at a fixed 2048-row chunk
  width through the existing scheduler, so scratch is constant and only
  KV arenas scale with context.
- Shipped configuration: equal slabs, 2 x 32768 ctx = 64k total context,
  validated at 8.5 GiB resident on gfx1101 with concurrent text + vision.
- Elastic (primary + reserves) and dynamic-range pool sharing were
  implemented and are retained behind `serve.multi_slot_pool_total`, but
  are gated as experimental: long-prompt prefill against variable-cap
  slabs produces degenerate output in the flash attention path. Equal
  slabs are the shipped default; upstream follow-up needed on
  variable-cap descriptor handling.

### Also

- `gen_start` emitted before the daemon VL route so the CLI stream-contract gate accepts serve-layer VL requests (previously hung, then aborted)
- `serve.multi_slot_slots` / `serve.multi_slot_ctx` added to the config schema (previously unsettable)
- `docker/quantizer.Dockerfile`: standalone CPU quantizer image

## Validation

gfx1101, ROCm 7.2, ornith-1.5-9b mq4 (760-tensor multimodal checkpoint → 6.2 GB):

- Quantize → serve → generate verified end-to-end through the multi-slot engine
- Vision grounding: color/count/position queries correct; black/white discriminators correct
- Concurrency: parallel text + vision lanes both correct; no cross-talk
- `cargo test -p hipfire-arch-qwen35`: 119 passed
- VRAM: ~11.9 GiB → ~6.5 GiB resident (tape fix); ~9.1 GiB at 2x50k context

Signed-off-by: ghazni101 <ghazni101@users.noreply.github.com>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -922,9 +922,11 @@ name = "hipfire-cli"
 version = "0.3.0"
 dependencies = [
  "anyhow",
+ "base64",
  "clap",
  "ctrlc",
  "hipfire-arch-qwen35",
+ "hipfire-arch-qwen35-vl",
  "hipfire-client",
  "hipfire-config",
  "hipfire-registry",
--- a/crates/hipfire-arch-qwen35/src/forward_slots.rs
+++ b/crates/hipfire-arch-qwen35/src/forward_slots.rs
@@ -1437,6 +1437,7 @@ fn run_fullattn_layer_slots(
     max_ctx_len: usize,
     single_slot: Option<(u64, usize)>,
     n_tiles: Option<usize>,
+    mrope_pos3: Option<&GpuTensor>,
 ) -> HipResult<()> {
     let attn_dtype = require_batchable_fullattn_layer(layer)?;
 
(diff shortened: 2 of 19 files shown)


Kaden-Schutt:
Triage: HOLD_WIP. The VL path is unique, but the VRAM/chunk/config pieces overlap merged #607/#608 and master already contains the quantizer F16 fallback. Please rebase and split a focused VL delta plus any residual engine fixes before review.

ghazni101:
Rebased onto master (`b8cd18f`) and split per the triage above:

- #636 — focused multi-slot VL delta (the unique part)
- #637 — residual engine fixes that survived the rebase (context-cap decode guard, MoE expert-3d guard, quantizer image)

Dropped as already-merged: config schema keys (#607), S-tape skip + chunk-width/prefill-scratch sizing (#608), quantizer F16 fallback (master). Elastic/dynamic-range pool sharing is deferred until variable-cap descriptor handling in the flash attention path is sorted.

This PR is superseded by the two above.

PR #629, opened 2026-08-23
maple: serve Maple-Preview 20B-A1B natively-ternary MoE (arch 15, qt=51 MQ2G256LloydU)

nwoolmer:
Serves **deepgrove/maple-preview** — a 20B-A1B natively-ternary MoE — as
`hipfire-arch-maple`, **arch_id 15**. Load, decode, batched prefill, HTTP
serving with tool calling, and a published registry entry.

**Targets `master` directly** — no longer stacked. #599 supplied the MQ2-Lloyd
container, its MoE kernel suite, and the `hipfire-quantize` layout; it merged on
2026-08-24, so that dependency is upstream and this branch was retargeted from
`quant/quality`. Review `master...quant/maple-preview` — 47 commits, 53 files.

qt=51 `MQ2G256LloydU` is claimed in the authoritative register and mapped
without disturbing qt44–50.

## The key property

Every linear weight in the published checkpoint is already `{-s_r, 0, +s_r}`
with one bf16 scale per output row. So this is a **packing problem, not a
quantization problem**, and the acceptance bar is exactness rather than KLD.

`MQ2G256LloydU` (**qt=51**, 2.250 bpw) is the unrotated sibling of MQ2-Lloyd:
byte-identical 72 B/group layout, so every existing MQ2-Lloyd kernel binds — it
simply must not FWHT-rotate the activation. **No new HIP kernels**, including
for the arch itself: the batched dense GEMM is the grouped MoE WMMA kernel
driven as a single-expert case (`E=1`, zero tile ids, identity slot index,
`x_row_div=1`).

Verified value-exact on the real checkpoint: `max|err| = 0` across all 9 shards,
with every bitwise difference accounted for as a signed zero (~19% of source
weights are `-0.0` and return `+0.0`). Cross-checked byte-for-byte against an
independently-written Python packer, and the full convert refuses non-ternary
rows rather than falling back — so a clean run proves the per-row claim on 100%
of the model, not a sample.

## Numbers (gfx1151)

| | |
|---|---|
| Model | 6.79 GB, 2.250 bpw |
| Prefill | 24.4 s → **2.00 s** on a 3,059-token prompt (**12.6x**) |
| Decode | ~120 tok/s, flat with context (18 of 24 layers are SWA-512) |

Prefill was measured as a controlled A/B from one binary via
`HIPFIRE_MAPLE_PER_TOKEN_PREFILL`, because the machine was under load.

## Quality

Three independent implementations agree, so the port is faithful:

| | PPL |
|---|---|
| this branch | 53.74 |
| llama.cpp bf16 reference, same tokens/protocol | 54.41 |
| vendor's own ternary GGUF vs our bf16, standard protocol | 57.50 vs 57.54 |

KL(reference ‖ ours) mean **0.084**; paired PPL ratio 0.9876, CI [0.962, 1.014]
— spans 1.0. The Q8 KV cache is specifically exonerated: KL does not accumulate
with context, and position 0 (no KV history) sits at 0.00014.

Maple's absolute wikitext PPL is ~57 where a peer scores ~7.7. That is the
**model**, not the packing — the bf16 reference, which cannot have an arch-15
defect, scores the same. Corpus and protocol were ruled out by holding the model
fixed and varying one at a time. Note for maintainers: **use mean KL 0.084 as
the numerics regression gate, not PPL** — PPL here is dominated by the model and
would hide a real kernel regression inside its own noise.

## Bugs this found in existing code

- `run_moe_decode` FWHT-rotated the gate→down intermediate for an unrotated
  routed dtype. Fixed, with a test requiring rotated and unrotated dtypes to
  disagree on both rotation flags so it cannot pass vacuously.
- `gemm_mq2g256_lloyd_moe_grouped_wmma` always called the pointer-keyed
  `ensure_fp16_x`, so a caller reusing one scratch buffer across layers silently
  got layer 0's activations. Now accepts an already-F16 activation, mirroring
  `gemm_q8_0_wmma`. Every existing caller passes F32 and is unaffected.
- **Pre-existing, NOT fixed here:** `generate_qwen2` never emits `gen_start`, so
  arch 7 over `hipfire serve` should hang on the first request exactly as arch
  15 did before this branch fixed its own copy.

## Serving

`hipfire serve` works, including tool calling: the vendor ships a Qwen-style
chat template in their GGUF metadata (HF's `tokenizer_config.json` has none), so
the converter now embeds it and the model emits well-formed `<tool_call>` JSON
using the supplied tool names. `maple_ar` is admitted to the tool-capable
producers, with `semantic_contract_version` deliberately left `None` (v2 would
misfile the `<think>` span) and a fail-closed guard for template-less files.

## Known limitations

- Greedy only — `temp`/`top_p`/`repeat_penalty`/`stop` are accepted and ignored
  (matches the `generate_qwen2` precedent).
- No prefix cache; each turn re-prefills. Cheap at 1500 tok/s, but an LCP cache
  is the obvious follow-up.
- ~132 MiB of batch scratch is allocated eagerly, even for decode-only
  processes. Sizing from `MAPLE_PREFILL_CHUNK` rather than the cap would cut the
  grouped block 93 → 69 MiB; the `forward_batch` guard already checks
  `state.max_b` so that change is safe to make.
- Arm A (TQ2G128 + a ternary MoE kernel suite) is not implemented, so the
  arm-A/arm-B logit differential described in the design doc was not run.

Design docs: `docs/design/2026-08-22-maple-preview-20b-a1b.md` and
`docs/design/2026-08-22-maple-batched-prefill.md`.


Kaden-Schutt:
Rebased onto current master and retargeted from quant/quality. qt=51 `MQ2G256LloydU` is now claimed in the authoritative register and mapped consistently without disturbing qt44–50. This is ready for a focused architecture review; please refresh the PR body’s stale stacked-on-#599 wording.

nwoolmer:
Rebase follow-up pushed at `152d8b338`, plus the PR body refresh you asked for.

### PR body

Refreshed. #599 merged into `master` at 00:38 on 2026-08-24, so the stacked-on wording was stale in both directions — the dependency is upstream now and the branch targets `master`. Review range is `master...quant/maple-preview`, 47 commits / 53 files.

### CI was red after the rebase; here is the split

**`build` — fixed, and it was never this branch's.** `crates/rdna-compute/examples/test_mq4v2_gate_up_bt_gfx1100.rs` calls `gemm_gate_up_mq4g256v2_wmma_gfx1100_mw_lds`, which did not exist at the merge base `18d46f765`. The example file is byte-identical to `master`'s. `master` fixed it in `bca38df0a fix(gfx1100): restore accepted mq4v2 multi-wave dispatch`, one of the 9 commits it had moved ahead. Merging current `master` resolves it — `cargo check --workspace --all-targets` is clean.

The merge conflicted only in `registry/v1.json`, and only on the `generated_at` line. Worth recording that the auto-merge dropped `qwen3.8:27b-fast`: that is correct, not a loss — `02cf491c3 registry(qwen38): publish mq v2 product ladder` removed it on `master`. Verified the merged file is still a strict superset of `registry/models.json` and retains all 77 of `master`'s entries plus `maple-preview`.

**`gates` — my share is fixed; the rest is `master`'s.** The rebase left `hipfire-arch-maple` unregistered with three gates. Measured against `master` HEAD `336dd0777`:

| metric | master | before | now | ceiling |
|---|---:|---:|---:|---:|
| `ungated_examples` | 46 | 52 | **46** | 32 |
| `layer_unlisted_crates` | 0 | 1 | **0** | 0 |
| `bypass_unlisted` | 0 | 1 | **0** | 0 |
| `bypass_regressions` | 1 | 1 | 1 | 0 |
| `bypass_total` | 232 | 236 | 236 | 225 |

- **layering** — layer 6. Maple's deepest dependency is `hipfire-runtime` at 5 and it has no arch-crate dependency, so it sits with gemma4/cohere2moe, not with qwen35 at 7.
- **bypass ledger** — 4 calls (1 `batch.rs`, 3 `forward.rs`), 0 registry refs, so `debt`.
- **ungated examples** — the six probes move behind `--features lab`, matching cohere2moe. Confirmed they still compile (`cargo check -p hipfire-arch-maple --examples --features lab`) so the gating hides nothing.

I raised `bypass_total` 221 → 225 with a declared `RATCHET-RAISE:` trailer, banking **only** maple's own 4. I deliberately did not absorb the rest.

**Two failures left, both pre-existing on `master`, both untouched here:**

1. `bypass_regressions = 1` — `hipfire-arch-qwen35` measures 128 against a recorded 117. It grew +11 in the MQ V2 release without the ledger being updated.
2. `bypass_total = 236` vs ceiling 225 — the gap is exactly those 11.

That is master's regression to bank or pay down, not mine to hide, so I left it visible. Say the word and I'll add the one-line ledger edit in a separate commit — that would take `gates` green for everyone, but it is your call whether +11 gets banked or migrated.

`ungated_examples` (46 > 32) and `clippy` are also red on `master` HEAD, so this branch cannot go green until those move.

### `check-crate-maps`

Was failing on 13 crates on `master` and 16 here. Added the missing `crates/hipfire-arch-maple/map.md` — it was the only arch crate without one — and regenerated the nine maps that named maple. Now 5, all of them `master`'s (`arch-deepseek4`, `arch-qwen35`, `cli`, `daemon`, `registry`), none maple-attributable.

### Verified

`cargo check --workspace --all-targets` clean · 201 unit tests pass across `hipfire-arch-maple` (29), `hipfire-quantize` (45), `saddle-quant` (127) · `ratchet-diff` OK, no expectation weakened.

No GPU runs — the numbers in the PR body are from the earlier hardware sessions, not re-measured here.

Kaden-Schutt:
Local PR-queue disposition: I integrated HEAD `152d8b33`, retaining the Maple architecture/crate, MQ2G256LloydU path, registry entry, pack/prefill/serve flow, and lab examples. I realigned Maple's capability to `ReasoningContract::QwenJinja` rather than `Unsupported`, threaded thinking enable/cap/prefix controls into Maple generation, and kept the unrelated Glimmer/Gemma reasoning paths byte-aligned with master. WMMA remains fail-closed where unsupported.

The stack builds/tests locally, but we did not have a Maple fixture for a fresh Maple KLD/serve run. Please account for the QwenJinja reasoning contract when rebasing. This is local preflight only: nothing has been pushed and nothing has been merged to `master`.

PR #630, opened 2026-08-23
fix(scripts): prepend weight-cache preamble in compile-kernels.ps1

### LINKED ISSUE #618: kernels: 52/630 fail to compile for gfx1100 from source (undeclared HIPFIRE_WEIGHT_RSRC / row_rsrc / gate_base, missing _512 helpers)
Fresh from-source install on a 7900 XTX (gfx1100) ends with `scripts/compile-kernels.ps1 gfx1100` reporting **578 OK / 52 FAIL**. Same result on master `80a572c824dc` and on `d5a28ba`, so it's not a single bad commit.

Most of the failures look like leftovers from the HUSRCF/Saddle refactor — the sources reference symbols that are declared nowhere in `kernels/src`:

- `gemv_hfq4g256_moe_gate_up_indexed.hip` → `gate_base`, `expert_rsrc` undeclared
- `gemv_hfq4g256_residual.gfx1100.hip` → `row_rsrc` undeclared (via `RESIDUAL_WEIGHT_LOAD`)
- `gemv_mfp4g32_e8_soa_u4_buffer.hip`, `gemv_hfq4g256_moe_projection_indexed_gfx1151.hip` → `HIPFIRE_WEIGHT_RSRC` undeclared (it's defined ad-hoc in *some* sibling files like `gemv_hfq4g256_residual_scaled.hip`, which defines it as `0`)
- `kv_cache_write_fwht3_hd512{,_batched}.hip` → `fwht_shfl_forward_512`, `turbo_quantize_3bit_512`, `TURBO_C3_512` missing (only the `_256` versions exist in `turbo_common.h`)
- `gemv_mq8g256.hip` → `'__builtin_amdgcn_sdot4' needs target feature dot1-insts` on gfx1100
- plus the `_gfx1151` projection variant being compiled under gfx1100

The rest of the fail list is gfx906/gfx12-only stuff (`*_mmq_gfx906_*`, `*_wmma_gfx12_*`, `*_wave64_dp4a`) which is probably fine to skip, but the HFQ4 ones above are the dense Qwen3.5 hot path, so a source install currently can't serve those models on gfx1100.

Repro:

```
git clone https://github.com/warpfront/hipfire
cd hipfire
.\scripts\compile-kernels.ps1 gfx1100
```

Example error (clang from ROCm 7.2):

```
gemv_hfq4g256_residual.gfx1100.hip:134:47: error: use of undeclared identifier 'row_rsrc'
  float sc0 = __builtin_bit_cast(float, RESIDUAL_WEIGHT_LOAD(gp0, (g + 0) * 136));
gemv_hfq4g256_residual.gfx1100.hip:50:37: note: expanded from macro 'RESIDUAL_WEIGHT_LOAD'
  HIPFIRE_WEIGHT_LOAD_PTR_U32(row_rsrc, ptr, offset)
```

I only verified this with ROCm 7.2's clang on Windows, but since these are plain undeclared identifiers in the sources I'd expect the same everywhere. Full FAIL list available if useful.

SergioSV96:
Fixes #618.

## Root cause

The 52 gfx1100 compile failures reported in #618 are **not** broken kernel sources — `compile-kernels.ps1` compiles raw `.hip` files, while HFQ4 sources reference the `HIPFIRE_WEIGHT_*` macros that `gfx12_weight_cache_policy.inc` provides and that `kernels.rs` concatenates ahead of them for JIT compilation (`crates/rdna-compute/src/kernels.rs`). Without that preamble clang fails with undeclared identifiers (`HIPFIRE_WEIGHT_RSRC`, `row_rsrc`, `gate_base`, `expert_rsrc`) even though the sources are valid.

In non-buffer mode the load macros ignore the rsrc argument entirely (`gfx12_weight_cache_policy.inc:50-53`), so prepending the plain preamble is sufficient — it produces ordinary-dispatch binaries, which is exactly what the offline cache is for. The JIT can still build its optimized buffer variants at runtime into its own hash-keyed cache.

## Changes

- Prepend the preamble when a source references those macros but neither defines nor includes it itself (avoids macro redefinition in self-contained files like `gemv_hfq4g256_residual_scaled.hip`)
- Skip sources pinned to another chip (`*_gfx1151`, `*_gfx906_*`, `*_gfx12_*`), keeping family variants like `.gfx11_dgpu`
- Skip sdot4/dp4a kernels on gfx11 (`dot1-insts` unavailable)
- Skip `attention_flash_fwht3_tile_hd512*` / `kv_cache_write_fwht3_hd512*`: they reference `fwht/turbo *_512` helpers absent from `turbo_common.h` (upstream TODO — note `asym_k_givens3_hd512` builds fine and is kept)
- Skip `gemv_mfp4g32_e8_soa_u4_buffer` (JIT-only opt-in route)

## Result (gfx1100, ROCm 7.2, Windows)

**581/581 compiled, 0 failed** — including `gemv_hfq4g256`, `fused_qkv/gate_up/qkvza_hfq4g256`, and the full MoE family. Remaining known gaps are upstream follow-ups, not script issues: the `_512` helper family and dot1-insts coverage for RDNA3.

### DIFF
--- a/scripts/compile-kernels.ps1
+++ b/scripts/compile-kernels.ps1
@@ -115,6 +115,41 @@ foreach ($arch in $Archs) {
             }
         }
 
+        # RDNA3 (gfx11) lacks dot1-insts: every sdot4/dp4a kernel cannot build.
+        if ($archFamily -eq "gfx11" -and ($name -eq "gemv_mq8g256" -or $name -match '_dp4a$')) {
+            Write-Host "  - $name SKIP (dot1-insts unavailable on $arch)"
+            continue
+        }
+
+        # fwht3 hd512 KV/attention kernels reference fwht/turbo *_512 helpers
+        # that do not exist in turbo_common.h yet (upstream TODO). Other
+        # _hd512 kernels (e.g. asym_k_givens3) build fine and stay.
+        if ($name -match '^(attention_flash_fwht3_tile|kv_cache_write_fwht3)_hd512') {
+            Write-Host "  - $name SKIP (missing _512 helpers in turbo_common.h)"
+            continue
+        }
+
+        # Temporal-buffer GEMVs are opt-in routes assembled by the JIT with
+        # CACHE_ELIGIBLE/OPT_IN defines; without them the source cannot link
+        # against ordinary pointer loads. Not part of the default dispatch.
+        if ($name -eq "gemv_mfp4g32_e8_soa_u4_buffer") {
+            Write-Host "  - $name SKIP (opt-in buffer route; JIT-only)"
+            continue
+        }
+
+        # Skip sources whose name pins them to a different chip
+        # (e.g. *_gfx1151, *_gfx906_*, *_gfx12_*) unless the embedded arch is
+        # this arch or its family (gfx11_dgpu style variants stay).
+        if ($name -match '[._]gfx(\d+)') {
+            $pinned = $Matches[1]
+            $archNum = $arch -replace '^gfx', ''
+            $famNum = $archFamily -replace '^gfx', ''
+            if ($pinned -ne $archNum -and $pinned -ne $famNum) {
+                Write-Host "  - $name SKIP (pinned to gfx$pinned)"
+                continue
+            }
+        }
+
         # Variant precedence
         $chipVariant   = Join-Path $SrcDir "$name.$arch.hip"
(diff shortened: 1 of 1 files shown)


Kaden-Schutt (approved):
The script now mirrors runtime source preamble assembly and skips only explicitly unsupported/JIT-only chip surfaces; the Windows 581/581 compile evidence covers the changed path.

PR #631, opened 2026-08-23
fix(config,rdna-compute): resolve arch device selectors at GPU init

### LINKED ISSUE #623: Windows: daemon always uses iGPU as device 0; visibility env vars ignored, arch selector not applied in --precompile
On a dual-GPU Windows box (integrated `AMD Radeon(TM) Graphics` = gfx1036 + RX 7900 XTX = gfx1100), the daemon always initializes the iGPU as device 0, and I couldn't find a way to steer it to the dGPU:

- `HIP_VISIBLE_DEVICES=1` → still reports `GPU dev 0: gfx1036 (13.0 GB VRAM, HIP 7.2)`
- `ROCR_VISIBLE_DEVICES=1` → same
- both set → same
- `hipfire config set devices arch:gfx1100` (writes `[hardware] devices = "arch:gfx1100"` in `~/.hipfire/config.toml`) → standalone `daemon --precompile` still picks gfx1036

I saw `apply_device_visibility` / the arch-selector machinery in hipfire-config, but it doesn't seem to be effective on this path (either because the Windows HIP runtime ignores visibility env vars entirely, or because `--precompile` initializes the GPU before config is consulted — didn't dig that far).

Consequence: kernels precompile for the wrong target and any serve/run would land on the iGPU.

Question: what's the supported way to pin a specific GPU on Windows? If it exists, it's not obvious from CONFIG.md/GETTING_STARTED.md; if it doesn't, maybe `hardware.devices` should be honored before first HIP init everywhere.

Env: Windows 11, ROCm 7.2 HIP SDK, master `80a572c824dc`, daemon.exe built from source.

SergioSV96:
Fixes #623.

Two coordinated changes — `hardware.devices = arch:gfxXXXX` never worked end to end:

## 1. hipfire-config: selector kinds are not ordinal lists

`synchronized_device_visibility` treated the whole selector string as a physical ordinal list: it counted one comma-free entry, computed `count = 1`, and set `HIP_VISIBLE_DEVICES = "0"` — masking every device except the first adapter (on dual-GPU Windows boxes, the integrated one). Selector kinds (`arch:`/`uuid:`/`pci:`) now skip ordinal-list env masking; they are stable identities resolved where live device information exists. Plain ordinal lists (`"1,2"`) keep the exact legacy masking path on all platforms, and inherited-env pairing logic is untouched.

## 2. rdna-compute: resolve the selector at GPU init

`Gpu::init()` hardcoded device 0, and on Windows the HIP runtime ignores `HIP/ROCR_VISIBLE_DEVICES` entirely, so masking could not have worked anyway. `init()` now resolves an `arch:` selector against per-device `get_arch()` and initialises that device; anything else falls back to device 0 exactly as before (Linux behaviour for ordinals and inherited-env setups is unchanged).

## Verified

Dual-GPU Windows box (iGPU gfx1036 = dev 0, RX 7900 XTX = dev 1), ROCm 7.2:

```
[hipfire] hardware.devices: using dev 1 (gfx1100)
GPU dev 1: gfx1100 (25.8 GB VRAM, HIP 7.2)
Pre-compiling kernels for gfx1100...
precompile: 10 ok, 0 optional failed
```

Previously the same invocation always picked the iGPU.

### DIFF
--- a/crates/hipfire-config/src/lib.rs
+++ b/crates/hipfire-config/src/lib.rs
@@ -700,6 +700,48 @@ pub static FIELDS: &[ConfigField] = &[
         "Serve requests concurrently on the multi-slot engine instead of one at a time."
     ),
     field!(
+        "serve.multi_slot_slots",
+        "multi_slot_slots",
+        Serve,
+        Process,
+        DefaultValue::Integer(4),
+        ValueRule::Integer { min: 1, max: 64 },
+        false,
+        false,
+        Some("HIPFIRE_SERVE_MULTI_SLOT_SLOTS"),
+        "Concurrent slots for the multi-slot engine."
+    ),
+    field!(
+        "serve.multi_slot_ctx",
+        "multi_slot_ctx",
+        Serve,
+        Process,
+        DefaultValue::Integer(8192),
+        ValueRule::Integer {
+            min: 512,
+            max: 1048576
+        },
+        false,
+        false,
+        Some("HIPFIRE_SERVE_MULTI_SLOT_CTX"),
+        "Per-slot context capacity (tokens) for the multi-slot engine."
+    ),
+    field!(
+        "serve.multi_slot_prefill_chunk",
+        "multi_slot_prefill_chunk",
+        Serve,
+        Process,
+        DefaultValue::Integer(1024),
+        ValueRule::Integer {
+            min: 1,
+            max: 1048576
+        },
+        false,
+        false,
+        Some("HIPFIRE_SERVE_MULTI_SLOT_PREFILL_CHUNK"),
+        "Prefill tokens taken from one slot per multi-slot step; batch scratch is sized n_slots x this."
+    ),
+    field!(
         "generation.temperature",
         "temperature",
         Generation,
@@ -3042,19 +3084,6 @@ pub fn synchronized_device_visibility(
 ) -> Result<Option<DeviceVisibility>> {
     let configured = config.legacy_value("HIPFIRE_DEVICES");
     if let Some(configured) = configured.as_deref() {
-        // Only an all-`arch:` selector list skips ordinal-list masking: those
-        // are stable identities resolved against live devices at GPU init
-        // (rdna-compute Gpu::init). uuid:/pci: selectors keep the historical
-        // verbatim passthrough until per-device probes exist, and plain
-        // ordinals take visibility_from_physical exactly as before.
-        let all_arch = !configured.trim().is_empty()
-            && configured.split(',').all(|part| {
-                let part = part.trim();
-                part.len() > "arch:".len() && part.starts_with("arch:")
-            });
-        if all_arch {
-            return Ok(None);
-        }
         return visibility_from_physical(configured).map(Some);
     }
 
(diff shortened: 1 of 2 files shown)


Kaden-Schutt (changes requested):
The arch: path is relevant, but synchronized_device_visibility now bypasses ordinal masking for every selector containing `:`, while Gpu::init resolves only `arch:`. Existing `uuid:` and `pci:` selectors therefore fall through to device 0. Please resolve all stable selector kinds at init (or only bypass visibility for selectors actually resolved here) and add pure coverage for arch/uuid/pci plus no-match behavior.

SergioSV96:
﻿Pushed revisions addressing both points:

- EAP is back to "Stop" globally; all git invocations now route through an Invoke-Git wrapper that relaxes the preference only for the call site (stderr progress no longer aborts fetch/checkout/describe) and restores it immediately. Explicit $LASTEXITCODE checks keep every other failure fatal.
- AdapterRAM is null-safe now (`if ($null -eq $_.AdapterRAM) { [uint64]0 } else { [uint64]$_.AdapterRAM }`).

Happy to narrow further if you'd prefer per-site try/catch instead of the wrapper.

Kaden-Schutt:
Local PR-queue disposition: this PR remains intentionally held out; it was not merged. The issue is architectural rather than a textual conflict: the branch creates a second, weaker `arch:` selector grammar beside `DeviceSelector`/`resolve_device_selector`, is case-sensitive and first-match/fail-open to device 0, and hooks only `Gpu::init`, so TP/multi-device `init_with_device` paths still bypass it.

Requested redesign before reconsideration: extend one canonical selector grammar at the config/visibility layer; fail closed on malformed/mixed/no-match/ambiguous selectors; resolve exactly once to ordinals; feed those ordinals into both single-GPU and multi-device/TP construction; and either fully support uuid:/pci: or do not bypass masking for them. Please add pure coverage for arch/uuid/pci, no-match, ambiguity, mixed lists, and a two-arch TP case. Also, the current PR comment about `Invoke-Git` appears to belong on installer PR #632—please untangle the cross-post.

Nothing was pushed and `master` was not modified.

Kaden-Schutt:
Reassessed against current `master` (`616336932`) after the landing train. An isolated rebase of this branch now collapses to a formatting-only delta in `rdna-compute/src/dispatch.rs`; there is no remaining semantic fix to merge.

The underlying #623 behavior is not solved by this branch's final shape. `synchronized_device_visibility` still sends configured `hardware.devices` through `visibility_from_physical`, while multi-device construction still expects integer ordinals. The canonical `DeviceSelector` type now on master is used by DeepSeek heterogeneous placement, not as a once-only resolver for process-wide `hardware.devices`.

So the architectural review request remains: resolve the one canonical selector grammar against a pure device inventory before visibility is installed; fail closed on malformed, mixed, no-match, and ambiguous selections; materialize ordinals once for both single-GPU and TP/multi-device construction; and cover arch/uuid/pci plus two-arch TP. This PR no longer carries that implementation, so I am closing it rather than merging a no-op. Issue #623 should remain open for a replacement change.

PR #632, opened 2026-08-23
fix(installer): run under Windows PowerShell 5.1 and prefer the discrete GPU

### LINKED ISSUE #619: install.ps1 cannot run under stock Windows PowerShell 5.1 (UTF-8 without BOM + EAP=Stop turns git stderr fatal)
`scripts/install.ps1` can't run at all under the PowerShell that ships with Windows (5.1). Two independent problems:

**1. UTF-8 without BOM.** The file has box-drawing characters in comments (`───`), and PS 5.1 reads BOM-less files as ANSI. Some multi-byte sequences decode into curly-quote characters, which breaks string parsing and produces a wall of parser errors before anything runs:

```
Token '}' inesperado en la expresión o la instrucción.
En install.ps1:178 Carácter: 5
```

Saving the same file as UTF-8 **with BOM** fixes it — zero code changes needed.

**2. `$ErrorActionPreference = "Stop"` + native stderr.** After fixing the encoding, checkout still dies randomly: with EAP=Stop, any native command writing to stderr (git prints its normal progress on stderr, e.g. `From https://github.com/warpfront/hipfire...`) becomes a terminating error in PS 5.1, even when redirected. This kills `Checkout-InstallRef` mid-fetch ("Checkout failed" even though fetch succeeds) and later trips on `git describe --tags`. Setting `$ErrorActionPreference = "Continue"` lets the whole installer run to completion; real failures are still caught because the script checks `$LASTEXITCODE` explicitly.

Repro: on a default Windows 11 box with only built-in PowerShell:

```
irm https://raw.githubusercontent.com/warpfront/hipfire/master/scripts/install.ps1 -OutFile install.ps1
powershell -ExecutionPolicy Bypass -File .\install.ps1
```

Suggestion: save/ship the file with a BOM (or note that pwsh is required), and either drop EAP=Stop or scope it away from the git calls. Happy to send a PR if that helps.

### LINKED ISSUE #620: install.ps1 GPU detection grabs the first AMD adapter - iGPU wins over the dGPU
On a system with both an iGPU and a discrete card, the installer detects the wrong GPU and then dies in an interactive prompt it can't run.

The detection does:

```powershell
$AmdGpu = $VideoControllers | Where-Object { $_.Name -match "AMD|Radeon" } | Select-Object -First 1
```

`Win32_VideoController` on my machine returns:

1. `AMD Radeon(TM) Graphics` (integrated) — matches no arch regex → `$GpuArch = "unknown"`
2. `AMD Radeon RX 7900 XTX` — would have matched `gfx1100`, but is never considered

Since `-First 1` grabs the iGPU, arch stays unknown, and then:

```
WARNING: Could not detect GPU architecture.
Read-Host : Windows PowerShell se encuentra en modo no interactivo...
```

i.e. it falls into `Read-Host "Enter your GPU arch"`, which throws when install.ps1 is run non-interactively (the documented `iex (irm ...)` flow).

Suggestion: prefer a discrete adapter before falling back to any AMD adapter — e.g. match `RX\s*\d` across *all* controllers first, or sort by AdapterRAM descending. Happy to PR that.

SergioSV96:
﻿Fixes #619, fixes #620.

Three issues made `install.ps1` unusable on a stock Windows box. All observed on Windows 11 with PowerShell 5.1 (the built-in shell), RX 7900 XTX + integrated Radeon.

## 1. UTF-8 without BOM (#619a)
Re-saved with BOM; no code change needed beyond that.

## 2. `ErrorActionPreference = "Stop"` + native stderr (#619b)
Scoped handling via `Invoke-Git` wrapper that relaxes the preference only for git invocations and restores it afterwards. Real failures remain fail-closed via explicit `LASTEXITCODE` checks.

## 3. GPU detection prefers the first AMD adapter (#620)
Now prefers the adapter with the largest reported memory, null-safe and deterministic across RX / AI PRO / embedded naming.

Tested on Windows 11 (PS 5.1, RX 7900 XTX + iGPU): installer completes, detects gfx1100, builds daemon+CLI from source.

### DIFF
--- a/scripts/install.ps1
+++ b/scripts/install.ps1
@@ -1,4 +1,4 @@
-# hipfire installer for Windows — detects GPU, installs deps, downloads binary + kernels.
+﻿# hipfire installer for Windows — detects GPU, installs deps, downloads binary + kernels.
 # Usage: irm https://raw.githubusercontent.com/warpfront/hipfire/master/scripts/install.ps1 | iex
 param(
     [string]$Ref,
@@ -8,6 +8,21 @@ param(
 )
 
 $ErrorActionPreference = "Stop"
+
+# Windows PowerShell 5.1 turns native stderr into a terminating error under
+# EAP=Stop, and git prints its normal progress (fetch etc.) on stderr. Scope
+# the relaxation to git invocations via this wrapper instead of weakening the
+# whole script; real failures are still caught by explicit $LASTEXITCODE
+# checks and try/catch at every call site.
+function Invoke-Git {
+    $prev = $ErrorActionPreference
+    $ErrorActionPreference = "Continue"
+    try {
+        & git @args
+    } finally {
+        $ErrorActionPreference = $prev
+    }
+}
 
 # ─── Paths ───────────────────────────────────────────────
 $HipfireDir  = "$env:USERPROFILE\.hipfire"
@@ -55,7 +70,7 @@ if ($InstallRefKind -eq "commit" -and $InstallRef -notmatch '^[0-9a-fA-F]{7,40}$
 }
 
 function Test-RemoteRef([string]$Repo, [string]$RemoteRef) {
-    & git -C $Repo ls-remote --exit-code origin $RemoteRef 2>&1 | Out-Null
+    Invoke-Git -C $Repo ls-remote --exit-code origin $RemoteRef 2>&1 | Out-Null
     return $LASTEXITCODE -eq 0
 }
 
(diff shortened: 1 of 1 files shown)


Kaden-Schutt (changes requested):
The BOM and discrete-GPU ordering are useful, but setting `$ErrorActionPreference = "Continue"` globally weakens every PowerShell failure in the installer, not only git progress on PS5.1. Please scope native-command stderr handling to the affected invocation(s) while retaining fail-closed behavior elsewhere; also handle null/non-numeric AdapterRAM deterministically.

SergioSV96:
﻿Addressed in the pushed revision:

- The visibility bypass is narrowed to all-`arch:` lists only; `uuid:`/`pci:` selectors keep the historical verbatim passthrough (pre-patch behaviour) instead of silently falling through.
- `Gpu::init()` warns loudly when it sees uuid:/pci: kinds it cannot resolve.
- Selector parsing/matching extracted into pure helpers (`device_selector_kinds`, `requested_arch`, `find_matching_arch`) with unit coverage including the no-match case, ordinal pass-through, mixed lists, and pci BDF colons.

CI red on this PR mirrors master-wide failures at tip; happy to rerun anything once base settles.

Kaden-Schutt:
Local PR-queue disposition: I integrated HEAD `9eb54ed1` unchanged. The retained implementation is the PS 5.1 BOM-safe installer, scoped `Invoke-Git` error handling, null-safe AdapterRAM handling, and largest-memory/discrete GPU selection. No maintainer code fix was needed.

The local queue had no Windows host, so Windows acceptance is still unpaid here. Before this is ready upstream, please confirm on stock Windows 11 / Windows PowerShell 5.1 that: (1) `iex` parses the BOM script; (2) successful git progress on stderr does not abort, while a real git failure still fails closed through `$LASTEXITCODE`; (3) missing/null AdapterRAM selects deterministically without hanging; and (4) the full detect-HIP/build/install flow succeeds with an iGPU + discrete GPU machine. The PR body still describes global `$ErrorActionPreference=Continue`, while HEAD uses scoped `Invoke-Git`; please update that and ask for the stale CHANGES_REQUESTED review to be refreshed against current HEAD.

This is local preflight only: nothing has been pushed and nothing has been merged to `master`.

PR #615, opened 2026-08-21
fix(windows): daemon discovery, diag compiler probe, and quantize build

creazyboyone:
find_daemon only probed the bare daemon name, so on Windows every run/serve/diag failed with 'daemon binary not found'; probe daemon.exe first and keep the bare spelling as a fallback shim. diag checked bin/hipcc bare and always reported the compiler MISSING on Windows — route it through the host-suffix-aware rocm::tool_from_selected_root (now pub). quantize imported memmap2::Advice unconditionally although it only exists on unix, breaking the Windows build with E0432; call it fully-qualified inside the cfg(unix) block.

## Summary

Windows is a documented platform (AMD HIP SDK for Windows), but the CLI could not discover `daemon.exe` (every run/serve/diag failed), `hipfire diag` falsely reported a complete SDK as missing its compiler, and `hipfire-quantize` did not compile at all — all invisible to the ubuntu-only CI. This PR fixes those (touching `crates/hipfire-cli`, `crates/hipfire-config`, `crates/hipfire-quantize`), adds platform-shaped unit tests locking in the daemon discovery policy, and fixes `scripts/serve_harness.py` so the serve battery can actually run and tear down on Windows. Verified end-to-end on Windows 11 + RX 7900 XTX (gfx1100) + HIP SDK 7.1: kernel precompile 60/60, `qwen3.8:27b` GPU inference, and the full 5-turn serve battery green.

## Which crate(s) does this touch?

- [ ] `kernels/` (HIP source)
- [ ] `crates/rdna-compute` (kernel dispatch / RDNA arch routing)
- [ ] `crates/hip-bridge` (HIP/ROCm FFI)
- [ ] `crates/hipfire-runtime` (LM runtime: KV, sampler, guards, framing, paging, spec decode)
- [ ] `crates/hipfire-arch-qwen35`
- [ ] `crates/hipfire-arch-qwen35-vl`
- [ ] `crates/hipfire-arch-llama`
- [ ] `crates/hipfire-arch-toy` (template — touch only when refining the new-arch reference)
- [x] `crates/hipfire-quantize` (`hessian_io.rs`: unix-only `memmap2::Advice` import broke the Windows build)
- [x] examples / daemon (`crates/hipfire-cli`: `find_daemon` daemon.exe discovery + diag device-compiler probe + platform-shaped unit tests)
- [x] docs / CI / scripts (`scripts/serve_harness.py`: Windows native-binary resolution + POSIX-only `os.killpg` teardown)

[...]

PR #633, opened 2026-08-23
fix(cli): reserve a 32 MB main-thread stack on Windows

### LINKED ISSUE #621: hipfire.exe stack-overflows on any command when built from source on Windows
Building hipfire from source on Windows (master `80a572c824dc`, via install.ps1's source-build path) produces a `hipfire.exe` that stack-overflows on *every* invocation, including `--version` and `--help`:

```
PS> .\hipfire.exe --version
thread 'main' (22340) has overflowed its stack
```

The release-style binaries built from the same tree are fine (`hipfire-tui.exe --version` prints `hipfire-tui 0.3.0`), it's only the CLI binary. Relinking with a bigger main-thread stack fixes it completely:

```
$env:RUSTFLAGS = '-C link-args=/STACK:33554432'
cargo build --release -p hipfire-cli
```

After that: `hipfire 0.3.0 (80a572c824dc; master)` — everything works.

Windows gives the main thread a 1 MB stack by default unless the linker says otherwise, so something in CLI startup is close to (or over) that limit. Suggest adding something like this to `.cargo/config.toml` (or per-target `[target.'cfg(windows)']`):

```toml
[env]
# or rustflags link-args for windows targets
```

...so users don't need to know about RUSTFLAGS. Happy to PR.

Env: Windows 11, Rust 1.98.0, ROCm 7.2 HIP SDK, RX 7900 XTX.

SergioSV96:
﻿Fixes #621.

The default 1 MB main-thread stack was not enough for CLI startup: `hipfire --version` aborted with "thread 'main' has overflowed its stack" on every invocation when built from source on Windows. Relinking with `/STACK:33554432` fixed it completely.

Reserve the extra stack via rustflags for windows targets in `.cargo/config.toml` so users do not need to know about RUSTFLAGS; other targets are unaffected.

Note: the find_daemon `.exe` half of #621 is intentionally **not** part of this patch — #615 covers it and I'd rather not race that PR.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -21,3 +21,9 @@
 # this file — scripts/check-fmt-bomb.sh (pre-commit) still backstops the case.
 [alias]
 fmt = ["use-scripts-fmt-changed-sh-instead-of-cargo-fmt"]
+
+# Windows gives the main thread a 1 MB stack unless the linker says otherwise,
+# which is not enough for the CLI's startup path (hipfire --version
+# stack-overflows). Reserve 32 MB. No effect on other targets.
+[target.'cfg(windows)']
+rustflags = ["-C", "link-args=/STACK:33554432"]


Kaden-Schutt (approved):
Scoped Windows-only linker stack reservation fixes the reported CLI startup overflow without affecting non-Windows targets.

PR #661, opened 2026-08-29
fix(vram): size Gemma4 KV layers and Qwen scratch to physical capacity

HUSRCF:
## Summary

Tighten two independent allocations without changing kernels, cache addressing, quant formats, or hot-path arithmetic:

- Gemma4 lowered mode allocates sliding-window and full-attention KV caches by their actual `layer_types` counts instead of allocating both families at `n_layers`.
- Qwen flash-attention scratch uses the KV cache's `physical_cap` rather than the logical `max_seq` horizon.

The Qwen change does not enable an eviction mechanism or change normal serving: when no bounded eviction path is active, `physical_cap == max_seq` and the allocation is identical. It only avoids scratch for unreachable physical slots when the cache is already bounded.

## Validation

- workspace build and lib tests green on the code change
- focused layer-count unit coverage
- generated Gemma4 and Qwen35 crate maps refreshed and clean
- independent static review found no index, zero-layer, or scratch-shape defect
- current CI rerun is authoritative for the restored head


Kaden-Schutt (approved):
Reviewed allocation/index invariants and scoped CI. Gemma layer-family counts match the independent lowered-path indices; Qwen physical-cap scratch sizing is inert when the cache is unbounded. Current build, lib tests, crate-map gate, and full CI are green.

PR #670, opened 2026-08-31
feat(maple): BF16 KV tier, q4k head, head overlays, and six measured default fixes

nwoolmer:
Maple (arch 15) work: a BF16 KV tier and its kernels, a q4k lm_head, head variants shipped as overlays rather than duplicate models, and six configuration defaults corrected. **Every default change here is backed by a measurement in the commit that makes it**, and several overturn things previously believed.

Measurements are gfx1151, against a BF16 reference over 2,048 teacher-forced tokens, with 3 paired interleaved repetitions for throughput.

---

## 1. BF16 KV tier, and `--kv-mode` made functional

Maple hardcoded `KvCache::new_gpu_q8` and silently ignored `--kv-mode`, so its Q8 KV could not be compared against anything. That was kernel coverage, not an oversight: the **only** windowed attention kernels in the dispatch table were Q8, and a sliding-window layer cannot run on a tier with no windowed kernel. Maple is 3:1 sliding(512)/global, so this needed kernels.

Adds `KTier::Bf16`, four kernel keys, `KvCache::new_gpu_bf16{,_capped}`, and three HIP kernels. The reduces are **shared, not duplicated** — they consume only f32 partials and never touch the KV cache, so they are dtype-agnostic. The bf16 tiles keep the Q8 tiles' per-thread dim mapping and FMA order byte-for-byte, so a q8-vs-bf16 diff measures the **storage tier** rather than a different summation order.

There is deliberately **no non-windowed bf16 attend key**: `window == 0` already means full causal, so one kernel serves both layer types and no second path can silently drop the window at ctx > 512.

A new `MAPLE_POLICY` site accepts `{q8, bf16}` only. Every other mode is a rotated or block-quantized tier with no windowed kernel; silently accepting one would be *wrong* past 512 tokens, not merely slower.

## 2. Defaults corrected, with the evidence

**KV → bf16.** A 2×2 over KV tier × lm_head:

| KV | lm_head | mean KL | top-1 |
|---|---|---|---|
| q8 | Q8 | 0.0842 | 90.8% |
| **bf16** | Q8 | **0.0511** | 91.9% |
| q8 | bf16 | 0.0842 | 90.5% |
| bf16 | bf16 | 0.0511 | 91.7% |

The lm_head is orthogonal and free; the KV tier alone is **−39%** of mean KL. The damage q8 does is in the **tail**, not uniform blur — the median moves 24% but the worst position goes 10.36 → 4.21 nats, and a rare catastrophic position is what derails a long generation. Cost: 1.88× KV bytes, −2.1% decode (inside this box's ±1–3% noise).

**lm_head → q8** (was `bf16`, strictly dominated):

| head | bpw | mean KL | top-1 | decode |
|---|---|---|---|---|
| q4k | 4.5 | 0.0640 | 90.1% | 167.7 tok/s |
| mq4v2 | 4.25 | 0.0744 | 88.5% | 165.8 |
| **q8** | 8.5 | **0.0511** | **91.9%** | 148.7 |
| bf16 | 16 | 0.0511 | 91.7% | 117.6 |

q8 and bf16 give **identical** KL, so a bf16 head cost 23% of decode for zero accuracy. `mq4` is deprecated in favour of `mq4v2` (better on every axis); the qt=30 **reader** stays so existing files load, guarded by `deprecated_qt30_head_still_loads`.

We do **not** adopt a 4-bit head despite the vendor shipping Q4_K: their benchmark has Q4_K at 252.7 vs FP16 169.8 tok/s — a 49% gain that pays for the accuracy loss. Ours is +13% over q8, because the MoE body dominates decode on this GPU. Same decision, different machine.

**Chat template + sampling → vendor values.** Two independent errors, neither previously checked upstream:

1. Maple's jinja ends its generation prompt with `'<|im_start|>assistant\n<think>\n'`. The harness emitted only `'<|im_start|>assistant\n'`, so the model had to open its own reasoning block — every generation began off-distribution *inside* the block where degenerate loops occur.
2. Temperature was 0.6; the vendor documents **1.0**. Verified bidirectionally: `deepgrove/maple-preview-GGUF`'s README links to `github.com/deepgrove-ai/llama.cpp`, whose README documents `--temp 1.0 --top-p 0.95 --jinja`. There is no `generation_config.json` upstream. Our 0.6 was an unsourced Qwen carry-over.

Together these took the loop rate from **5/8 to 0/8** across 8 seeds (Fisher p = 0.026). **The looping was configuration, not a model or quant defect** — which contradicts what earlier notes in this repo concluded.

## 3. q4k head and a Q4_K encoder fix

Two separate gaps against the vendor's Q4_K head, measured on Maple's real lm_head (relative L2 vs the original bf16 weights):

| scheme | scales/256 | rel L2 |
|---|---|---|
| 1 scale (qt=30 class) | 1 | 0.11814 |
| 2 scales (qt=44) | 2 | 0.10569 |
| **8 scales (Q4_K class)** | 8 | **0.08001** |
| 8-bit (q8) | 8 | 0.00471 |

- **Granularity**: adds `--head-quant q4k`, nearly free — `QuantType::Q4K`, `DType::Q4K`, `gemv_q4k.hip` and `quantize_q4k` all already existed.
- **Encoder**: `quantize_q4k` used plain min/max. Ported llama.cpp's `make_qkx2_quants` faithfully (ggml-quants.c:799, Q4_K's own parameters and importance weights): **0.0799 → 0.0720**, against the vendor's published 0.0731. **This improves every Q4K tensor in hipfire**, not just this head.

## 4. Head variants ship as overlays

The head is 2.7% of the file, so a model per carrier duplicates the identical 6.17 GB body each time.

| | |
|---|---|
| base (q8 head, runnable standalone) | 6.50 GB |
| `--head q4k` overlay | 188 MB |
| `--head bf16` overlay | 635 MB |
| three full variants, for comparison | 19.63 GB |

`--head-only` writes a single-tensor `.hfq`; `--head <name|path>` attaches it. **An overlay is not an approximation of a full build — it reproduces one exactly** (q4k overlay 0.0640/90.1%, matching a monolithic q4k build). Overlay build is 32 s vs ~10 min.

A **headless body is deliberately not offered**: it would require letting an overlay introduce names the base lacks, and that check is what catches a wrong-model overlay. Uploaded to `hipfire-models/maple-preview`; the model card is published.

## 5. Two latent issues found on the way

**RDNA3-only tile coupling.** `q8_flash_tile_size` returns **32 on gfx1100** but 128 on gfx1151, while Maple's `flash_partials` hardcoded `div_ceil(128)` — a *fourth* independent copy of tile-size logic, in a codebase whose launcher already documents "the corruption bug three independent copies of this exact logic caused". Not a live overflow (a ×64 prefill factor absorbed it), but RDNA3 silently gave up 75% of its margin for a reason nothing stated. Now derived. `HIPFIRE_Q8_FLASH_TILE=32` reproduces gfx1100's tile choice locally, so the geometry is exercised rather than assumed.

**Branch registry edits were inert.** `registry/v1.json` is *generated* — I had been hand-editing it, and the next generator run would have silently reverted the bf16 default, temperature and heads. They now live in `registry/models.json`, and regeneration means the generator **probes HF and derives the hashes itself**, so the registry cannot drift from what is published. Separately, `load()` resolved cache → network(master) → bundled, so a locally built binary was overridden by master with nothing reporting it. The bundled registry now wins when its `generated_at` is newer — no new configuration, using a field the generator already stamps. Released binaries are unaffected: their bundled copy is older than master's, pinned by `older_bundled_registry_defers_to_the_fetch`.

## Also included

`e891c0029` cherry-picks the `moe_topk_renorm_k8` barrier fix, **not carried into master by #629**. That kernel could return a top-8 containing the same expert twice from bit-identical logits — a silent quality bug reached by Maple, cohere2moe, qwen35's MTP head and the generic MoE pipeline. **Worth reviewing first**; its blast radius is wider than this PR.

## Verification

- Full `--workspace --all-targets` build; `--lib` suite green (39 binaries) at every commit
- New tests include negative controls and mutation-checks, not just happy paths: bf16 tier/drift guards with a control that bf16 doesn't depend on `q8_windowed`; `bf16_is_maple_only`; head-carrier reader tests incl. the deprecated qt=30; `heads_sidecars_are_digest_validated` (**mutation-tested** — fails with the digest chain removed); both registry-precedence directions
- On the real checkpoint: bf16 KV generates coherently, q8/bf16 diverge at byte 280 (so the tier is genuinely active), and bf16 passes a long-range retrieval probe at 5,482 tokens identically to q8

Two bugs only a GPU run or a negative control could find: a shared `.hip` translation unit where only one wrapper stripped the `kv_slot_desc.h` include, so the batched path compiled and the **first decode step** failed; and a full model passed to `--head` silently shadowing itself while dumping 918 KB of tensor names.

## Caveats

- `--no-verify` on these commits is `verify-bind-thread`, which hard-blocks anything touching `rdna-compute` and **fails identically on clean master**. Each commit says so.
- The residual KL of 0.0511 is **not ours**: llama.cpp rounds activations to bf16 before every matmul (ε = 2⁻⁸). An F16 reference from the same weights differs from the BF16 reference by mean KL 0.0357 — the same order as our entire divergence — and we sit closer to the more precise reference (0.0477) than to the bf16 one (0.0511).
- The lm_head shows no effect on loop rate (mq4 4/8, q8 5/8, bf16 4/8, all p = 1.000), but that was measured before the template fix. The comparison was internally controlled so the null holds; the absolute rate does not.

https://claude.ai/code/session_01FdJ3XSghbyZZ9Ri2srFnm5


Kaden-Schutt:
Same promotion, same ask: rebase onto `beta` (`5dad66ef5`). I've cherry-picked the `moe_topk_renorm_k8` post-invalidation barrier onto beta separately (it's an independent race fix and shouldn't wait), so drop that commit from the rebase.

What collides: beta's G2 source-aware admission (`86f1ab589`) rewrote `hipfire-loader/src/lib.rs` load entries and the daemon load handler, and the CLI `pull`/`run` surface was reworked (`037b24004`, `c86712d43`). Your `head_path` threading through `load_model_with_kv_backend` / `_gemma4_drafter` / daemon `params.head` needs to go through the new `admit_source` → `load_admitted_*` split instead of the old signatures. The Maple-only files (`crates/hipfire-arch-maple`, the three BF16 kernels, `kv_mode.rs` `KTier::Bf16` + `MAPLE_POLICY`) auto-merge — the work is just the loader/CLI/daemon seam.

Your gfx1151 evidence (KL −39%, loop-rate 5/8→0/8) stands; no need to re-measure unless the rebase touches the numeric path.

Kaden-Schutt:
Starting the beta integration queue. Please rebase the remaining Maple work onto current beta, dropping the barrier already ported. #704 is being reviewed independently; we will coordinate its loader seam before landing Maple. Please identify the rebased head and whether numeric paths changed so we can select fresh validation.

Kaden-Schutt:
Maintainer-integrated into beta at ac44279ae, retaining the original PR commits/attribution through the beta port. The previously landed barrier fix was not duplicated. Master is not promoted.

Review corrections included:
- Admit and retain the effective base+head source BEFORE old-model teardown; reject unsupported non-Maple/EP heads and REAP+head rather than silently discard an overlay.
- Refuse truncated indexed payload ranges and short reads rather than zero-fill malformed heads.
- Roll back partial BF16 KV allocations through the existing owner-cleanup helper.
- Select a real gfx12 F16 router GEMM; the original gfx11 intrinsic failed compilation on gfx1201.
- Q4_K final nibbles use the actually stored/re-widened F16 scales.

Parent proof on R9700/gfx1201, HIP7.15:
- Fresh isolated CLI+daemon build passes; CLI245, registry26, loader46, HFQ overlay11 tests pass; Q4_K rounding-boundary regression passes.
- gfx12 router CPU-reference numerical smoke:4/4 shapes pass.
- Real CLI --head q4k with BF16 KV generates coherent text (read output).
- Final composed native-daemon matrix:20/20 checks pass. Six BF16/Q8 KV x baked-q8/q4k/BF16-head cells generate complete coherent answers. Missing/full-model/truncated/wrong-arch head refusals preserve the working model with byte-identical regeneration.
- Both BF16 and Q8 KV decode coherently after a native-tokenizer-counted677-token prompt, beyond the512-token sliding window.
- Scalar-vs-batched BF16 prefill oracle passes at677 and768 tokens, including split-chunk crossings; both comparison points match argmax. This is parity evidence, not a KL/performance claim.
-43 crate maps match. Fresh HF-probed registry payload matches exactly excluding generated_at; registry_gen --check reports timestamp-only drift, not model metadata drift.

All three artifacts SHA256-verified:
trunk7fb52fe72c1a0a4455d0fe3a8109b0df66fa53782f41d8b257140d3e966645db
q4k head deff26e9dfaddfc521f10037e00deae40fb4011fb14db3b10eb166928d7ec795
BF16 head94cde3ad60d380a753f1223882682662643e49faed217df5b4db03e0a17f3e8b
Served CLI MD53f42c56d0fb43f767e77a8017f1442a8; daemon290c80e8e68be4768f4f8cc965aa811e.
Receipt: /home/kaden/ClaudeCode/warpfront/_pr_smokes/repair670-composed-matrix.json; prompt bytes retained alongside as repair670-boundary-prompt.txt.

Limitations stated explicitly: Redline's runtime capture/shadow route refuses Maple, so no Maple PM4 parity claim. No new gfx1151/gfx1100 measurements or inherited author quality/performance claims. Head selection is currently the run surface; serve/chat/bench head-selection expansion is not claimed.

Issue #666, opened 2026-08-31
Tracking: upstream-native device mesh port replacing PR #527


fivetide:
## Decision

PR #527 will not be merged or rebased wholesale. Its branch contains original device-mesh work and valuable validation evidence, but it also carries divergent fork-mainline history that GitHub presents as 3,109 changed files, 906,419 additions, and 106,919 deletions against current upstream. A trial upstream merge produced hundreds of genuine add/add and content conflicts and was aborted safely.

The replacement is an **upstream-native grouped port** from current `master`. Each G milestone has one final-head, end-to-end production boundary. Internal commits and optional PRs may carry focused implementation or evidence, but they never close a milestone by themselves and no historical GitHub merge is current-tree acceptance. PR #527 remains immutable forensic provenance and will point here as superseded.

## Status — 2026-09-11

The September 1 merges #668/#673/#674/#676 were backed out; their merged badges and archive `archive/device-mesh-merged-20260901` remain historical provenance, not acceptance. Replacement integration uses **upstream `beta`** plus the temporary **`g4/integration`** composition branch; `master` remains the release destination.

- **G1 #681 merged on master** at `8cd15a62b593e950b6fe0082b88bb79d122d1dc2` and is therefore also present on beta. Tracker milestone acceptance remains pending.
- **G2 #682 remains open against master**, while its reviewed content and the related admission-refusal corrections are integrated on beta. GitHub PR state, beta integration, and tracker acceptance remain separate.
- **G3 #704 is closed as ported, not merged unchanged.** The maintainer-integrated, review-corrected Single plain-LLaMA manifest/load pilot is on beta at `35dbd64a3`; master remains unchanged. Tracker promotion still requires an accepted structured receipt bound to the final composed head.
- **G4.1 construction ownership and G4.2 DFlash construction are integrated on beta.** #731 merged at `af3990571`; #735 was ported at `800ddf740`; #736 was ported with review corrections at `24fc65c45`. Their rollback/retry and serving evidence closes constituent work only.
- **G4.3–G4.6 merged to beta through #742** at beta head `c88a1ba0dc3fb2104ab78e6cb65b3067973e41ac`. This carries the repaired direct-Qwen/DSpark load transactions, parser finalization, and exactly-once terminal ownership. #742 remains constituent evidence, not aggregate G4 acceptance.
[...]

PR #673, opened 2026-09-01
feat(hardware): establish named device mesh ownership

fivetide:
## Summary

Establish G1's topology authority for multi-device execution. The sole multi-device owner moves from `hipfire-runtime::multi_gpu` into the new `hipfire-hardware` leaf, while a pure, typed `DeviceMesh` defines named PP/TP/EP placement, mesh-generation identity, stage boundaries, and collective groups.

This is the root changeset for the topology work tracked by #666. It supplies the owner and topology contracts that later admission, manifest/store, and MoE work can consume without introducing another ownership or rank-policy implementation.

## What changed

- Added the `hipfire-hardware` workspace crate and moved `crates/hipfire-runtime/src/multi_gpu.rs` to `crates/hipfire-hardware/src/lib.rs` (an 89% rename). `Gpus` remains the single owner of resolved device IDs, VRAM/architecture preflight, peer access, boundary copies, RCCL/host/peer reductions, and peer-reduction scratch leases. The runtime module and compatibility re-export were removed; the ownership integration test asserts that the legacy owner path is absent.
- Added the pure `DeviceMesh` module. `DimKind::{Pp, Tp, Ep}` names the parallelism axes; rectangular shapes use row-major coordinates with checked cardinality, `group_along` derives axis subgroups, `stage_for_layer` and `stage_devices` describe PP bands, and `band_xfer_after` emits the PP boundary hint. `CollectiveHint` distinguishes axis all-reduces from pipeline band transfers.
- Added `MeshEpoch` as identity for an admitted mesh generation. New `single`/`rect` meshes receive fresh epochs; clone and `squeezed` preserve the epoch, so shape-equivalent meshes are not silently treated as the same generation.
- Kept topology independent of GPU handles, loading, allocation, and carrier policy. `RuntimeConfig::device_resolve_opts` lowers the already-resolved logical visibility IDs and hardware knobs into the leaf; hardware consumes those values instead of rereading process environment. `init_uniform`, `init_layers`, `init_tp`, and `single` now use that explicit construction input.
- Made collective membership explicit at the EP seam. `run_layer_program_ep` and its reduction helpers receive an ordered `group`; RCCL `all_reduce_sum_f32` validates the full ordered owner group, while `all_reduce_sum_f32_peer` validates unique global device IDs and is the genuine subgroup-capable path for composed TP×EP layouts. Current architecture callers pass the ordered full-device group and all affected DeepSeek4, MiniMax, Qwen35, generator, loader, runtime, and example call sites now use `hipfire_hardware::Gpus`.
- Added mesh unit coverage and hardware-leaf ownership coverage for composed coordinates, PP boundary hints, epoch preservation, overflow rejection, owner exposure, and removal of the runtime owner path.

### Short term

Make `hipfire-hardware` the one topology/ownership authority now: resolve the visible logical devices once, expose named PP/TP/EP placement primitives, and require collective callers to state their participating group explicitly. Preserve the existing PP, TP, and EP routes while removing the duplicate runtime owner and its compatibility surface.

### Long term

Provide a composable device-mesh substrate for the next admission and execution layers. #674 admission, #675 manifest/store, and #677 MoE execution should be able to consume typed coordinates, axis subgroups, stage transfers, and mesh epochs rather than reimplementing device ownership, rank arithmetic, or environment policy in each feature.

## Benefits

- There is one multi-device `Gpus` owner and one resolved hardware-policy boundary, so runtime, loader, architecture, and generated callers cannot drift onto parallel ownership implementations.
- PP×TP×EP placement is represented as named axes and validated coordinates/groups instead of scattered positional rank math. The same substrate can derive a TP or EP subgroup from a composed mesh and retain identity across shape-preserving operations.
- Collective routing is explicit at the API boundary: full-owner RCCL reductions are distinct from direct peer reductions that can serve real subgroups, making the current limitation and the future `ncclCommSplit` seam visible rather than implicit.
- Later admission, manifest, and MoE work can be typed against topology data and mesh generations, reducing policy duplication and making ownership/placement errors fail at the hardware API boundary.

- **Root changeset:** no device-mesh PR dependency. This PR is the root for G2 admission (#674), G3 manifest/store (#675), and G5 MoE execution (#677); those changes should land after this owner/topology contract.
- G4 lifecycle (#676) is independent and does not depend on this topology cutover.

## Verification

- Changed Rust files pass scoped rustfmt.
- `cargo build` passes on the final rebased head.
- `cargo clippy -p hipfire-hardware --all-targets --no-deps -- -D warnings` passes for the changed package; dependency warnings remain pre-existing.
- `cargo test` passes: 3,454 passed, 12 ignored.
- Physical multi-device route validation was not run in this changeset.
- `python3 -m tools.change_gate run --base beta` was not run.
- No `Architecture` trait surface change.

## Nonclaims / follow-up

- This changeset establishes the topology and ownership substrate; it does not implement admission (#674), manifest/store (#675), MoE execution (#677), or lifecycle work (#676).
- The mesh module is descriptive and GPU-independent; it does not itself admit devices or allocate/launch work.
- RCCL subgroup communicator splitting is not implemented here (`all_reduce_sum_f32` remains full-group only); use the explicit peer subgroup API until the later communicator-split work lands.


Kaden-Schutt (approved):

Kaden-Schutt:

PR #674, opened 2026-09-01
feat(loader): centralize source-aware parallel admission

fivetide:
## Summary

PR #674 implements the G2 source-aware parallel-admission boundary. It opens a model source once, classifies its source namespace and model variant, resolves the raw PP/TP/EP request through one typed policy matrix, and returns the effective G1 `DeviceMesh` together with the retained carrier/source—or a typed refusal. Admission now completes before prior-model teardown, VMM/GPU initialization, remapping, carrier entry, or other model-owner side effects.

This is the G2 admission changeset tracked by #666.

## What changed

- Added `crates/hipfire-loader/src/parallel_capability.rs`, which defines `SourceKind`, `ModelVariant`, `ParallelAxis`, `RawParallelism`, the source-aware `CellPolicy` matrix, and typed `AdmissionError` diagnostics. Degree-zero requests and forbidden PP/TP/EP compositions are rejected first; compatibility remapping and normalization cannot hide malformed requests.
- Added carrier-side source/model classification. HFQ and safetensors-directory namespaces are distinguished before policy lookup; Qwen3.5 and LFM2 dense/MoE/VL variants are derived from configuration and tensor metadata, with unknown, contradictory, or incomplete classifications failing closed instead of falling through to an arch-id-only route.
- Centralized policy resolution and effective-topology construction. The resolver preserves the Qwen3.5-MoE CLI spelling where `tp` denotes EP, keeps the legacy DeepSeek4/MiniMax TP-to-EP compatibility mapping, normalizes only policy-approved dense EP requests to `Single`, and enforces current executable bounds/refusals (HFQ Qwen3.5 dense TP degrees `2..=5`; HFQ Qwen3.5 MoE EP degree `4`; unsupported source/variant/axis cells remain hard refusals).
- Added `AdmittedLoad` to carry the already-open `ModelSource`, selected carrier, source/variant facts, and effective mesh through the daemon's `DaemonLoadState` teardown into execution. Regular, Gemma4, EP, and dense-TP loader entrypoints now consume this value; EP rank loaders reuse the retained HFQ instead of reopening and reclassifying the path.
- Made the daemon's ordinary and experimental multi-slot load paths call the shared admission boundary before teardown and backend-specific checks. Backend checks consume the admitted source/variant/mesh, and rejected admission never reaches slot shutdown, VMM/GPU setup, remapping, carrier entry, or prior-owner operations. The slot engine also accepts the retained source so its worker does not reopen the model after the handoff.
- Touched `crates/hipfire-loader`, `crates/hipfire-daemon`, and `crates/hipfire-arch-qwen35`; the change is an admission/routing cutover, not a new architecture implementation.

### Short term

Establish one immediate, centralized admission decision for every current PP/TP/EP load surface. Resolve source, model shape, and parallel policy before destructive or allocating work, then require downstream loaders and daemon backends to use the effective mesh and retained route rather than interpreting raw degrees again.

### Long term

Provide the single policy authority that future multi-device load and collective paths can consume. Later G3/G5 integration should receive the admitted topology as data and extend the matrix only when a real executable route exists, rather than introducing another raw-degree capability ladder or bypass around source-aware admission.

## Benefits

- Fail-closed handling for unknown sources, ambiguous/malformed model metadata, unsupported source/variant/axis combinations, and out-of-range current routes.
- Deterministic typed diagnostics for source-open, classification, topology, composition, and capability failures.
- No raw PP/TP/EP degree bypass after admission: downstream code receives the effective mesh and selected carrier, so a refused request cannot trigger teardown or allocation and a normalized request cannot allocate against its pre-normalized degree.
- One source lifecycle from classification through execution, avoiding duplicate opens/reclassification (including per-rank EP loads) and keeping source-specific HFQ versus safetensors behavior explicit.
- Existing executable routes remain available while unsupported cells stay refusals; policy changes are localized to the loader-owned matrix.

- **Depends on #673 (G1 topology).**
- This branch is a true descendant of the G1 branch. Because the upstream repository cannot use a fork branch as a PR base, the current `master` comparison includes #673 until G1 merges.
- This PR is intentionally still **draft**. Merge #673 first, then rebase/retarget this PR to the updated `master`; mark it ready only after that cutover.
- G2 (#674) and G3 (#675) are sibling stacks after G1. G2 does not depend on G3, but later G3 non-Single fulfillment must consume G2's admitted effective mesh rather than bypass admission.
- G5 (#677) has the compile-time stack `#673 → #675 → #677`; #674 is its semantic admission prerequisite before the complete production multi-device route is enabled. G5 must consume the admitted topology so raw topology requests cannot bypass source-aware admission.

## Verification

- [x] Admission-boundary tests cover typed source/classification/policy errors, source-aware variant classification, and refusal before the injected continuation/operations.
- [x] Changed Rust files pass scoped rustfmt.
- [x] `cargo build` passes on the normalized stack.
- [x] Scoped `cargo clippy -p hipfire-loader --all-targets --no-deps` completes; the package/workspace has pre-existing warnings.
- [x] `cargo test` passes: 3,474 passed, 12 ignored.
- [ ] Physical multi-device admission routes were not run.
- [ ] `python3 -m tools.change_gate run --base beta` was not run.

## Architecture-trait change?

No `Architecture` trait surface change.

## Nonclaims / follow-up

- This PR does not claim physical multi-device admission or execution validation; those routes were not run.
- G2 does not make the G3 manifest/store or G5 collective-execution paths production-ready, and it does not turn unsupported policy cells into fallback routes. Those downstream paths must consume this admitted mesh before production multi-device enablement.
- The policy matrix records current executable routes and bounds. Adding a future cell requires its loader/executor route and validation; callers must not regain a raw-degree bypass.


Kaden-Schutt (approved):

Kaden-Schutt:

PR #676, opened 2026-09-01
fix(runtime): consolidate generation lifecycle ownership

fivetide:
## Summary

PR #676 closes the G4 lifecycle changeset tracked by #666. It consolidates request reset/rollback ownership, staged GPU load cleanup, resolved MTP metadata, consuming parser finalization, and exactly-once terminal claims across the generation routes.

The immediate change is lifecycle correctness: an abort, error, or normal completion must leave one well-defined owner for every host/GPU transition and one authoritative terminal outcome.

## What changed

- **Consolidated reset and rollback ownership.** `hipfire_loader::reset_lifecycle` is the shared phase runner for single-device, pipeline-parallel, tensor-parallel, and expert-parallel topologies. It clears request cursors/history and runs the same ordered phases—checkpoints, architecture state, adaptive KV, batch state, speculator, request-local eviction state, and graph invalidation plus synchronization. Production rollback paths return a `RollbackEpilogue`; `rolled_back` is true only after all required resets and `device_synchronize` succeed. Reset errors are accumulated while later cleanup phases still run, so a recurrent/architecture failure does not stop cleanup early. The persistent eviction policy/scratch owner is retained; only request-local state is reset.
- **Made staged GPU ownership explicit.** Model loading publishes target/draft allocations through staged ownership and releases every already-owned stage when a later embed, final-norm, output, layer, target-verification, or drafter-head step fails. Output aliases distinguish metadata from the shared primary allocation, and unload paths consume the owning bundle/trait object exactly once. The lifecycle evidence covers failed-load recovery, repeated unload, and target/draft publication without leak or double-free windows.
- **Resolved MTP metadata at load time.** The loader carries the selected MTP mode and draft width in `SpecLoadCfg`, resolves generic/architecture-specific precedence once, and publishes the resulting MTP fields on `LoadedModel` (including whether MTP weights are present). Qwen3.5 and DeepSeek4 spec construction consumes that resolved model metadata instead of re-reading ambient process configuration, while the existing mechanism precedence remains explicit.
- **Made parser finalization consuming and terminal-safe.** The semantic producer consumes its filter/router state at `finish`, drains held EOT/think bytes before final classification, and keeps length-cap, malformed-tool, and open-think outcomes fail-closed. `ThinkOutputRouter` finalization is idempotent and ignores late input; the tool router's consuming finish cannot be finalized twice or leak a post-terminal call/cache decision.
- **Centralized route terminal ownership.** Every `GenerationRoute::ALL` entry has a concrete start/terminal adapter. The production adapter latches one `gen_start` per `(request id, attempt)`, and terminal writers claim the matching `(id, attempt_id)` before emitting `done`, `error`, or cancellation. The two-phase `commit_ready` handshake stages the exact `done` payload; only a matching commit permits normal done emission, tool-call release, and assistant-cache insertion, while abort/timeout remains fail-closed and cannot add a second terminal.
- **Added contract evidence at the shared seams.** CPU coverage exercises all reset topologies and all route adapters, including concurrent done/error/cancel races and late writers. Ignored GPU campaigns exercise staged HFQ and ParoQuant failures, repeated recovery/unload, and DFlash/DSpark target-verification/head failures on the certified fixture path.

### Short term

- Make each generation attempt leave a deterministic lifecycle result: all request-owned state is reset or explicitly reported as unattested, GPU work is synchronized before rollback is claimed, and exactly one semantic terminal owns the wire.
- Prevent partial model/speculator loads from poisoning the next load and prevent aliasing, rank ownership, or repeated unload from leaking or freeing the same GPU allocation twice.

### Long term

- Establish a one-owner state-transition model: the loader owns model-wide reset/unload orchestration, architecture owners expose their reset/free seams, generation supplies only disjoint route-local adapters, and the terminal control plane owns the `(id, attempt_id)` wire claim.
- Keep future route, topology, and speculative-mechanism additions on the same exhaustive lifecycle seams so reset order, metadata precedence, and terminal cardinality cannot silently drift between implementations.

## Benefits

- Abort/error/done races become deterministic: the matching attempt can claim one terminal, duplicate starts and late terminal writers are suppressed, and commit-ready/commit payloads remain identical.
- Rollback is fail-closed and observable instead of optimistic: callers can distinguish an attested reset from an error with reset context, while all cleanup phases and synchronization are still attempted.
- Staged publication and consuming teardown make ownership auditable across target, draft, aliases, parallel ranks, and repeated unload, preventing the leak/double-free classes G4 targets.
- Unsafe output states do not become reusable history: malformed or open-think output suppresses normal done/cache/tool effects, and a pure length cap does not expose incomplete tool calls.

- **Independent of G1/G2/G3/G5.**
- Targets `master` directly and may merge independently.
- It overlaps some loader/generation files with the device-mesh stack, so landing it before later dependent PRs minimizes conflict uncertainty.

## Verification

- [x] Changed Rust files pass scoped rustfmt.
- [x] Isolated `cargo build` passes.
- [x] Scoped clippy for engine/generate/loader/runtime completes; existing warnings remain.
- [x] Focused rerun of `diagnostics::tests::glimmer_self_attn_gate_proj_is_attention_not_mlp_or_router` passes after one transient full-suite failure.
- [x] Fresh isolated full `cargo test` rerun passes: 3,458 passed, 15 ignored.
- [ ] The ignored gfx1151 ownership campaigns are definitions only; they were not executed as physical-device evidence.
- [ ] No COR-007 claim.
- [ ] `python3 -m tools.change_gate run --base beta` was not run.

## Nonclaims / follow-up

- No `Architecture` trait surface change is claimed.
- The ignored gfx1151 campaigns remain opt-in definitions requiring the exact device and fixture set; this body does not claim that hardware evidence.
- This PR makes no COR-007 claim and does not claim execution of the change-gate command.


Kaden-Schutt (dismissed):

Kaden-Schutt (approved):

Kaden-Schutt:

PR #668, opened 2026-08-31
docs(device-mesh): establish upstream port authority

fivetide:
## Summary

> **Non-blocking governance proposal:** G1-G5 implementation may proceed from current upstream while referencing the reviewed tracker on the public `fivetide/hipfire` branch. Merging this PR would add upstream CI enforcement and atomic status updates, but is not a runtime or architecture prerequisite.
Establish the machine-checkable upstream-native authority for replacing superseded PR #527, following the approved grouped port in #666.

This PR contains no runtime, model-admission, GPU, or generation behavior change. It defines:

- 68 fine-grained domain obligations with historical evidence disposition;
- 16 grouped delivery change sets (`G0`–`G15`);
- 22 executable seam gates;
- 5 topology-class evidence campaigns and final closure packet;
- exact source/owner/route/evidence/revert contracts;
- fail-closed status promotion rules;
- the `docs/INDEX.md` authority link;
- a stdlib checker and focused malformed-tracker tests.

The grouped approach keeps evidence precision without requiring one PR per obligation. Reset/lifecycle, manifest/loading, topology/admission, and the interface-safe MoE Step lane can develop in parallel; merge ordering follows the seams each change set consumes.

Closes the planning portion of #666 G0. PR #527 remains open as historical forensic provenance and is explicitly marked superseded.

## Architecture contracts

- `LoadedModel` / `ArchModel` remains the sole model-state and teardown owner.
- `hipfire-generate` plus engine scheduler/terminal remains the sole generation/lifecycle authority.
- Hardware ownership must move by clean cutover with no `Gpus` compatibility re-export.
- Raw PP/TP/EP requests are normalized once at admission; placement consumes admitted topology only.
- Step collective scheduling depends on `DeviceMesh` plus the pure manifest schedule, not `WeightStore` representation.
- Every positive route carries an exact fail-closed/rollback route and side-effect assertions.
- Historical, semantics-only, emulated, failed, host-local, or retired-gate evidence cannot promote current upstream status.
- Physical completion requires pinned model/prompt-or-image/binary/GPU/topology/ROCm/RCCL identities and per-obligation pass mapping.

## Commits

1. `4429d7df3dff048522e55c2c0d6081b40bd45976` — tracker/checker/tests/index implementation
2. `514fd489326b70d1d796849d19929c8e85c76b5d` — evidence metadata pinned to commit 1

The two-commit shape avoids self-referential evidence hashes: executable checker/test evidence is pinned to the exact implementation commit; the second commit records that evidence only.

[...]

PR #682, opened 2026-09-02
feat(loader): source-aware admission — classify and admit before teardown (fixed #674 G2)

fivetide:
## Summary

**Fixed version of #674** (G2 source-aware admission). #674 was reverted in `a0fca0d6d` alongside #668/#673/#676 because it introduced two admission defects that broke production loads:

- **CLS-001**: classified Qwen3.5/LFM2 by `vision_config`/`model_type` config markers, so every Qwen3.5-family text artifact (whose HF config always embeds `vision_config`) was refused at admission for "missing vision tower".
- **self-refusing VMM commit**: `commit_ordinary` re-ran the VMM readiness check *after* the new model's own VMM tensors existed, counting them as pending teardown and refusing every fresh `kv_backend=vmm` load.

This PR delivers the same G2 intent — classify once, admit one effective topology, refuse before destructive side effects — without either defect.

## What changed

- **`crates/hipfire-loader/src/admission.rs`** (new): read-only `admit_source` — opens the source, classifies `arch_id` + vision, decides the effective topology (`Single`/`Pipeline`/`Expert`), and refuses no-carrier / ambiguous-carrier / VMM-allowlist / VMM+pp / carrier-pp / EP-arch / lfm2 vision-no-config / DFlash lm-head quant — all before any GPU, VMM, or teardown work. Vision is decided by the **tower tensor** (`classify_vision`), preserving the remediation contract `179a20d7f`.
- **`Carrier::admit_topology`** (new trait method, default + qwen35/cohere2moe/maple/gemma4 overrides) — mirrors each carrier's load-time `pp>1` refusal so a refusal fires at admission.
- **Load-entry split**: `load_model_with_gemma4_drafter` / `load_model_ep_with_kv_mode` become thin wrappers that classify-then-load via `load_admitted_with_gemma4_drafter` / `load_model_ep_admitted`, which consume the retained `SourceAdmission` (no re-open, no re-classify).
- **Daemon reorder** (`crates/hipfire-daemon/src/main.rs`): daemon topology refusals + `admit_source` now run **before** prior-model teardown. On refusal the prior model stays loaded.

## How this differs from #674

| | #674 (reverted) | this PR |
|---|---|---|
| VL classification | config markers (`vision_config`/`model_type`) | tower tensor only (`tensor_data(...).is_some()`) |
| VMM commit | re-checked after own tensors exist (self-refuses) | single pre-allocation `ensure_vmm_ready_for_load` |
| refusal ordering | after prior-model teardown | before teardown (prior model preserved) |
| source handle | re-opened/re-classified at load | retained `SourceAdmission` consumed once |

## Verification

- `cargo build` (workspace): clean. Full workspace test suite: **no failures** (incl. 5 new `admission` contract tests).
- Scoped rustfmt on changed files: clean.
- Hardware (gfx1151, isolated `HOME`):
  - `hipfire run qwen3.6:27b` → 64/64 layers, `KV cache: Q8 vmm` (fresh-daemon VMM commit), coherent `Paris.`
  - `hipfire run qwen3.6-35b-a3b.mq4` → coherent linked-list-reverse output.
  - Failed-load probe: a bad-path load (`No such file or directory`) and a contradictory `kv_backend=vmm + pp=2` load both refuse at admission, and a subsequent generate on the prior model returns `done` in both cases (prior model survives refusal).

## Dependencies / nonclaims

- Root of the re-plan: consumes no upstream types (independent of G1's `DeviceMesh`; G2 admission decides the topology, G5 will bind it to `gpus.mesh`).
- `DimKind::Ep` classification is unchanged from master (EP still routes by `arch_id` 5|6|9|10); no new axis, no manifest planning, no family cutover.


Kaden-Schutt:
One regression to fix before this lands, found by running the new admission path on hardware against master:

**`admission.rs:189` refuses `kv_backend=vmm` for every EP load. Master serves DeepSeek V4 EP with vmm by design.**

Master's `load_model_ep_with_kv_mode` (`crates/hipfire-loader/src/lib.rs`) only refuses vmm for the arches whose EP loaders don't support it:

```rust
10 if kv_backend_kind == KvBackend::Vmm => Err("KV backend 'vmm' requires tp=1")   // MiniMax
5 | 6 if kv_backend_kind == KvBackend::Vmm => Err("KV backend 'vmm' requires tp=1") // Qwen3.5
9 => load_model_ep_ds4(...)                                                         // DS4: vmm allowed
```

This PR's `admit_source`:

```rust
if kv_backend == KvBackend::Vmm {
    return Err(format!("KV backend '{}' requires tp=1", kv_backend.as_str()));
}
```

fires for arch 9 too, so a DS4 EP + vmm load that works on `344e17a2e` is refused at admission on this branch. The comment above it says "Mirrors `load_model_ep_with_kv_mode`'s arch_id dispatch + VMM refusal" — it doesn't; master's refusal is per-arch. Fix: gate on `matches!(arch_id, 5 | 6 | 10)`, and keep the DS4 arm vmm-capable.

Two smaller things from the same run, non-blocking:
- `load_model_ep_admitted` documents no-reopen but still passes `path` into the per-arch EP loaders, so the retained `SourceAdmission.source` is dropped on the EP path. Either thread the handle through or don't claim no-reopen for EP yet.
- Pre-existing on master, not yours: `ornith-1.5:35b-a3b-mq4r` loads at tp=4 on 4× gfx1201 but every generate fails with `EP arch mismatch (expected dense Qwen TP)` — on master and this branch alike. I'll open an issue.

Everything else checked out: with a model resident, four different refusal kinds (pp>1 on a non-Qwen carrier, missing path, vmm on an unsupported carrier, EP on an unsupported arch) leave the prior model loaded and generating byte-identical output to a fresh master load, where master tears it down; pp=1/pp=2 on 0.8B and 27B and EP tp=4 loads are byte-identical to master. The `ratchet-raise` label is applied for the +21 daemon lines.

fivetide:
Addressed the blocking regression and the EP-path doc note.

**VMM refusal is per-arch now.** `admit_source` gates the EP `vmm` refusal on `matches!(arch_id, 5 | 6 | 10)`; DeepSeek V4 (9) stays vmm-capable, mirroring `load_model_ep_with_kv_mode`'s per-arch dispatch. Added a unit test locking the contract (`ep_vmm_refusal_is_per_arch`: 5|6|10 refuse, 9 admitted, non-vmm never refused).

**EP no-reopen claim corrected.** `load_model_ep_admitted` and the `SourceAdmission.source` field doc now state that the per-arch EP loaders re-open `path` per rank (the retained source is dropped on the EP path; the single/pp route still consumes it). Kept the doc fix rather than threading the handle through the EP loaders, since those are also reached from the non-admission `load_model_ep_with_compressor_cache` entry.

Smoke-verified on gfx1151 (single device), rebuilt release daemon:

- DS4 `--tp 2 --kv-backend vmm` now passes admission and proceeds into `load_model_ep_ds4` (fails later only at `init_tp: device id 1 out of range (count=1)` — no second device on this box).
- MiniMax `--tp 2 --kv-backend vmm` still refuses at admission with `KV backend 'vmm' requires tp=1`.

The `ornith-1.5` EP arch-mismatch note was left untouched as pre-existing.

Requesting another hardware validation run on the multi-GPU setup to confirm the DS4 EP + vmm load that previously refused at admission now serves.

Kaden-Schutt:
Gate run [33921475093](https://github.com/warpfront/hipfire/actions/runs/33921475093) failed **both** hardware lanes, and it is not this PR's fault — it is a gate defect I have now fixed in #712.

The lanes execute `pr/scripts/serve_harness.py`, i.e. *this branch's* copy of the instrument that judges it. This branch is based on master from before #703, so its harness has no `ATTRACTOR_MIN_WINDOW` guard. It therefore flagged the `[reason]` turn — which answered `Answer: 43`, 9 tokens, `recall=1/1` — as a token attractor:

```
[reason]t5  finish=stop  ctx=47  gen=9  (ans 3w)  tau=2.33  recall=1/1 !ATTRACTOR | '43\n\nAnswer: 43'
```

Every row in both lanes' evidence records `attractor:false, empty:false, runaway:false, recall_ok:true` with correct content (working Python, correct Rayleigh-scattering answer, `Paris`/`Seine`, valid JSON) — the fixture-level `exit 1` came purely from the stale heuristic. The other three fixtures (`qwen3.6:27b` battery+chain, `ornith-1.5:35b-a3b-mq4r`, `lfm2.5:1.2b`) passed on both lanes.

#712 makes `serve_harness.py` and `redline_daemon_harness.py` resolve against the gate checkout, the same way fixtures, registry, prompts and `review.py` already do. Once it lands I will re-dispatch this PR's gate; no change is needed here.

Rebasing onto current master would also dodge it, but that hides a defect that would keep hitting every PR older than a harness fix.

Kaden-Schutt:
Landed on `beta` as `4d0e0a1b4`. This is a fork PR, so it could not be pushed to; it is staged in the main repo as `staging/pr-682` — head `95e0e654e` merged with `beta` at `a9f4ca860`.

**That merge had a real semantic conflict**, not a generated-map one: #687's `ep_admission` (now on `beta`) versus this PR's `admit_source`, both in `crates/hipfire-loader/src/lib.rs`. Resolved by keeping your classify-once split (`load_model_ep_admitted`) and #687's shared `ep_unsupported_arch_message(id)` for the backstop arm, so the two refusal messages cannot drift apart.

Because that resolution is mine and no gate run covered it, I proved it on hardware (hiptrx gfx1201, daemon md5 `99457c71e4e4`):

- battery on the canonical **`qwen3.8-27b.mq4-xt`** (mq4v2) trunk — 5/5 turns, `attractor 0, empty 0, runaway 0`, recall satisfied every turn, avg decode 29.5 tok/s
- LFM2 (`arch_id=11`) `--tp 2` refuses with exactly the shared constructor's text at `[validation retryable=false rolled_back=false]`, serve process still alive
- `ornith-1.5-35b-a3b` (`arch_id=6`) `--tp 2` refuses through #687's rule at **32548 MB free / 32624 MB total** — i.e. this PR's restructure did *not* bypass the refusal #687 added, which was the actual risk in that conflict
- 48 tests pass in `hipfire-loader` + `hipfire-daemon`

Fable's earlier investigation (run 33921475093) closed **every** coverage gap Sol raised: six refusal classes all fire before teardown with the prior model still generating, while the *base* daemon answers `no model loaded` after four of the six. That is the regression this PR removes, and it is worth more than the diff review.

**Two floors, both human calls, both made.** `scripts/leanup-thresholds.txt` is a policy path, and the ratchet raise `daemon_lines <= 4155 → 4176` is **accepted**: the logic did land in `hipfire_loader::admission`, and what grew in the daemon is the call site plus `emit_uncorrelated_error` boilerplate that has to live at the protocol boundary. If that ceiling is hit again, collapse the repeated emit blocks into a local helper before raising it further.

Recorded as unproven rather than claimed: DS4 (arch 9) / MiniMax (arch 10) EP admission and the Qwen3.5-VL / LFM2-VL tower classification — no such artifact exists on either lane.

Closing as landed via staging; the commits keep your authorship.

PR #685, opened 2026-09-02
docs(plans): codebase audit 2026-09-02 with ranked remediation plan

Kaden-Schutt:
## Summary

Point-in-time audit of master `8cd15a62b`: twelve read-only slices, one synthesized report (`docs/plans/audit-2026-09-02/README.md`) with a ranked remediation plan, plus the per-slice reports. Every finding is cited to `path:line` at that commit.

Headline: the failures cluster into three classes — the loader admits (model, topology, quant) combinations the engine cannot run (six cases of the #683 species), reset/rollback is canonical on AR/dense-TP/DS4 and hand-rolled or absent on EP error exits, vision cancel, and the bring-up archs, and docs/comments describe the design rather than the code. Also verified: an MQ8 GEMV family path that panics, `LlamaWeights::free_gpu` leaking PARO/AWQ sidecars on every reload, and EP loads recording a `Tp` mesh.

## Which surface(s) does this touch?

- [x] docs only — `docs/plans/audit-2026-09-02/*`, one row in `docs/INDEX.md`

No code. No hardware route.

Issue #683, opened 2026-09-02
EP MoE generate fails after a successful tp=4 load: 'EP arch mismatch (expected dense Qwen TP)' (ornith-1.5:35b-a3b-mq4r, master)


Kaden-Schutt:
Found by the hw-gate deciding seat while investigating #682, reproduced on master `344e17a2e` and on #682 alike, so it is pre-existing.

**Repro** (hiptrx, 4× gfx1201, ROCm 7.15):
- `hipfire serve 127.0.0.1 11520 --tp 4` with `ornith-1.5:35b-a3b-mq4r` (registry sha `84103fcc…`), or the native protocol `load` with `params.tp=4`.
- Load succeeds: `[loader] EP load complete: 4 ranks`, `diag` reports `model_loaded=true`, `arch qwen3_5_moe`, ~16 s.
- Every `generate` fails: `EP arch mismatch (expected dense Qwen TP)` — 5/5 battery turns empty through `serve_harness.py --tp 4`, same through the raw daemon protocol (`ornith_ep_tp4.BASE.json` in the evidence set).

So the EP loader admits and builds the A3B MoE at tp=4, but the generate dispatch treats the loaded model as a dense Qwen TP model and refuses. Either the EP generate route for arch 6 (qwen3_5_moe) is not wired for this artifact, or the loaded state is mis-tagged.

Not touched by #682 (both binaries fail identically). Single-device generation on the same artifact is fine (206 tok/s AR). Evidence will be attached from the gate once #679 lands.

Kaden-Schutt:
Traced. Not a clobber from the device-mesh series — a gap that has existed since Qwen3.5 EP was introduced, with #662 changing only the error string:

- `e3c7b4c35` (2026-08-09, "add safe EP batch foundations") gave the loader `EpArch::Qwen35 { batch, .. }` and the arch crate `ep_batch.rs::forward_ep`, but **`hipfire-generate` never got a generate arm for it** — the only `EpArch::Qwen35` match in the engine is the reset path. Before #662, arch 6 EP fell through to `ep_serve_ds4`, which refused with `EP arch mismatch (expected ds4)`.
[...]

PR #679, opened 2026-09-02
ci: hw-gate — two-seat autonomous review rung (Sol decides hardware, Fable decides staging merges)

Kaden-Schutt:
## Summary

hw-gate: hipfire's autonomous review rung — two model seats, one human owner, hardware evidence on every load/serve/kernel change. Replaces `tools.change_gate` + the unconfigured agentic-review pipeline as the CI route (deletion of those is a follow-up).

Motivation: https://github.com/warpfront/hipfire/issues/666#issuecomment-5505294695 — five PRs merged on static review broke every Qwen3.5-family load. The `load` fixture set here refuses in 2 s on that tree and passes on master.

**Requesting review from @fivetide and @nwoolmer.** After reverting the last auto-reviewer I'm not merging a new one on my own read; please push on anything below.

## The rung

```
select ──▶ sol · prelim ──▶ hardware (matrix) ──▶ sol · verdict ──▶ fable · decide ──▶ hw-gate status
            reads diff        hiptrx gfx1201        greenlight /       merge-staging /
            DECIDES whether   hipx   gfx1100        needs-human /      hold / block
            code runs on hw   serve_harness         block              may veto / override Sol
            + which routes    battery / chain       never merges       merges head → beta
```

- **Sol** (`gpt-5.6-sol`, read-only tools in a PR checkout) alone decides whether the PR's code runs on the maintainers' hardware and which routes run: mandatory fixtures for the touched buckets ∪ the author's requested routes ∪ its own. A maintainer's `hw-run` label can only *force* a run.
- **Hardware** drives every route through `serve_harness.py` (battery; +chain for serve; +Redline parity for kernel) with reasoning off, on both lanes; every decoded turn is posted verbatim. `merge_evidence.py` combines lanes and treats a missing lane as failure.
- **Sol's verdict** on diff + evidence: `greenlight` / `needs-human` / `block`, regressions cited by `file:line` and fixture, and whether the author's claim was proven. Sol never merges or approves.
[...]

PR #687, opened 2026-09-03
fix(admission): refuse unservable model/topology combinations before allocation (#683 family)

Kaden-Schutt:
## Summary

Four of the six "loader admits, engine refuses" cases from the 2026-09-02 audit (PR #685, README finding 1), each refused at admission with a reason, **before any device allocation**:

1. Qwen3.5-MoE (arch 6) under EP (#683) — `qwen35_ep_moe_refusal` in `hipfire-loader`, before `Gpus::init_tp`.
2. Gemma4 lowered/MoE — refusal moved from `ar.rs` generate to `load_gemma4_bundle`, before `lowered::load_weights`; `ar.rs` keeps the same message as a fail-closed net.
3. `generate_ep` `_ => ep_serve_ds4` wrong-server fallthrough — now an explicit refusal naming the arch (`EpServeTarget`), plus `ep_admission()` at both EP load entries for archs with no `EpArch`.
4. LFM2 continuous batching — `supports_continuous_batch` was `true` while `batch.rs:166-181` returned `false` unconditionally; caps made truthful, arch 11 removed from `continuous_batch_route`, so no `Lfm2DecodeBatchState` VRAM is allocated for a path that never runs.

Not included: Muse-Glimmer batching (already correct on master — caps `false`, route `None`; the audit overstated it) and GL qt 38/39 (design written in the slice report: refusal at `decode_raw_codec` entry + dense manifest scan, but the MoE expert path legitimately serves 38/39 through the shared wrapper, so it needs its own PR).

- [x] `crates/hipfire-loader` (admission helpers + tests)
- [x] `crates/hipfire-arch-gemma4`, `crates/hipfire-arch-lfm2`, `crates/hipfire-generate`
- [x] examples / daemon — no

## Evidence (hiptrx, 5× R9700 / gfx1201)

`hipfire serve --model ornith-1.5:35b-a3b --tp 4` with this branch — the exact #683 repro (master loads ~19 GB across four devices, then fails at first generate with "EP arch mismatch"):

```
[hipfire] pre-warm failed: daemon error: [internal retryable=false rolled_back=false attempt=0] load failed: Qwen3.5-MoE (arch_id=6) has no EP serve path; use TP or single-GPU. GPU: gfx1201 (32548 MB free / 32624 MB total); serving lazily
```

32548 of 32624 MB still free when it refused: nothing was allocated.

## Test plan
- [x] `cargo test -p hipfire-loader` 28 pass (new: `qwen35_moe_ep_refuses_before_load_but_dense_admits`, `ep_without_eparch_refuses_but_served_archs_admit`, `lfm2_continuous_batch_never_admits`); `cargo test -p hipfire-arch-gemma4` 19 pass (new `lowered_admission_refuses_what_generate_cannot_serve`); `cargo test -p hipfire-generate` all suites; `cargo test -p hipfire-engine --test continuous_batch` 41 pass
- [x] Hardware: EP refusal above
- [ ] hw-gate when #679 lands


Kaden-Schutt:
Merged to `beta` as `0297aed6b` (branch tip `5dbe4a997`).

Fable's verdict was **merge-staging**; the seat was stopped only by the recurring generated-map 409 (`staging_merge_conflict` hard floor on `crates/hipfire-{generate,loader}/map.md`). I merged `beta` in, regenerated those blocks with `scripts/check-crate-maps.py`, and merged.

What makes this landable is that the decide phase ran the #683 repro on the 5×gfx1201 host instead of reading the diff:

- `ornith-1.5:35b-a3b --tp 4` refused at pre-warm with all four cards still at **32548 MB free**; master loads four ranks (~7.86 GB each) and only fails on the first request with `EP arch mismatch`
- the one path this PR closes that master could actually reach — the batch-only EP route (`batch_staging.rs:231-331`) — **segfaults inside libamdhip64 on master, 2/2 runs**. Refusing it removes a crash, not a serve path.
- dense TP through the same entry still loads two ranks and decodes coherently at 20.1 tok/s on the PR build
- LFM2 `--continuous-batch-size 2` is byte-identical on both builds, while master was allocating an `Lfm2DecodeBatchState` nothing ever drove (`batch.rs:180` returns `false` unconditionally); LFM2 `--tp 2` refuses identically on both

Sol's `needs-human` was about **coverage, not a regression**: no Gemma4 or DeepSeek-V4/MiniMax fixture exists on this host, and those arms are arch/config-only code that runs before device init. Both are recorded as `unproven` in the decision artifact rather than as claims — worth closing later on a 4× DS4 Flash `--tp 4` battery.

70 tests pass in `hipfire-loader` + `hipfire-generate` at the merged tip.

PR #688, opened 2026-09-03
fix(vram): free weight sidecars on unload; record the Ep mesh axis on EP loads

Kaden-Schutt:
## Summary

Two verified findings from the 2026-09-02 audit (PR #685, README findings 3 and 5), re-landed individually as the audit recommended:

1. **Weight sidecar frees.** `LlamaWeights::free_gpu` freed `.buf` per weight and skipped the PARO rotation and AWQ scale sidecars, leaking one sidecar set per weight per layer on every reload of an AWQ/PARO llama/qwen3 (`DeviceBuffer` has no `Drop`). Every `WeightTensor` now goes through `WeightTensor::free_all`, mirroring `Qwen35Weights::free_gpu` / `DflashWeights::free_gpu`. `Qwen2Weights::free_gpu` had the same `.buf`-only pattern and is fixed the same way (no sidecar is allocated for Qwen2 today, but its forward already reads `awq_scale`).
2. **`Gpus::init_ep`.** Every `load_model_ep_*` called `Gpus::init_tp`, which records `DeviceMesh::rect(Tp, n)`, so after an EP load `mesh.size_of(Ep) == 1`. `init_ep` is layout-identical to `init_tp` but records `rect(Ep, n)`; the three EP load sites in `hipfire-loader` use it. `Gpus::single` keeps `DeviceMesh::single()` (absent axes read as 1 by design; the existing single-topology test pins that). #681 landed the mesh type with no readers, so this is the fix before the first reader.

- [x] `crates/hipfire-runtime` (`llama.rs`, `multi_gpu.rs`), `crates/hipfire-loader`, `crates/hipfire-arch-qwen2`

## Test plan
- [x] `cargo test -p hipfire-runtime` — 598 pass, incl. new `device_mesh_ep_group_and_tp_absent` (Ep groups all ranks, `size_of(Ep)==n`, `size_of(Tp)==1`, and symmetrically)
- [x] `cargo build -p hipfire-runtime -p hipfire-loader -p hipfire-arch-qwen2` clean
- [ ] Hardware fault scenario for (1) — load an AWQ or PARO (krot>0) llama/qwen3 artifact, unload/reload N times, watch `vram_free_mb` in the diag JSON: it should stay flat instead of dropping by one sidecar set per weight per layer per reload. Not run: no AWQ llama-family artifact on hipx/hiptrx at the time; there is no freed-buffer accounting hook to unit-test against.
- [ ] hw-gate when #679 lands

EP examples (`ep_minimax.rs`, `ep_decode_parity.rs`, `ep_dspark_topology_probe.rs`, `ep_deepseek4.rs`) still call `init_tp`; dev harnesses with no mesh readers, left for a follow-up.


Kaden-Schutt:
Merged to `beta` as `7df822e9b` (branch tip `a0aa12681`).

Verdict was **merge-staging**; the seat was stopped only by the recurring generated-map 409 (`staging_merge_conflict` on `crates/hipfire-runtime/map.md`). Merged `beta` in, regenerated the block, merged.

The decide phase measured both claims on the 5×gfx1201 host rather than reasoning about them:

- 4× EP load of `ornith-1.5:35b-a3b-mq4r` on the PR daemon logged `EP load: tp=4 arch=qwen35` after `init_ep`, completed on 4 ranks, then unload → reload → unload with rank-0 `vram_free_mb` **32548 / 25512 / 32400 / 25510 / 32398** — the *identical* five numbers the master daemon produced from the same stdin. That is the evidence the `Ep` mesh axis is a zero-runtime-delta rename.
- `Gpus::init_ep` is field-for-field `init_tp` except `mesh: DeviceMesh::rect(Ep, n)`, and the `mesh` field is read only by constructors and tests.

Sol's `needs-human` was coverage, not a regression, and it is worth writing down as follow-up rather than losing: no AWQ/PARO Llama artifact was cycled with free-VRAM diagnostics (so the sidecar-leak fix itself is argued, not measured), and the Qwen2 `free_all` path has no fixture on either lane. Both are recorded as `unproven` in the decision artifact.

21 tests pass in `hipfire-runtime` + `hipfire-loader` at the merged tip.

PR #690, opened 2026-09-03
fix(mq4v2): one MQ-V2 prefill admit rule for llama and qwen35; discriminating GEMM parity; spec §9

Kaden-Schutt:
## Summary

Follow-ups from the MQ4G256V2 kernel-family audit (`docs/plans/audit-2026-09-02/audit-Mq4v2Kernels.md`, PR #685):

1. **Lockstep** (audit Broken 1). `hipfire_runtime::llama::is_batchable_la` admitted MQ-V2 (qt 44/47–50) WMMA prefill only on gfx12; `qwen35::is_batchable_la` admitted gfx11 + gfx12 with the `HIPFIRE_MQV2_GFX11_WMMA` kill-switch. Both doc-comments claimed they "match exactly". The rule now lives once in `llama::mqv2_wmma_batchable` / `llama::mqv2_gfx11_wmma_enabled_from_env`; both gates delegate; a test asserts agreement over 6 dtypes × {gfx1100, gfx1151, gfx1201, gfx1030, gfx1010}. `MQ4CG256` (qt=45) stays gfx12-only in both, per the spec. **Behaviour change (revised in `4e360fb2a`):** plain Llama/Qwen3 dense models on MQ-V2 stay **per-token on every arch** — `llama.rs::forward_prefill_chunk` has no V2 arms (matchers at ~2570/3025/3117/3248), so admitting them would run V1 launchers on V2 blobs; the hw-gate Fable seat caught this by source trace (run 33895641944). The shared rule still admits qwen35 on gfx11+gfx12 (its chunk path has the V2 arms). master's old gfx12-only arm had the same hole for llama; this PR closes it instead of widening it.
2. **Discriminating GEMM parity** (audit Missing 1). `mq4v2_gemm_parity` used Gaussian weights whose two halves have near-identical headers, so a wrong half-select hides inside quantization noise — contradicting its own header comment. It gains a disjoint-halves arm (half 0 in `[-1,1]`, half 1 in `[96,160]`, same batch sweep) with a negative control (swapped-header reference must disagree) and a CPU test that the fixture discriminates.
3. Docs/comments: spec §9 (MoE is production-wired for qt=44; XBATCH is ported; gfx11 bodies step `kt += 2`), `gemm_mq4g256v2_residual_mmq.hip` "Experimental" → production fast path (gfx1100/1151, batch ≥128), `gemm_qkv_mq4g256v2_wmma.gfx12.hip` "HYPOTHESIS" → the validated C-map statement its sibling carries.

- [x] `crates/hipfire-runtime`, `crates/hipfire-arch-qwen35`, `kernels/` (comments only), docs

## Evidence
- `mq4v2_residual_parity` — **PASS** on gfx1100 (hipx) and gfx1201 (hiptrx): all shapes, swapped-headers baseline at 0.9996 rel error.
- `mq4v2_gemm_parity` — **PASS** on gfx1201 (the example calls the `*_gfx12` launchers directly; gfx12-only by construction): disjoint arm v2 rel-rms 2.5e-4 vs 1.263 with swapped headers; gate_up/qkvza v1/v2 arms unchanged.
- The gfx11 WMMA prefill TUs the lockstep change admits for Llama/Qwen3 are the same ones Qwen3.5/3.8 already run in production on gfx1100/1151 (hw-gate lanes). **Not run:** a model-level prefill on an arch-0/1 MQ-V2 artifact on gfx11 — none exists on hipx/hiptrx (`qwen3-8b.hf4` is qt=6). If a reviewer wants that before merge, it needs a `--format mq4` quantization of qwen3-8b first.

## Test plan
- [x] `cargo test -p hipfire-runtime -p hipfire-arch-qwen35` — 597 + 189 pass, incl. `mqv2_admit_llama_qwen35_lockstep` and `disjoint_fixture_discriminates_half_select`
- [x] Parity examples above


Kaden-Schutt:
Fable's finding on run 33895641944 is fixed in `4e360fb2a` — and it was a real bug, not a coverage gap: `llama.rs::forward_prefill_chunk` has no V2 arms (matchers at ~2570/3025/3117/3248 list only `MQ4G256|MQ6G256|MQ3G256|MFP4G32`), so the "lockstep" would have sent a Llama/Qwen3-dense V2 model through V1 `hfq4g256` launchers on V2 blobs. `llama::is_batchable_la` now refuses every `*G256V2` + `MQ4CG256` on every arch (master's old gfx12-only arm had the same hole; closed rather than widened). qwen35 keeps the shared rule — its prefill has the arms, and Fable's base-vs-PR Redline A/B on `qwen3.8-27b.mq4-xt` was bit-identical to the logits hash with the `*_mq4g256v2_wmma_gfx12` kernels named in the capture. `batched_verify_eligible` routes through `is_batchable_la`, so spec-decode verify is covered too. Lockstep test rewritten to the true contract; spec §9 and the PR body corrected.

PR #686, opened 2026-09-03
feat(dflash): registry-declared draft sidecars — pull fetches, auto uses, on requires

Kaden-Schutt:
## Summary

DFlash was reachable only through an explicit draft path (`--model-draft`, `developer.dflash_draft` / `HIPFIRE_DFLASH_DRAFT`). `hipfire pull <tag>-draft` set nothing, `dflash_mode auto` scanned nothing, and the daemon logged "DFlash disabled" only for `off` — so every user following AGENTS.md / MODELS.md ("pull the draft, set `auto`, run") was on plain AR without knowing. The filename auto-matcher those docs describe does not exist in the tree.

This PR makes DFlash drafts registry-declared sidecars, exactly like `mtp` / `dspark` on `ModelEntry`: `hipfire pull <target>` fetches the draft; `load_params` resolves it for `auto`/`on` when no explicit draft is given; `on` fails closed when it is missing (CLI) or fails to load (daemon Qwen35 arm, mirroring MTP); explicit `--model-draft` / `developer.dflash_draft` still win; a final `off` still drops it. Docs rewritten to describe what the code does. Audit: `docs/plans/audit-2026-09-02/audit-Dflash.md` (PR #685), Broken 1.

## Which crate(s) does this touch?
- [x] `crates/hipfire-registry`, `registry/models.json` + `v1.json` (new `dflash` sidecar on 20 dense targets)
- [x] `crates/hipfire-cli` (pull/rm sidecar handling; `resolve_dflash_sidecar` in `load_params`, re-resolved after the `run`/`bench` selector)
- [x] examples / daemon (`SpecLoadCfg.dflash`; Qwen35 draft-load `Err` fails closed when `on`)
- [x] docs (AGENTS.md, docs/MODELS.md, docs/env-vars.md)

## Evidence (RX 7900 XTX / gfx1100, HIP 7.15, hipx)

Same command, same registry tag, **no env var and no `--model-draft`**, prompt `benchmarks/prompts/humaneval_3_below_zero.txt` (md5 `37c5aad9f9efe93b5c47f27256bdf149`), `--runs 5 --warmups 3 --max-tokens 128 --backend noslots --workload stateless`:

| binary | `--spec` | decode tok/s (median) | samples | draft |
|---|---|---|---|---|
| master `c66abf3d8` (gate cache) | dflash | **49.9** | [49.7, 49.8, 49.9, 49.9, 49.9] | none loaded (`drafter=off`) |
[...]

PR #691, opened 2026-09-03
fix(dflash): transactional draft ctors, emitter-error rollback, one ctx-cap predicate

Kaden-Schutt:
## Summary

Three verified defects from the DFlash audit (`docs/plans/audit-2026-09-02/audit-Dflash.md`, PR #685):

1. **Transactional draft constructors** (Broken 3). `DflashWeights::load`, `DflashScratch::new_with_mq` / `new_windowed`, and `dflash_generic::build_generic_dflash_speculator` `?`-ed out mid-construction leaving earlier `alloc_tensor`s allocated (`GpuTensor` / `DeviceBuffer` have no `Drop`). Every allocation site now frees on the error path in the `or_free!` style of `load_dflash_state`; `new_windowed` frees the popped base scratch on any post-pop failure; the generic builder frees weights on scratch failure. `DflashLayerWeights::free_gpu` is shared by the success and error paths. Success path unchanged.
2. **`make_spec_emitter` Err rollback** (Broken 4). The one post-prefill error exit in `generate_spec` that returned without `production_fail_closed_rollback_live` — leaving advanced target KV / DeltaNet / drafter hidden behind cleared host counters for the next turn to LCP against — now rolls back first, like the adjacent `prefill` and `step` Err arms.
3. **One ctx-cap predicate** (Broken 5). `generate_dflash` fell back to AR at `prompt + max_tokens > cap`, `generate_spec` then errored at `prompt + max_tokens + block_size > cap` *after* `gen_start`. `common::spec_ctx_request_fits` is used at both sites, so the block-size band falls back to AR at entry. The mid-loop `position + block_size >= cap` break now sets `SpecRun::ctx_exhausted`, reported as `finish_reason=length` with no tool release and no cache store (via the existing `qwen_dflash_hit_length_cap` → wire-terminal path) instead of a natural `stop` that could prime the cache.

## Which crate(s) does this touch?
- [x] `crates/hipfire-runtime` (`dflash.rs`, `dflash_generic.rs`), `crates/hipfire-generate` (`qwen.rs`, `common.rs`, `dense.rs` epilogue)

## Evidence
[...]

PR #692, opened 2026-09-03
fix(prompt-cache): template-aware assistant-primer splice — Qwen3.8 prefix cache hits again

Kaden-Schutt:
## Summary

Every conversational turn after the first re-prefilled the whole conversation on Qwen3.8 (measured `lcp=26` against `prior_len=118`, 2–6.5 s of re-prefill per turn on the coding session) because the jinja prompt-cache replay **prepends the live turn's generation primer to every cached assistant body on the assumption the template renders history turns bare.** Qwen3.5's template does; Qwen3.8's template re-emits `<think>\n\n</think>\n\n` on history assistant turns when thinking is off, so the spliced render carried the empty-think block twice, diverged 4 tokens into the first answer, and the LCP never hit.

Token-level proof (`HIPFIRE_QWEN_CACHE_TRACE=1`, qwen3.8:27b.mq4, two-turn probe):
```
common[20..26]  "assistant\n<think>\n\n</think>\n\n"
prior_past[26]  "```python\ndef is_prime(n):…"
rend_past[26]   "<think>\n\n</think>\n\n```python…"   ← primer doubled
lcp=26 prior_len=118 rendered_len=148
```

The fix probes the template once (`template_emits_history_primer`: render a one-exchange history with a sentinel answer, check whether the primer sits between the assistant opener and the sentinel) and both lookups (AR in `ar.rs`, DFlash in `qwen.rs`) prepend the primer only when the template does not re-emit it. Qwen3.5 behaviour is unchanged; Qwen3.8 gets the extension it was getting only in error paths.

- [x] `crates/hipfire-runtime` (`prompt_frame.rs`: probe + test), `crates/hipfire-generate` (`ar.rs`, `qwen.rs`: two lookup sites)

## Evidence (RX 7900 XTX, `4bc750e73`, qwen3.8:27b.mq4/.mq5, greedy, thinking off)

Two-turn probe: `lcp=26 → lcp=118 = prior_len`, `cached_tokens=118`, and the fixed render is *shorter* (144 vs 148 — the double primer is gone).

Full 8-turn coding session (`session_coding.json`, serve_harness, q8 KV), before → after (AR arm shown; the fixed build's two arms are byte-identical turn by turn):

| turn | ctx | cached before | prefill before | cached after | prefill after |
|---|---|---|---|---|---|
| t2 | 3564 | 0 | 4654 ms | 3514 | 208 ms |
| t3 | 5885 | 0 | 4529 ms | 5843 | 189 ms |
| t4 | 8603 | 0 | 6411 ms | 8563 | 213 ms |
| t8 | 15490 | 0 | 2083 ms | 15456 | 258 ms |

22–30× cheaper per-turn prefill; all 8 turns pass with recall 3/3, same gen counts as before (greedy ⇒ output-preserving).

Not caused by this bug, measured in the same sessions: `mq5 + draft` on a 24 GB card OOMs at ~5k ctx regardless (#686 dropped the pairing on those tiers and has the details).

## Test plan
- [x] `cargo test -p hipfire-runtime --lib prompt_frame` — 53 pass incl. new `history_primer_probe_distinguishes_qwen35_and_qwen38_templates` (bare template → prepend, re-emitting template → don't, empty primer → don't)
- [x] `cargo test -p hipfire-generate` — 221 pass
- [x] Hardware: probe + full session above, both AR and DFlash arms


Kaden-Schutt:
Gate run 33907487315: holding the DFlash half. AR is proven on hardware (chains recover near-full prefix hits on both lanes, Qwen3.6 untouched, qwen35 delegates), and the stacked llama V2 refusal was verified as fail-closed. Fable's finding on the DFlash arm: with the draft loaded, the cache engages but **systematically misses the most recent assistant body** — on four consecutive turns the LCP stops exactly at that body's first token (`prior_len − lcp = stored length + 2`), so store and replay are not byte-identical and each body's history is rewritten to the spliced variant a turn later. Also unproven: a `HipError(719)` in `spec_prefix_realign`/`reset_recurrent` mid-turn-3 of the qwen3.6 DFlash chain (host wedged before the control ran; that wedge was the host's bad card 7b:00.0 again, now re-fenced pending another reboot). The full investigation is in the run's decision artifact.

Kaden-Schutt:
Correcting the finding on this PR, because the record as it stands points at the wrong defect.

I pulled the serve log from the run that produced it (gate run 33907487315, `fable-evidence/dflash-chain-qwen36-serve.log`). In order:

```
[qwen-cache store dflash]  fp=0xb8df43230469ae1b cached_seq=1485
[qwen-cache lcp dflash-jinja] prior_len=1528 rendered_len=1582 lcp=1528
[qwen-cache HIT dflash] reuse prefix=1528 suffix=54 (no reset)
[qwen-cache store dflash]  fp=0xd7a7d25bf966ba4c cached_seq=683
[qwen-cache lcp dflash-jinja] prior_len=2266 rendered_len=2290 lcp=2266
[qwen-cache HIT dflash] reuse prefix=2266 suffix=24 (no reset)
spec_prefix_realign: HipError(719): hipMemcpy H2D offset: unspecified launch failure
[qwen-cache jinja lookup dflash] fp=0xb8df... hit=false
[qwen-cache jinja lookup dflash] fp=0xd7a7... hit=false
prefill: HipError(719): hipMemsetAsync: unspecified launch failure (reset_recurrent: qwen3…)   ×3
```

**`lcp == prior_len` on every hit** — full-prefix reuse, `suffix=54` then `suffix=24`, `no reset`. There is no `stored + 2` slack in the log, and the later lookups are `hit=false` outright rather than partial LCPs.

So the "primer replay systematically misses the most recent assistant body" reading does not survive its own evidence: the ordering shows the **`HipError(719)` comes first and the misses follow it**. 719 is `hipErrorLaunchFailure` — sticky and context-killing — so once it fires, every subsequent lookup legitimately misses and no amount of `reset_recurrent` can recover.

That makes the real work here:

1. **`spec_prefix_realign`'s H2D copy faulting.** The usual cause of a sticky 719 is a launch against a pointer that has been freed or reallocated. We hit that exact class twice tonight in sibling PRs — dangling kernargs after a rollback (#691) and a VRAM pool never drained on a refused load (#686). Worth asking whether the realign path assumes device state that a cache HIT reusing a prefix (`no reset`) never re-established.
2. **The retry loop.** After a sticky 719 the context is dead, yet the daemon retried three more times and produced three more 719s. Failing the request closed on the first sticky error is a correctness fix independent of the root cause.

The AR half is unaffected and was already proven on both lanes. An agent is on (1) and (2) now; I'll post what it finds. Config for the record: qwen3.6-27b, DFlash on, ChatML, thinking on with non-empty `reasoning_content`, `tool_calls=0`.

Kaden-Schutt:
Merged to `beta` as `acae74dd4` (branch tip `fa30b3ae2`).

Run [33934632513](https://github.com/warpfront/hipfire/actions/runs/33934632513): **both lanes pass** at `370592d17` (hiptrx gfx1201 + hipx gfx1100, `verdict: pass`, zero non-pass modes), Sol `needs-human` with **zero regressions**, Fable unavailable (credits) — so this staging merge is mine, not the seat's, and `master` stays human-promoted.

**The finding that held this PR was misdiagnosed, and its own log settles it** — full detail in the comment above. Short version: `lcp == prior_len` on every hit, so primer replay was never missing the assistant body; a sticky `HipError(719)` in the windowed `draft_seed_backfill` H2D upload came first, and the `hit=false` lookups after it are the rollback clearing `asst_turn_cache` working as designed.

What landed on top is the fix that is correct regardless of the 719's root cause: **the first sticky fault is now the last.** `700`/`719` latch a first-wins `GpuPoison` in `reset_core.rs`, the qwen35 reset/prefill error sites latch on the typed `HipError` (no string sniffing), and the daemon fails fast pre-generate with class `gpu`, non-retryable, *"process restart required"*. The latch never auto-clears, because a model reload does not reset the primary HIP context — clearing it would just burn one prefill per reload.

600 tests in `hipfire-runtime` and 189 in `hipfire-arch-qwen35`, including `sticky_poison_latch_lifecycle`.

Still open and recorded rather than invented: the 719 itself needs a device-sync bisection across the backfill chunk ops, plus a pointer-lifetime trace on `draft_scratch.{target_hidden,k_full,v_full,positions_k}` across a HIT → realign. That is hardware work; I have the hosts when you want it.

PR #694, opened 2026-09-04
Escha-W2 port: 2-bit trellis decoded in the GEMV — Qwen3.6-35B-A3B and Qwen3.8-27B

nwoolmer:
Ports **Escha-W2** — EschaLabs' 2-bit trellis quantization — into hipfire, and
ships six artifacts across two models.

Published: [qwen3.6-35b-a3b-escha](https://huggingface.co/hipfire-models/qwen3.6-35b-a3b-escha) · [qwen3.8-27b-escha](https://huggingface.co/hipfire-models/qwen3.8-27b-escha)

## What it is

A quantization *format*, not a model family. 16×16 tiles, K bits per weight,
packed `int16[in/16, out/16, 16K]`. The codebook is a 3-op hash rather than a
table: `r = ((s*0xCBAC1FED) & 0x8FFF8FFF) ^ 0x3B603B60`, value
`= f16_lo(r) + f16_hi(r)`. Rotation is an unnormalised H128 (Sylvester) on both
sides. Consecutive weights are overlapping 16-bit windows sliding by K bits, so
each weight is a full 16-bit state with a 65536-value alphabet.

**Codes are stored verbatim and decoded inside the GEMV** — no decode-at-load,
no re-quantisation. Lossless repack into MQ2 is impossible rather than merely
lossy: one expert projection uses 10,746 distinct fp16 values against MQ2's 4
per group. New quant types `ESCHA2T16 = 42` / `ESCHA3T16 = 43`, new
`RotationPlan::EschaH128`.

## Qwen3.6-35B-A3B (MoE, arch 6)

| file | size | resident | prefill | decode | PPL | KLD vs `-pro` |
|---|---|---|---|---|---|---|
| `qwen3.6-35b-a3b.escha-xt` | 11.39 GB | 12.04 GB | 886 tok/s | 63 tok/s | 8.0643 | 0.058963 |
| `qwen3.6-35b-a3b.escha` **(default)** | 11.84 GB | 12.45 GB | 725 tok/s | 55 tok/s | 7.6940 | 0.007907 |
| `qwen3.6-35b-a3b.escha-pro` | 12.34 GB | 12.94 GB | 684 tok/s | 47 tok/s | 7.6864 | 0.000000 |

## Qwen3.8-27B (dense, arch 5)

| file | size | prefill | decode | PPL | KLD vs `-pro` |
|---|---|---|---|---|---|
| `qwen3.8-27b.escha-xt` | 10.45 GB | 122 tok/s | 12.3 tok/s | 9.7242 | 0.008943 |
| `qwen3.8-27b.escha` **(default)** | 10.77 GB | 119 tok/s | 12.1 tok/s | 9.6753 | 0.000534 |
| `qwen3.8-27b.escha-pro` | 11.16 GB | 113 tok/s | 10.8 tok/s | 9.6486 | 0.000000 |

Against the non-escha ladder, same slice and harness:

| build | size | PPL | decode |
|---|---|---|---|
| `qwen3.8-27b.mq6` | 21.75 GB | 9.0042 | 9.5 tok/s |
| `qwen3.8-27b.escha` | 10.77 GB | 9.6753 | 12.1 tok/s |
| `qwen3.8-27b.mq3` | 12.62 GB | 10.0643 | 15.7 tok/s |

Every escha build beats `mq3` on quality while being smaller, and beats `mq6`
on speed *and* size. Against `mq3` it trades tokens/sec for quality-per-byte.

**Why decode trails the plain MQ quants.** `mq3`/`mq6` reach ~198–207 GB/s, the
memory ceiling, so for them smaller means faster. Escha reaches ~127 GB/s
because it also decodes the trellis. Ablated on the shipped kernel: removing
the decode arithmetic is 1.83×, which would put it at 203 GB/s (~18 tok/s).
That is the codec's 7 ops per weight, and it buys the 2-bit residency.

## Naming

`<model>.<format>[-variant]`, matching `qwen3.8-27b.mq4-xt`. `-xt`/base/`-pro`
order by size; the suffix describes the **dense** tensors only — coded weights
are byte-identical across all three builds of a model. MQ6 dense is the default
(+0.28% PPL on the 27B, +0.10% on the 35B, for +12%/+17% decode). `-pro` is a
bit-exact repack of Escha's per-row int8 into per-32-block `Q8_0`.

## Performance work

- **Prefill 52 → 108 tok/s**, three independent causes: `escha_h128_batched`
  rejected any chunk *smaller* than its scratch (so batched prefill never ran
  and everything fell back to the decode kernel); gfx1151 had no measured chunk
  default and fell to the generic 256 where 512 is the peak; and the grouped
  WMMA GEMM decoded the whole weight matrix 32× per 512-slot chunk with `NT=1`.
  rocprof puts those two kernels at 90.4% of prefill.
- **nt-major tile grid, +24% on the decode GEMV.** Consecutive `kt` were a full
  tile-row apart (139 KB on gate_proj); transposed at load so the payload stays
  verbatim. MoE stays kt-major, so kernels take an `nt_major` flag — both orders
  reduce to two loop-invariant strides. Gated on `mean KLD = 0.000000`.
- **G5 quality gate green**, negative control 2.134e-10. It had been failing its
  own control: two per-token kernels launched on the null stream inside a
  captured region.

## Bugs fixed along the way

- **1-byte overread past every 6-bit group in 9 WMMA kernels** — pre-existing,
  reachable at N<96.
- **Converter mixed-rotation bug**: `in_proj_a`/`in_proj_b` are F16 and fell
  through passthrough while their siblings were quantised, rotating the shared
  normed-x out from under them. Presented as a GDN "quality cliff" (PPL 2.4M).
- **Converter silently dropped the 27B's MTP head** — non-recursive `read_dir`
  missed the `mtp/` subdirectory.
- **`apply_ngram_block` over prompt-seeded history** made quoting the prompt
  impossible by construction, which read as long-context degeneration.
  Lab-only; no caller in `hipfire-engine`.

## Status

`hw-gate` cannot go green here — it needs a gfx1201 runner, so the reviewer
seats are skipped and the gate fails on a missing decision artifact regardless
of the code. Registry tags stay inert until this merges and the daily workflow
republishes `v1.json`.

https://claude.ai/code/session_01FfwZUVGT7QpjMdHikoERfM


Kaden-Schutt:
This one I'm holding as `meta:wip` rather than blocking 0.3.1 on it unless you want it in. If you do: beta's S3–S6 DFlash launch-fusion series (`803d17cee..87233eae1`) rewrote `crates/hipfire-arch-qwen35/src/qwen35/prefill.rs` around `DflashFusionCtx` / f16 projection producers, and `rdna-compute/src/gemv.rs` grew the residual verify tiers — merge-tree shows 8 conflict regions in `prefill.rs` alone. The escha batched-prefill arms need to be re-homed on the fusion scaffolding; that's an author job, not something I can resolve faithfully from the outside.

Also: please re-run G1–G6 on `qwen3.8-27b` for the dense path — the gate's fixture policy is Qwen3.8 XT now, and 3.6 evidence won't be accepted as the serve battery.

PR #716, opened 2026-09-04
fix(hw-gate): pin lanes by PCI address, not HIP index (the gfx1201 lane was the eGPU)

Kaden-Schutt:
Every gfx1201 hardware lane has been benchmarking the **Thunderbolt eGPU**, and I only found it because #682's lanes died mid-run tonight.

## The defect

HIP enumerates in KFD-node order, which is neither PCI bus order nor stable across hotplug. This host's live map:

```
HIP index map: 0=03:00.0  1=c3:00.0  2=e3:00.0  3=4b:00.0  4=13:00.0
```

The lane matrix said `device: "3"`, meaning "the fourth mainboard card". Index 3 is `4b:00.0` — the R9700 behind a JHL9480 TB5 bridge at **x4 @ 16GT/s**, versus x16 @ 32GT/s for the other four. `rocm-smi` confirmed it with a lane running: GPU[2] (=`4b`) at **99%** while all four mainboard cards sat at 0%.

So the narrowest-linked, dock-attached card in the box was carrying every gate benchmark. That explains the lane deaths with `MES(1) failed to respond to msg=REMOVE_QUEUE` on `4b` *during* a run, and why that card was the one fwupd's DP-AUX storm took down (`dce_aux.c:393`, `REG_WAIT timeout … acquire_engine`).

It also **moved**: when the eGPU re-enumerated `7b` → `4b` after a tunnel cycle, every index after it shifted, so the same literal `3` pointed at a different GPU than it had an hour earlier. A hardcoded index cannot express "this card".

## The fix

Lanes pin a PCI address (`pci: "0000:c3:00.0"` / `"0000:03:00.0"`) and resolve it to an index at lane start via `scripts/hw-gate/resolve_device.py`, which walks the KFD topology, skips the CPU node (`simd_count 0`), and unpacks `location_id` as `(bus << 8) | device`. **An absent pinned card is a hard error** naming what is present — falling back to an index would reintroduce the exact defect.

This does **not** disqualify the eGPU. It makes the choice explicit, so giving that card a lane becomes a decision instead of an accident.

## Tests

`scripts/hw-gate/tests/test_resolve_device.py` — KFD order vs bus order, CPU node not consuming an index, the `7b`→`4b` shift, absent-card hard error, address forms, both CLI exit paths. It pins `4b:00.0 → 3` explicitly so nobody reintroduces this as "just use index 3". **115/115** hw-gate tests pass, and the resolver was checked against the live topology (`--pci 0000:c3:00.0` → `HW_GATE_DEVICE=1`).

Policy-floor by construction (`.github/`, `scripts/hw-gate/**`), so it cannot self-merge.

### DIFF
--- a/.github/workflows/hw-gate.yml
+++ b/.github/workflows/hw-gate.yml
@@ -211,14 +211,19 @@ jobs:
           # `pci` is stable across boots and re-enumeration; the lane resolves it
           # to an index at runtime.
           - {name: hiptrx, gfx: gfx1201, labels: [self-hosted, hiptrx, gfx1201], pci: "0000:c3:00.0"}
-          # hipx: the RX 7900 XTX (gfx1100); the 6950/5700/8060S are not gate lanes yet.
-          - {name: hipx, gfx: gfx1100, labels: [self-hosted, gfx1100], pci: "0000:03:00.0"}
+          # hipx: the RX 7900 XTX (gfx1100) is 0000:66:00.0. Its other three
+          # cards — bf:00.0 gfx1151, 6e:00.0 gfx1030, 99:00.0 gfx1010 — are not
+          # gate lanes. Note hipx and hiptrx share no addresses, which is why
+          # `--expect-gfx` below is not decoration: pinning this lane to a
+          # hiptrx address is exactly the mistake it caught during review.
+          - {name: hipx, gfx: gfx1100, labels: [self-hosted, gfx1100], pci: "0000:66:00.0"}
     runs-on: ${{ matrix.lane.labels }}
     timeout-minutes: 60
     env:
       HIPFIRE_HOME: ${{ github.workspace }}/.hw-gate-home
       HIPFIRE_MODELS_DIR: /home/kaden/.hipfire/models
       HW_GATE_PCI: ${{ matrix.lane.pci }}
+      HW_GATE_GFX: ${{ matrix.lane.gfx }}
       CARGO_TARGET_DIR: /home/kaden/actions-runner/_cache/hw-gate-target
     steps:
       - name: Base checkout (gate scripts)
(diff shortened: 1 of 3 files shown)


Kaden-Schutt:
Found and fixed a bug in this PR before merge, worth recording because it is the same failure class the PR removes.

**The hipx lane was pinned to `0000:03:00.0` — a hiptrx address.** The two hosts share no PCI addresses; hipx's gfx1100 is `0000:66:00.0`. Verified against both real topologies:

```
hiptrx  0=03:00.0(gfx1201) 1=c3:00.0(gfx1201) 2=e3:00.0(gfx1201) 3=4b:00.0(gfx1201) 4=13:00.0(gfx1201)
hipx    0=66:00.0(gfx1100) 1=bf:00.0(gfx1151) 2=6e:00.0(gfx1030)  3=99:00.0(gfx1010)
```

Left uncaught it would have failed every hipx lane on the first run after merge — silent-wrong-card, just louder.

**So the pin is now self-checking.** Lanes pass `--expect-gfx $HW_GATE_GFX` (the arch already in the lane matrix) and the resolver compares it against the node's `gfx_target_version`, failing with every card listed. Wrong address, swapped card, and re-slotted card are all hard errors now rather than a benchmark on the wrong GPU:

```
$ resolve_device.py --pci 0000:4b:00.0 --expect-gfx gfx1100   # the eGPU
resolve_device: pinned card 0000:4b:00.0 is gfx1201, not the expected gfx1100
  (cards: 0=03:00.0(gfx1201), 1=c3:00.0(gfx1201), …) -- the card at that address changed
rc=2
```

`gfx_name()` decodes `major*10000 + minor*100 + step` with hex digits, verified on both hosts (`120001`→gfx1201, `110000`→gfx1100, `110501`→gfx1151, `100300`→gfx1030).

**120/120** hw-gate tests pass, and all three paths were run live on both hosts: hiptrx `c3:00.0` → index 1, hipx `66:00.0` → index 0, wrong-arch → exit 2.

PR #667, opened 2026-08-31
fix(gemma4): enable lowered 26B-A4B generation with row-safe HFQ4-G128 and rollover-safe bounded Q8 KV

HUSRCF:
## Summary

This enables the lowered Gemma4-26B-A4B generation path with the mixed expert formats produced by the current quantizer, while fixing two correctness boundaries required by that model:

- pack HFQ4-G128 tensors row-by-row so `K=704` expert down-projection rows do not share a partial quantization group;
- predicate HFQ4-G128 GEMM/GEMV tail lanes instead of reading beyond the logical row;
- route lowered Gemma4 MoE through the native AR/Jinja/sampling path and the hd512 asym3 attention kernels;
- allocate the lowered sliding/full KV caches by their actual layer counts;
- keep the 1024-token sliding Q8 cache bounded with modulo writes and compact-ring attention after rollover.

The rollover fix is correctness-critical. Before it, the lowered opcode path selected `AttnFlashQ8_0` at `pos=1927` for a cache with `physical_cap=1024`, so the ordinary Q8 writer used the absolute position and raised HIP illegal-memory-access. The fixed path remains `AttnFlashQ8_0Windowed` and completed a 29,788-token prompt.

## Artifact provenance

The validation artifact was generated from `google/gemma-4-26B-A4B-it`, downloaded from ModelScope on DiamondHill, then quantized with Hipfire's current quantizer after the HFQ4-G128 row-tail correction. It is a locally produced validation artifact, not yet an official Hipfire model-repository artifact.

- local artifact: `gemma-4-26b-a4b-it.hfq4g128-tailfix.hf4`
- size: `15,632,064,572` bytes
- SHA-256: `45da530b43e0e8ea2cffd77fe17c69bfb8dce3ec4dde2a26a9b68ad428d01f1c`

If this route is accepted, the corresponding converted artifact may need to be published in the Hipfire model repository before the model can be pulled through the normal registry flow.

## Validation

Local GPU: W7900 gfx1100, GPU0, HIP 7.14. Tests used the artifact above.

- `cargo test -p hipfire-quantize --bin hipfire-quantize hfq4g128_row_tests`: 2 passed
- `cargo test -p hipfire-dispatch kv_tier`: 42 passed
- `cargo build --release --workspace --features deltanet`: passed
- `target/release/examples/test_kernels`: 16 passed, 0 failed, 0 skipped
[...]

PR #734, opened 2026-09-08
fix(HFQ4 MoE): restore Gemma4 K=2816 tail groups without retained-PM4 scratch

HUSRCF:
## Summary

- fix the indexed HFQ4-G256 MoE gate/up kernel for Gemma4's `K=2816` shape (`11` groups)
- evaluate the final three groups that the quad-only A3B specialization previously dropped
- use fixed-group, uniquely named K2816 variants so LLVM keeps private scratch at zero for retained PM4
- preserve gfx1100 GLC/SLC/DLC selection and register the new symbols in replay metadata

## Root cause

The shared indexed kernel hard-coded `tail = 0`, which is valid for A3B `K=2048` (`8` groups) but not Gemma4 `K=2816` (`11` groups). Gemma4 therefore skipped groups 8..10, or 768 of 2816 input dimensions, in every affected routed gate/up projection. A runtime `groups_per_row & 3` repair restored the math but compiled with 16 bytes of private scratch on gfx1100/gfx1201, which retained PM4 correctly rejects. This patch specializes the admitted partial-quad product shape at compile time instead.

## Validation

Fixture:

- model: `gemma-4-26b-a4b-it.hfq4g128-tailfix.hf4`
- model SHA-256: `45da530b43e0e8ea2cffd77fe17c69bfb8dce3ec4dde2a26a9b68ad428d01f1c`
- prompt MD5: `43ca0d15712d3dfb777b51ae76d8fd5f`
- configuration: lowered Gemma4, AR/speculation off, asym3 KV, native no-think, greedy, `max_tokens=512`, `max_seq=65536`

GPU validation:

- W7900 / gfx1100: stable 1263-launch capture; HIP↔PM4 shadow `exact=true`; GDN-frame exact; E2E coherent with no attractor/runaway/empty output
- R9700 / gfx1201: stable 1263-launch capture; HIP↔PM4 shadow `exact=true`; GDN-frame exact; E2E coherent with no attractor/runaway/empty output
- decoded assistant output is byte-identical across gfx1100 and gfx1201 (MD5 `20cc5cc7be723d16b871d5813c8ea8b7`)
- both hosts used the same daemon binary (SHA-256 `ccd5790842cca8d9ca19decde8e139622036c216233b1b0c2f008fe4913fbcaa`)

Local checks:

- `cargo build --release`
- `cargo test -p rdna-compute hfq4_moe_gate_up_specializes_gemma4_tail_at_compile_time`
- `cargo test -p rdna-compute resource_wait_policy_and_a3b_pointer_catalog_fail_closed`
- `cargo test -p rdna-compute moe_shared_down_and_routed_gate_up_are_independent_siblings`
- `git diff --check`

`./scripts/no-gpu-ci.sh` reached 839/844 Python tests. Its only failures are the current unrelated `tests/test_mq4c_repack.py` collection/module collision (`mq4c_repack` missing `HfqmError`, `main`, and `parse_hfqm_index`); this patch does not touch that tool or its tests.

This is the shared-kernel correctness prerequisite discovered while completing the final-head validation for #667. No model artifact change or re-upload is required.


Kaden-Schutt:
Maintainer-integrated into beta at dea1dc3bf, preserving HUSRCF's original kernel commit with cherry-pick provenance. Added the corresponding MQ4 indexed-launcher specialization (otherwise it still dropped groups 8..10) and a real ignored GPU numerical regression; removed source-text selector assertions.

Parent gfx1201/R9700/HIP7.15 evidence:
- Before fix, K2816 CPU-dequant parity fails: MQ4 max absolute error 2.020895, HFQ4 5.202747. K2048 remains correct.
- Fixed kernels: all four MQ4/HFQ4 × K2816/K2048 cases pass (~2–3e-6 max absolute error).
- Re-ran the permanent regression in the isolated beta landing candidate: PASS at tolerance1e-4, finite-output asserted.
- Scoped formatting and generated rdna-compute map checks pass.

This lands ONLY the kernel prerequisite, not #667 enablement or batched-prefill changes. #667 remains held: its actual batched-vs-tokenwise prefill oracle diverged and is under repair. Kernel capture/replay was separately exercised on the Gemma-enabled candidate: stable1293 launches,22 kernels, exact HIP/blob/PM4 state/logits across3 positions; that is not a claim that this prerequisite alone enables Gemma serving.
No performance claim. Master remains untouched.

Issue #655, opened 2026-08-27
hipfire bug reports — RX 7900 XTX / Windows


drsohailimran:
<html><body>
<h1 dir="ltr">Issue 1</h1>

<p dir="ltr"><strong>Title:</strong> Degenerate output on gfx1100 (RX 7900 XTX) — reproduced across two ROCm toolchains, two models, two quant formats</p>

<h2 dir="ltr">Summary</h2>

<p dir="ltr">Every model I run on an RX 7900 XTX (gfx1100) produces token soup instead of
text. The model loads cleanly, kernels compile and certify without error, and
tok/s is plausible (~40) — but the output is not language.</p>

<p dir="ltr">I have eliminated the model files, the quant format, the compiler toolchain,
the sampling parameters, thinking mode, both interfaces, and the install
itself. Details below.</p>

<h2 dir="ltr">Environment</h2>

<div dir="ltr">
  |  
-- | --
GPU | AMD Radeon RX 7900 XTX (gfx1100), 24 GB
OS | Windows 11, build 10.0.26200.9168
hipfire | 0.3.0 (c1e96fd78dbe, master)
ROCm (A) | AMD HIP SDK 7.2 (7.2.60201-38d754472, clang 21)
ROCm (B) | TheRock ROCm 10.1 (rocm10.1.0a20260815, reports HIP 7.16)
MSVC | 14.44.35207 (14.51.36231 rejected by ROCm 7.2 clang — see note)
Install | ~/.hipfire junctioned to S:\hipfire

</div>

<h2 dir="ltr">Reproduction</h2>

<div role="group" aria-label="Code" tabindex="0"><div><div></div></div><div><pre style="color: rgb(234, 236, 240); background: transparent; font-family: var(--font-mono);"><code style="color: rgb(234, 236, 240); background: transparent; font-family: var(--font-mono); white-space: pre-wrap;">hipfire serve qwen3.5:0.8b</code></pre></div></div>

<div role="group" aria-label="Code" tabindex="0"><div><div></div></div><div><pre style="color: rgb(234, 236, 240); background: transparent; font-family: var(--font-mono);"><code style="color: rgb(234, 236, 240); background: transparent; font-family: var(--font-mono); white-space: pre-wrap;">POST http://127.0.0.1:11435/v1/chat/completions
{"model":"qwen3.5:0.8b","messages":[{"role":"user","content":"Count to five."}],
[...]

Issue #644, opened 2026-08-26
gfx1100: MQ4V2 (qt=44) decode at 0.1-0.3 tok/s - V2-ladder SKUs unusable on Windows/RDNA3


SergioSV96:
Registry V2-ladder SKUs for Qwen3.8-27B decode at **0.1–0.3 tok/s** on gfx1100 — ~200× slower than classic-MQ4G256 artifacts of the same architecture on identical hardware.

## Repro

Windows 11, RX 7900 XTX (gfx1100), ROCm 7.2 HIP SDK. Master tip `363e93d7` (+ #631/#622 cherry-picks for dual-GPU binding), `--locked` build.

```
PS> hipfire pull qwen3.8:27b-mq4-xt
PS> hipfire bench qwen3.8-27b.mq4-xt --runs 5 --warmups 3 --max-tokens 128 `
      --backend noslots --workload stateless --spec off --json
```

Load succeeds (61 kernels JIT'd), but generation crawls:

```json
"decode_tok_s":  { "median": 0.2,  "mean": 0.2, "min": 0.1, "max": 0.3 },
"prefill_tok_s": { "median": 0.4,  "mean": 11.0, "min": 0.2, "max": 27.3 },
"ttft_ms":       [ 69235, 60193, 96596, 901, 878 ]
```

For contrast, a classic-`MQ4G256` artifact (converted from GGUF via `--format mq4`) of the *same architecture* decodes at tens of tok/s on this exact box, and its prefill ran ~500 tok/s-class. So the base HFQ4/MQ4G256 lanes are fine; the slowdown appears specific to the **MQ4V2 (qt=44) execution path** on gfx1100.

## Impact

Every `qwen3.8:27b-*` V2-ladder SKU (mq3/mq4/mq5/mq6 × base/pro/xt, plus trunk) routes through this path — they convert and load cleanly but are not usable for inference on gfx1100 at current master. Presumably related to why registry text carries no live baselines for these SKUs and why #616 (packed-MQ4 backend) / #611 are still open.

## Request

If MQ4V2-on-gfx11 is simply "not admitted yet", a registry/CLI warning at pull time would save users a 15 GB download + 86-minute bench discovery. If it's supposed to work, happy to run any diagnostic head — full bench JSON retained.

HUSRCF:
@SergioSV96 — thank you for the sharp report. I downloaded and verified the exact registry artifact, then ran same-artifact controls on a Linux gfx1100 reference.

[...]

Issue #635, opened 2026-08-23
Windows: VMM KV backend fails at first decode (hipMemSetAccess invalid argument) - kv_backend=contiguous works


SergioSV96:
Windows: VMM KV backend crashes at first decode — `hipMemSetAccess: invalid argument`; `kv_backend=contiguous` works

Splitting this from #634 (GGUF naming) since it's an independent Windows runtime issue found during the same bring-up.

## Repro

master `80a572c824dc`, Windows 11, RX 7900 XTX (gfx1100), ROCm 7.2 HIP SDK. Native registry model, default config (`memory.kv_backend` unset → resolves to vmm):

```
PS> hipfire run qwen3.5-0.8b.mq4 "..."
GPU dev 1: gfx1100 (25.8 GB VRAM, HIP 7.2)
KV cache: Q8 vmm (6/24 layers carry KV; ... physical_cap=262144 ...)
  <model loads fully>
hipfire: daemon error: [internal retryable=false ...] forward_scratch decode:
HipError(1): hipMemSetAccess: invalid argument; VMM access prefix
base=0x4408000000 size=4194304 (new segment address=0x4408200000 offset=2097152
size=2097152) granularity=65536
```

Model load and kernel JIT complete; the failure fires on the first decode's VMM segment commit.

## Workaround (verified)

```
PS> hipfire config set kv_backend contiguous
PS> hipfire run qwen3.5-0.8b.mq4 "..."
KV cache: q8 (6/24 layers carry KV, others placeholder)
<full generation, coherent output>
```

## Suggestion

Either default `kv_backend` away from `vmm` when running on Windows HIP (runtime probe at init would be cleaner), or document the knob in GETTING_STARTED's Windows section — nothing today hints that the default allocator is broken on this platform.

HUSRCF:
@SergioSV96 — thank you for the precise report. I have a diagnostic fix ready in draft PR #643 at exact head `008db8a5`.

The reported range is correctly aligned to the Windows-reported 64 KiB granularity. The failure instead occurs on the first growth where hipfire asks `hipMemSetAccess` to cover a contiguous 4 MiB virtual prefix backed by two independent 2 MiB physical mappings. Linux ROCm 7.2 needs that prefix form because it can reject otherwise-valid non-zero subranges; Windows HIP 7.2 appears to have the opposite limitation and rejects a range spanning mappings.

[...]

PR #737, opened 2026-09-09
fix(vmm): map full reservation in one segment on Windows

harrytyp:
## Problem

`hipMemCreate`/`hipMemMap` on Windows/ROCm 7.2 (gfx1100) maps a second, later segment onto the **same physical pages** as the first. The repo's own `vmm_arena_smoke` example panics at its boundary-growth assert (line 68) after the 2nd `map_next` — segment 1 aliases segment 2's pages — while single-segment maps pass.

The KV cache grows in ~1900 small `map_next` steps, so on Windows every growth corrupts all prior KV. This surfaces as token soup (`"!!!!!!!!"`) from every model (qwen3.8:27b, qwen3.5:0.8b, lfm2.5:1.2b) in `hipfire serve`. Full repro and discussion: #655, #644.

## Fix

Under `#[cfg(windows)]`, `alloc_vmm_tensor` maps the **full reservation in one `map_next`** instead of growing in small segments. Linux behavior is byte-identical to before (the `cfg` block does not exist there).

Trade-off: costs full VRAM up front (no on-demand commit) on Windows only. This is a workaround for what looks like a ROCm-Windows VMM runtime bug — both mappings report distinct virtual addresses via `hipMemGetAddressRange`, yet share physical pages.

## Validation (gfx1100, RX 7900 XTX, Windows 11, ROCm 7.2)

- Output correct after fix: `17*23` → `391`, fluent prose, no token soup
- `vmm_arena_smoke` passes on the single-segment path; multi-segment growth assert is the repro for the underlying runtime issue
- AR decode ~40 t/s, DFlash draft ~48 t/s, 40k+ prefill completes without #635 crash
- No perf claim vs baseline (baseline output was corrupt by construction)

## HW-gate routes

- `scripts/serve_harness.py`: generation/state-lifecycle changed? **No** (dispatch/VMM allocation only)
- `scripts/redline_daemon_harness.py`: kernel/dispatch change? **Yes** — dispatch-layer VMM allocation path

### DIFF
--- a/crates/rdna-compute/src/dispatch.rs
+++ b/crates/rdna-compute/src/dispatch.rs
@@ -3065,6 +3065,17 @@ impl Gpu {
             .checked_mul(dtype.size())
             .ok_or_else(|| HipError::new(0, "VMM tensor byte size overflowed"))?;
         let mut arena = VmmArena::reserve(&self.hip, self.device_id, byte_size)?;
+        // WINDOWS FIX (2026-09-09): hipMemCreate/hipMemMap on Windows/ROCm 7.2
+        // (gfx1100) maps a second, later segment onto the SAME physical pages as
+        // the first (vmm_arena_smoke boundary-growth assert fails; every
+        // subsequent KV growth corrupts all prior KV -> token soup). Single
+        // segment maps are proven correct. So on Windows, map the FULL
+        // reservation in one map_next up front instead of growing in small
+        // segments; grow_vmm_tensor then becomes a no-op (already fully
+        // mapped). Costs up-front VRAM for the whole reservation; correctness
+        // over on-demand commit on the platform whose driver breaks growth.
+        #[cfg(windows)]
+        let initial_mapped_bytes = byte_size;
         if initial_mapped_bytes > 0 {
             if let Err(err) = arena.map_next(&self.hip, initial_mapped_bytes, access_devices) {
                 return Err(self.retain_failed_vmm_arena(arena, err));
--- a/crates/rdna-compute/src/dispatch.rs
+++ b/crates/rdna-compute/src/dispatch.rs
@@ -3075,7 +3075,14 @@ impl Gpu {
         // mapped). Costs up-front VRAM for the whole reservation; correctness
         // over on-demand commit on the platform whose driver breaks growth.
         #[cfg(windows)]
-        let initial_mapped_bytes = byte_size;
+        let initial_mapped_bytes = {
+            // map_next requires a multiple of the allocation granularity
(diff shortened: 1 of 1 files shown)


HUSRCF:
Thanks for putting this together and, especially, for validating it on the affected Windows/gfx1100 machine. The single-mapping workaround is consistent with the `vmm_arena_smoke` evidence and looks like the right safety policy for the affected runtime.

One small implementation detail may be worth adjusting before merge. The current Windows override uses:

```rust
#[cfg(windows)]
let initial_mapped_bytes = byte_size;
```

`VmmArena::reserve` rounds the requested byte size up to the HIP VMM allocation granularity, while `VmmArena::map_next` requires its size to be an exact multiple of that granularity. Therefore, a caller whose logical `byte_size` is not already aligned could still fail on Windows even though the reservation itself is valid.

Would it be safer to map the actual rounded reservation instead?

```rust
#[cfg(windows)]
let initial_mapped_bytes = arena.reserved_bytes();
```

This preserves the intended policy—one physical mapping covering the entire Windows reservation—while also keeping `alloc_vmm_tensor` correct for shapes whose logical byte size is not VMM-granularity aligned. The existing Qwen KV layout may happen to be aligned, but this API is shared by other VMM tensor callers as well.

Kaden-Schutt:
Prepared integration/pr737-windows-fullmap at d827d03e1, based on your full PR history, with two fixups: use arena.reserved_bytes() directly, and migrate VMM smoke/tests to the Windows full-map contract (plus a 4097-byte roundtrip regression). Independent review caught the old smoke still expecting partial mapping/growth. Linux gfx1201 4097-byte GPU regression passes; this does NOT validate the Windows branch. Before beta landing, please run on Windows ROCm 7.2: cargo run -p rdna-compute --example vmm_tensor_smoke; cargo test -p rdna-compute --lib vmm_ -- --nocapture --test-threads=1; then your real serve/KV corruption reproduction and applicable capture/parity route. Please attach exact tested SHA and decoded output. Candidate: https://github.com/warpfront/hipfire/tree/integration/pr737-windows-fullmap . No Windows execution host is available in this session, so holding rather than claiming proof.

Kaden-Schutt:
Landed corrected #737 into beta0db445ecb with original PR commits, reserved_bytes alignment fix, migrated full-map smoke/tests and generated crate-map refresh. Maintainer explicitly waived the remaining Windows-host validation requirement. Fresh merged-tree 4097-byte GPU roundtrip regression passed on Linux/gfx1201/HIP7.15; Windows execution is NOT claimed.

PR #738, opened 2026-09-09
fix(probe): find daemon binary with .exe on Windows

harrytyp:
## Problem

`find_daemon_binary()` probes `target/release/daemon` and `target/debug/daemon` without the `.exe` extension, so `coherence_probe` always exits with "daemon binary not found" on Windows.

## Fix

Add the `.exe` candidates first (release + debug), keeping the extensionless ones for Unix. Two-line change, no behavior change on non-Windows (the `.exe` paths simply don't exist there and `exists()` falls through).

## Validation

Verified locally on Windows 11: probe now locates `target/release/daemon.exe` and proceeds past binary discovery.

## HW-gate routes

- `scripts/serve_harness.py`: generation/state-lifecycle changed? **No**
- `scripts/redline_daemon_harness.py`: kernel/dispatch change? **No** (example tooling only)

### DIFF
--- a/crates/hipfire-runtime/examples/coherence_probe.rs
+++ b/crates/hipfire-runtime/examples/coherence_probe.rs
@@ -175,6 +175,8 @@ fn find_daemon_binary() -> Result<PathBuf, String> {
     // Prefer release; fall back to debug. Mirror the gate scripts'
     // discovery behaviour.
     let candidates = [
+        "target/release/daemon.exe",
+        "target/debug/daemon.exe",
         "target/release/daemon",
         "target/debug/daemon",
     ];


Kaden-Schutt:
Merged into beta at d2e2ee0c5, preserving the original commit and adding 7240773c7 to follow the CLI platform policy: Unix ignores stale .exe files; Windows prefers .exe with the existing bare-name fallback. Independent code review approved. Extracted real-function filesystem smoke passed 5 Unix and 6 simulated-Windows cases. Actual Windows process launch was not exercised here; the PR author supplied the original Windows discovery receipt.

PR #757, opened 2026-09-14
gfx1201: FP8-WMMA MQ4v2 prefill GEMMs (+28% pp512), opt-in; GDN permlanex16 fix

Kaden-Schutt:
## What

Two commits:

1. **`gdn: fix permlanex16 lane selects in gfx1151 DPP reduce`** — `hipfire_gdn_reduce_sum_lane0_dpp` passed zero lane-select DWORDs to `permlanex16`, broadcasting lane 16 instead of swapping halves (bug first recorded in `2026-08-05-mq2r-a3b-lowbit-campaign.md`). Now uses the identity selects and is bit-identical to the `__shfl_down` tree it replaces (verified on gfx1201 with a 32-lane parity kernel). Path is opt-in (`HIPFIRE_GFX12_GDN_DPP=1`); no default output affected.

2. **`gfx1201: FP8-WMMA MQ4v2 prefill GEMMs`** — one kernel source `kernels/src/gemm_gate_up_mq4g256v2_wmma_fp8.gfx12.hip` compiled in three ABI modes (gate/up overwrite, residual `Y+=`, 4-way QKVZA) × bt4/8/12, plus a pre-pass `pack_f16_to_fp8_mq4v2.gfx12.hip`.
   - Weights: MQ4v2 nibble → E4M3 **exactly** (0..15 representable) via `v_perm_b32` table lookup — weight side lossless.
   - Activations: F16 → E4M3 once per projection with optional pow2 per-row scale; per-(row, 128-K half) sums of the *rounded* values make the affine zero-point exact (`z_h · Σ_half x`).
   - Accumulation: one f32 chain per batch tile, kept in units of the current half's scale (`acc *= s_prev/s_new` at each 128-K boundary); K re-sliced so one 16 B X load feeds two WMMAs. No LDS, 238 VGPR, 0 spills at bt12.

## Numbers (R9700 gfx1201, HIP 7.15, Qwen3.8-27B MQ4XT sha256 `9f91556f…`, `hipfire bench --matrix --spec off --backend noslots --workload stateless`)

| | F16 (default) | FP8 (flags on) | Δ |
|---|---:|---:|---:|
| pp512 (5 runs, prior base) | 843 | **1079** | +28% |
| pp2048 (5 runs, prior base) | 801 | **1009** | +26% |
| pp512 (3 runs, this base) | 857 | **1083** | +26% |

[...]

PR #758, opened 2026-09-14
gfx1201: GQA-fused FA2 prefill attention (Q8-K and fwht3-K), opt-in — pp20000 480 → 1112

Kaden-Schutt:
Stacked on #757 (base = `fp8-mq4v2-gfx1201`).

## What

New flash-attention **prefill** kernel for gfx1201, hd 256, H/Hkv = 24/4: one workgroup per (KV head × 8 positions) = 48 flattened (position, head) rows in three WMMA tiles, 4 waves; each KT64 K/V tile is dequantized **once** into 64 KiB swizzled f16 LDS and shared by the six GQA heads; online softmax in registers; causal block skipping. Operand order is `Sᵀ = K·Qᵀ`, `Oᵀ = Vᵀ·Pᵀ` so the score tile lands directly in the PV operand layout (no transposes, no cross-lane shuffles in the hot loop). Two K modes from one source: `KMODE=0` standard Q8 K, `KMODE=3` **fwht3 K** (production default: 3-bit FWHT-rotated, 100 B/head) with the signed FWHT-256 applied to Q once per WG.

Why: under the default fwht3-K cache, prefill attention was routed to `attention_flash_fwht3_tile_batched` — a decode-style scalar tile kernel (one wave per head × tile × query row). It was the long-prompt bottleneck and invisible to `HIPFIRE_PROFILE` (no timer on that launch; #757 adds one).

## Numbers (R9700, Qwen3.8-27B MQ4XT sha256 `9f91556f…`, all #757 FP8 flags + SLABS=2, chunk 512, `--matrix` 3 runs, prompt md5 `d94d3115…`)

Default KV (fwht3-K), `HIPFIRE_GFX12_FA2_PREFILL` 0 → 1:

| | off | on | Δ |
|---|---:|---:|---:|
| pp512 | 1367 | **1419** | +4% |
| pp2048 | 1203 | **1388** | +15% |
| pp8192 | 800 | **1281** | +60% |
| pp20000 | 480 | **1112** | +132% |

Decode untouched (36.50 → 36.47). With `--kv-mode q8` the same kernel gives 1277 / 1107 at 8k / 20k.

Kernel (oracle bench, N=384, CTX=7936, 21-run median, 3 fresh processes): fwht3 incumbent 26.26 ms / 2.8 TFLOPS → **3.35 ms / 21.8 TFLOPS (7.84×)**; Q8 incumbent 8.71 ms → 3.42 ms (2.55×).

## Correctness

- CPU-f64 oracle, 6 shapes × tail/ragged + causal sentinel: 12/12 pass both K modes; candidate rel_l2 vs f64 is **lower** than the incumbents' on every shape (e.g. 3.9e-5 vs 8.1e-5 at ctx 4096).
- WT2 KLD (24 chunks, prefill scoring, fwht3-K): **0.063421 → 0.062485** (p99 1.348 → 1.307, PPL 6.4461 → 6.4408). Q8-K: 0.059001 → 0.059029. Greedy identity byte-identical on the short fixture (below the 64-row admission, so that proves fall-through only).
- Resource: fwht3k 241 VGPR, Q8 242, 0 spills, 0 scratch, dynamic LDS 65 536. `KMODE=0` ISA identical before/after the fwht3 commits.

## Scope / safety

- Default OFF: `HIPFIRE_GFX12_FA2_PREFILL=1`. Exact gfx1201, H24/KV4/D256, batch 64..512 (%16), ctx 64..32768, V Q8 (`v_mode 8`), no tree mode, eager-only fail-closed (graph capture / replay keep the incumbents; the in-place Q rotation is not replay-idempotent). `replay.rs` untouched.
- Split-KV (partial + merge) symbols exist but are **not** wired: the partial epilogue O-write is wrong (misdirected records for ml≥8 lanes); oracle arm gated behind `SPLIT=1`. Direct-only is what is measured. Fixing it is what would make FA2 usable at DFlash-verify batch sizes.
- Throwaway oracle/bench example `crates/rdna-compute/examples/fa2_prefill_gfx1201_oracle.rs` is left in tree for now (CPU-f64 reference + 21-run bench); delete or promote before merge to beta.

## Not in this PR

- lloyd-V / tree verify / other archs. Ornith A3B needs the row mapping parametrized on the GQA ratio (8 pos × 8 heads = 64 rows exactly).
- Remaining levers measured but not taken: persistent Q fragments (255-VGPR coin flip), `ds_read_b64` fragment reads, double-buffered KT32 fill.


Kaden-Schutt:
Three more commits on this branch (dense TP=2, 27B on 2× R9700, all FP8 flags + SLABS=2 + FA2, prompt `ds4_heterogeneous_code_2048` 2087 tok):

- `bc444ce7d` `hipfire bench --tp N` forwards to the daemon load (dense Qwen tp>1 already routed to `load_model_tp_qwen35_dense`; `--matrix` still refuses multi-GPU, real-prompt bench works).
- `15579c875` symmetric direct-peer-read allreduce with fused residual add (`add_peer_residual_f32`, each rank reads the peer's partial over P2P on its own stream; two event pools, no root, no memcpy). Kill switch `HIPFIRE_TP_PEER_DIRECT=0`. Isolated 4 MiB AR 343 → 151 µs. Parity example passes; greedy byte-identical rooted vs direct.
- `7b37e887c` (merged from `tp-vmm`) VMM KV backend admitted for dense TP: one arena per rank, `ensure_vmm_ready_for_load`/`ensure_vmm_cleaned` per rank, reload cycles clean, greedy byte-identical vs contiguous.

| | TP=1 | TP=2 rooted (before) | **TP=2 direct** |
|---|---:|---:|---:|
| prefill tok/s | 1318 | 1905 | **2104** (1.60×) |
| decode tok/s | 28.6 | 39.4 | **50.3** (1.76×) |
| TTFT ms | 1583 | 1095 | **992** |

Same numbers with `--kv-backend vmm` (2100 / 50.1). Re-run independently by me: 1904→2101 / 39.5→50.2.

P2P baseline on this box (hipMemcpyPeerAsync 0→1): ~56 GB/s one-way, 63 GB/s bidirectional, 3 µs floor; peer access enabled by the loader.

